PR #168, opened 2015-01-16
Free the buffer after use.

foollbar:


FROGGS:
I reverted it in #177:

There are two issues with this PR:

1) unrelated whitespace cleanups shall not just sneaked in with a PR
2) an MVMString (and MVMObject) cannot be freed just like taht, the garbage collector will care.

The MVM_free(buf);is valid though.

PR #172, opened 2015-01-30
Zero-fill CArrays

raydiak:
See discussion around http://irclog.perlgeek.de/perl6/2015-01-29#i_10027737

I don't know what the performance implications of this are.  One suggestion was to make this default but optional.  I'm not too experienced with either C, or the plumbing to pass/store those options around between zavolaj/nqp/moar, so I didn't do that part yet.  But at least for my uses, this patch solves a couple problems right away, since iterating over a large array from perl 6 code is a major bottleneck.


raydiak:
I am wondering if the consts on lines 186/187 and 199/200 will always be the same.  If so, should they be combined into one pair of declarations before line 185?

PR #177, opened 2015-02-25
fix allocating an empty (no members) CStruct bug

sergot:
it segfaults when we try to allocate memory for an empty CStruct, this fix prevents from allocating memory of size -1


FROGGS:
I think the problem was more that we read the uninitialized memory in case the struct had no members and we allocated a single byte... But anyway, it fixed the segfault.

PR #189, opened 2015-03-16
Test “--no-jit” instead of “--enable-jit”

baby-gnu:
From Configure.pl:

    The --enable-jit flag is obsolete, as jit is enabled by default.
    You can use --no-jit to build without jit


zhuomingliang:
Thank you, I merged it from https://github.com/MoarVM/MoarVM/pull/205.

PR #195, opened 2015-04-02
Allow nativecast from VMArray e.g. Blob/Buf

raydiak:


raydiak:
spectest clean other than an unrelated-looking double-free/corruption error in S17-lowlevel/thread.t

PR #197, opened 2015-04-05
Lint 20150405

fperrad:
some cleanup found with a lint tool


paultcochrane:
@fperrad  Just out of interest, which lint tool did you use?

fperrad:
[PC-lint](http://www.gimpel.com)

PR #200, opened 2015-04-08
refactor pthread_yield detection

fperrad:
the current detection is fragile.
for example, with the musl libc on linux, it fails.

the detection must be done during the configure step.


FROGGS:
:+1:

PR #202, opened 2015-04-12
enforce C99 compliance

fperrad:


FROGGS:
@fperrad Why do we enforce C99 compliance when we want to comply to C89?

fperrad:
In 2015, with 3rd party components, this goal seems unreachable.

On Parrot, I remember that this constraint (C89) was painful.

jnthn:
The actual important constraint is "don't break Visual C++, which lacks various C99 things". Any third-party thing that fails to be at least cross-platform enough to do Windows is probably not too interesting for us anyway. :-)

FROGGS:
I reverted 56dcbaa `enforce C99 compliance` as d8ee582.

PR #209, opened 2015-04-19
Loop over waitpid to get correct exit code for closed pipe

usev6:


usev6:
This fixes the problem described in https://rt.perl.org/Ticket/Display.html?id=123563. Also t/spec/S32-io/pipe.t from roast passes with this patch on FreeBSD.

I'm not sure, whether it is reasonable to loop over waitpid instead of calling it once.

PR #229, opened 2015-07-06
Clear up some clang warnings

virtualsue:
There were some niggly warnings, of the following sorts:
- format macro adjustment to match argument type
- uint/int casting needed
- an extra space accidentally put in front of a %s format specifier (which is useful with signed int formatting)


FROGGS:
Hi, this warning showed up now on my linux+gcc box:
```
src/spesh/dump.c:390:17: warning: format ‘%lld’ expects argument of type ‘long long int’, but argument 3 has type ‘MVMint64’ [-Wformat=]
                 appendf(ds, " P6int(%lld)", MVM_repr_get_int(tc, (MVMObject*)seen_table[log_index]));
                 ^
```

I changed the %lld there to %"PRId64", which probably also works fine for clang. (https://github.com/MoarVM/MoarVM/commit/f090305760)
I applied (cherry-picked) your non-merge commits.

PR #231, opened 2015-07-09
Free strings used for exception formats

hoelzro:


hoelzro:
Merged via eb0c964c1f

PR #233, opened 2015-08-14
Add support for parsing Unicode numbers

ShimmerFairy:
This change allows MVM_radix and MVM_bigint_radix to parse numbers containing any Nd category characters as well as letters from scripts with characters that can be used as hex digits (according to the Hex_Digit property), which effectively means just adding the upper- and lowercase fullwidth versions of the ASCII alphabet as valid characters.

### DIFF
--- a/src/core/coerce.c
+++ b/src/core/coerce.c
@@ -358,9 +358,19 @@ MVMObject * MVM_radix(MVMThreadContext *tc, MVMint64 radix, MVMString *str, MVMi
     }
 
     while (offset < chars) {
-        if (ch >= '0' && ch <= '9') ch = ch - '0';
+        // as of Unicode 6.0.0, we can be assured that Nd numerals are within
+        // 0..9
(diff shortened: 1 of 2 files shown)


ShimmerFairy:
I did some benchmarking at japhb's suggestion, and found a slight decrease in speed. Though I don't think it would've affected things too badly, I restructured things so that the speed decrease wasn't an issue.

Here's a table summarizing the speed averages (I could gist the whole file I kept track in if desired). "uninum" refers to my branch when initially pull-requested (86e4ac0), and "fastpath" refers to the changes I've pushed just now (813bbd9).

 | master | uninum | uni - master | fastpath | fast - master | fast - uni |
--------|--------|--------|--------------|----------|---------------|-------------
 NQP   |  0.338 |  0.587 |       +0.249 |    0.343 |        +0.005 |     -0.244 |
 Perl6 | 10.696 | 11.106 |       +0.410 |   10.769 |        +0.073 |     -0.337 |

PR #247, opened 2015-08-30
Plug resource leak in fixedsizealloc.c

paultcochrane:
The variable `to_add` is being assigned and memory for it allocated, but
this isn't freed before returning from the functions where it is used.  This
patch frees the potentially leaked memory.  The NQP tests pass with this
patch.  Nevertheless, please review!

### DIFF
--- a/src/core/fixedsizealloc.c
+++ b/src/core/fixedsizealloc.c
@@ -204,6 +204,7 @@ static void add_to_overflows_safepoint_free_list(MVMThreadContext *tc, MVMFixedS
         orig = al->free_at_next_safepoint_overflows;
         to_add->next = orig;
     } while (!MVM_trycas(&(al->free_at_next_safepoint_overflows), orig, to_add));
+    MVM_fixed_size_free(tc, al, sizeof(MVMFixedSizeAllocSafepointFreeListEntry), to_add);
 }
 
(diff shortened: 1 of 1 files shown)


jnthn:
Pretty sure this is wrong; to_add is clearly being cas'd into the list and so is referenced.

paultcochrane:
On 30 August 2015 3:27:38 pm GMT+02:00, Jonathan Worthington <notifications@github.com> wrote:
>Pretty sure this is wrong; to_add is clearly being cas'd into the list
>and so is referenced. 
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/MoarVM/MoarVM/pull/247#issuecomment-136138304

So this is a false positive I should tell Coverity about?

paultcochrane:
Actually that was a silly question: of course I should mark it as a false positive.  The better question, however, would be: how can I test this situation (or ones like it) such that I can prove that the problem Coverity highlights actually *is* a false positive?  The reason why I ask is that there are other similar(ish) situations being flagged by Coverity and I'd like to be able to test them myself before submitting a PR.

jnthn:
My first guess is that it doesn't understand the CAS operation (that is, that it is a potential assignment).

PR #251, opened 2015-09-02
Avoid possible divide-by-zero error

paultcochrane:
This merely requires that the `multiple_of` variable is initialised to 1 instead of to 0.  This change avoids a potentially nasty issue and I don't believe changes actual behaviour.  The NQP tests pass, however, please review to check that I've not changed behaviour unintentionally.


FROGGS:
Good catch!

PR #255, opened 2015-09-04
Fix write to null handle on async socket.

kanatohodets:
Related to this blurb: https://gist.github.com/kanatohodets/6642786a927c2616f6f2

I think this is safe because the handle is caught by MVM_free in close_cb when the close_socket task is processed by the event loop.

Feedback welcome!


zhuomingliang:
I'm not sure about it. sine ```data->handle = NULL``` is to avoid ``` if (data->handle) {``` (https://github.com/kanatohodets/MoarVM/blob/master/src/io/asyncsocket.c#L462) run again, ```MVM_free()``` won't set it to NULL.

PR #271, opened 2015-09-25
Lint 20150925 instrument

fperrad:
some linting in src/instrument

PR #272, opened 2015-09-25
Lint 20150925 gc

fperrad:
some linting in src/gc


AlexDaniel:
Thanks, but have to close. Explanation here: https://github.com/MoarVM/MoarVM/pull/271#issuecomment-615863053

PR #273, opened 2015-09-27
Lint 20150927 core

fperrad:
some linting in src/core


fperrad:
with c++, I like
``` c
    while (true)
```
same thing, with c99
``` c
    #include <stdbool.h>

    while (true)
```
with c89 (ie. without `bool`), I try to avoid condition with a literal integer, so I prefer
``` c
    for (;;)
```
instead of
``` c
    while (1)
```

In the same way, *real* conditions are more readable (or less confusing) than single variables
``` c
    if (val)
    if (val != 0)
    if (ptr)
    if (ptr != NULL)
```

PR #279, opened 2015-10-07
Check return value of uv_listen(). Throw AdHoc exception if it's failed.

tokuhirom:


tokuhirom:
This helps to handle "Address already in use".

PR #287, opened 2015-11-01
fix logical-op-parentheses compile warnings...

leejo:
in normalize.h:

    src/strings/normalize.h:71:33: warning: '&&' within '||' [-Wlogical-op-parentheses]
    if (in < 0x20 || in >= 0x7F && in <= 0x9F || in == 0xAD)
                  ~~ ~~~~~~~~~~~^~~~~~~~~~~~~
    src/strings/normalize.h:71:33: note: place parentheses around the '&&' expression to silence this warning
        if (in < 0x20 || in >= 0x7F && in <= 0x9F || in == 0xAD)
                                    ^
                         (                       )
    1 warning generated.
    compiling ....

wrap the >= && <= test in parentheses to fix this warning which is
showing (?) a potential precedence issue in the multiple || clauses
(there can't really *be* an issue here because we only ever test the
same variable, in, on both sides of the && clause, but should that
change in the future then there's the chance a bug could appear)

w/r/t: a4c5058392e7a0d4628f9f2206e943099893e1ad

### DIFF
--- a/src/strings/normalize.c
+++ b/src/strings/normalize.c
@@ -470,7 +470,7 @@ static void canonical_composition(MVMThreadContext *tc, MVMNormalizer *n, MVMint
  * the handling of breaking around controls much earlier, so don't have to
  * consider that case. */
 static MVMint32 maybe_hangul(MVMCodepoint cp) {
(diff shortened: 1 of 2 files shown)


leejo:
Apparently these warnings have been fixed, closing.

PR #294, opened 2015-11-15
Add size_t and bool native type, bool is standard in C99

Skarsnik:
see for rakudo https://github.com/rakudo/rakudo/pull/587

### DIFF
--- a/src/6model/reprs/P6int.c
+++ b/src/6model/reprs/P6int.c
@@ -1,3 +1,5 @@
+#include <stdbool.>
+#include <stdref.h>
 #include "moar.h"
 
 /* This representation's function pointer table. */
@@ -116,7 +118,7 @@ static void compose(MVMThreadContext *tc, MVMSTable *st, MVMObject *info_hash) {
                 case MVM_P6INT_C_TYPE_LONG:     repr_data->bits = 8 * sizeof(long);      break;
                 case MVM_P6INT_C_TYPE_LONGLONG: repr_data->bits = 8 * sizeof(long long); break;
                 case MVM_P6INT_C_TYPE_SIZE_T: repr_data->bits = 8 * sizeof(size_t); break;
-                case MVM_P6INT_C_TYPE_BOOL: repr_data->bits = 8 * sizeof(_Bool); break;
+                case MVM_P6INT_C_TYPE_BOOL: repr_data->bits = 8 * sizeof(bool); break;
             }
 
             if (repr_data->bits !=  1 && repr_data->bits !=  2 && repr_data->bits !=  4 && repr_data->bits != 8
--- a/src/6model/reprs/P6int.c
+++ b/src/6model/reprs/P6int.c
@@ -115,6 +115,8 @@ static void compose(MVMThreadContext *tc, MVMSTable *st, MVMObject *info_hash) {
                 case MVM_P6INT_C_TYPE_INT:      repr_data->bits = 8 * sizeof(int);       break;
                 case MVM_P6INT_C_TYPE_LONG:     repr_data->bits = 8 * sizeof(long);      break;
                 case MVM_P6INT_C_TYPE_LONGLONG: repr_data->bits = 8 * sizeof(long long); break;
+                case MVM_P6INT_C_TYPE_SIZE_T: repr_data->bits = 8 * sizeof(size_t); break;
+                case MVM_P6INT_C_TYPE_BOOL: repr_data->bits = 8 * sizeof(_Bool); break;
             }
 
             if (repr_data->bits !=  1 && repr_data->bits !=  2 && repr_data->bits !=  4 && repr_data->bits != 8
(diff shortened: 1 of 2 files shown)


FROGGS:
What are we going to do on MSVC, where `_Bool` does not seem to be available?
Also, stackoverflow seems to suggest that we have to include `stdbool.h`.

Skarsnik:
It was me being lazy on this part. Yes including stdbool.h and replace _Bool with bool should work. I was assuming every compiler as a _Bool internal type and bool was always a macro to it.

Skarsnik:
I was also wondering if I should add a cenum or a real work should be made on it to map C enum to Perl enums. From what I gathered enum should be int32 in size, but can vary depending on the compilers (and compiler option)

FROGGS:
@Skarsnik you've got a type in the first line of the latest patch.
And, what do we need `stdref.h` for?

This check reports that MSVC 2012 does neither support `_Bool` nor `bool`: https://gist.github.com/FROGGS/647950041e5b920c0f99

FROGGS:
Btw, yes, we should try to get better enum support into rakudo somehow.

FROGGS:
From reading stackoverflow it seems that MSVC 2013 and better come with `stdbool.h`.
Though, what are we going to do with compilers that don't support it?

Skarsnik:
stdref.h is for size_t. It should probably not be needed since lot of stuff include it (it work without it)

Hm, tricky question for MSVC2012 or older compilater. On gcc (x86-32 and amd64) the size is the size of unsigned char (4 bytes). It should probably be different for each plateform. I think for MSVC<2013 we should return the size MSVC2013 use for bool. Since a binding that need the bool type is probably using a library compiled with MSVC2013++

FROGGS:
I'm still hesitating to merge it.

I think we need to probe for the presence/usability of stdbool.h, stdref.h, _Bool, bool and size_t, and provide sensible defaults, whatever that means per platform.

Skarsnik:
size_t is standard in ISO C. I can't find the proper date  or rfc, but I think I remember in 2002 dealing with only one old arch that did not have it.

For bool I can maybe workaround with: 
__STDC_VERSION__ and __STDC__
The issue is how to fallback: I can fallback to sizeof(unsigned char). But the test will fall, I need to write a separate test file for bool since it will not compile. But the test falling could be warning for futur developpers to check. 
Or it can be totatly different with setting a dumb size for it and have a warning when someone use bool and say that perl6 has be compiled with a compiler that don't support this type and it's likely to fail.

I can separed the two modifications for now (If I figure git magic x)). size_t is really important (even it can be replaced by OpaquePointer for struct size issue) since in my case it's not the same on x86-32bit and 64bits, but you lost the value.

Skarsnik:
Ok for bool what do think of this:

I add the proper check somewhere (it's in probe.pm or something? of directly check the right macro fo c99). fall back to a dumb value (0-1) that will make the test for bool fail (the c part of the test will fall back to like 42). and put a message that incist people to find a solution for their environement?
An alternate solution is just having NC.pm see that the size of bool is a dumb value and warn the user that the bool type is likely to not work on their environement.

You have to remember that the main goal is for NC/Perl6 to use a lib that is C99 compiled

FROGGS:
I dont see that the main goal is to load C99 libs only. Especially when it comes to cross-platformability the libs are often enough C89 compliant.

Give me some time and I figure out a way to not break MoarVM on windows.

Skarsnik:
I mean, to be able to use a C99 compiled library. Like I wrote a binding for https://github.com/google/gumbo-parser and it has bool in some of its type (for example)

cygx:
@FROGGS As Skarsnik mentioned, the issue is that currently, you're unable to call into libraries that use  `_Bool` in function signatures without knowing about the boolean representation onthat particular platform so you can substitute a properly sized integer type instead.

What I would do is create a platform header for `stdbool.h`. The header *should*  be available on all platforms we care about (it's been there since gcc 2.95) except for WIndows with older versions of MSVC.

So just check `_MSC_VER` and if it's below the appropriate value, mock up your own `stdbool.h` that typedefs `bool` as `unsigned char` (the MSVC C++ `bool` type is 1 byte wide) and unconditionally include `stdbool.h` otherwise.

Skarsnik:
I added your verification

FROGGS:
@cygx: I don't like these checks, when can probably do better...

What if we would add a check for the usability of `stdbool.h` and the presence of _Bool and bool like I proposed?
Then we can still fall back to what you posted.
This will mean that we do not have to add more platform/compiler conditions to your approach when we want to compile on HPUX or other weird platforms.

cygx:
@FROGGS HPUX comes with a C99 compiler. Feel free to come up with a better check, but as far as I'm aware, You Aren't Gonna Need It...

Skarsnik:
There is a _STDC_VERSION_ macro to check for C99 but I guess that bool can be implemented prior to that. 

I think it's better that it fail to compile on an unknow environement because the failback to unsigned char as a size could be wrong. I mean it could also fail back as a dumb size and NC could warn the user that using bool on this environement is hazardous.

But the real question is: Should we care about old plateform for a "modern" language?

FROGGS:
I implemented it as https://github.com/MoarVM/MoarVM/commit/df3bbce97f

Thanks!

PR #301, opened 2015-11-18
Implement a istty_fh op.

pmurias:


timo:
please add the new op before the first spesh op, rather than at the very end

jnthn:
Squashed/merged it manually.

PR #310, opened 2015-12-09
Refactoring conditional directives.

wmarquesr:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

* [Linux kernel coding style](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892)
* [Living in the #ifdef Hell](https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/)

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/src/gc/collect.c
+++ b/src/gc/collect.c
@@ -604,6 +604,7 @@ void MVM_gc_collect_free_gen2_unmarked(MVMThreadContext *tc, MVMint32 global_des
     MVMGen2Allocator *gen2 = tc->gen2;
     MVMuint32 bin, obj_size, page, i;
     char ***freelist_insert_pos;
+    int col_in_directives;
     for (bin = 0; bin < MVM_GEN2_BINS; bin++) {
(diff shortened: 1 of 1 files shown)


zhuomingliang:
Build failed.

wmarquesr:
But the code is note wrong.

zhuomingliang:
Since it doesn't pass travis-ci, I'm closing it. Please feel free to push a new one that passes travis-ci, thanks.

jnthn:
For me variables and the state they introduce, which needs to be understood
to understand the code in question, are a bigger maintenance issue. This
patch introduces a variable at the top level function scope, when it seems
it could have been more narrowly scoped. So I'm not especially convinced
the patch - even if it had compiled on the build server - was an overall
improvement.

wmarquesr:
@jnthn I do not understand whether you agree or not with the patch, if so, should I submit again and pass the tests, would be accepted?

PR #319, opened 2015-12-21
RFC: make it possible to intermix binary and text mode syncfile access

cygx:
Not sure all edge-cases are covered (flush?), but should be functional.


timo:
now that text mode file access has been completely thrown out, this PR is obsolete, i believe. if not, feel free to re-open and we'll figure out what can be kept

PR #327, opened 2016-01-04
Fix a gc bug in string concatenation.

pmurias:
Add a MVMROOT that's needed when concatenating strings that consist of a large amount of strands.


FROGGS:
Was the libuv submodule change intentional? Besides that, the patch is correct IMO.

pmurias:
Made a new pull request at https://github.com/MoarVM/MoarVM/pull/328

PR #329, opened 2016-01-04
[readme] declare the supported Unicode version

patch:
This is important to clearly display to users and I had a hard time finding it.

I found it in commit 7575ca4b and docs/ChangeLog for 2015.07.


jnthn:
Might I also suggest tweaking the Unicode database update script to spit out a reminder to update this at the end of its output, otherwise it *will* be forgotten. :-)

PR #331, opened 2016-01-07
Add subsecond file time ops stat_time and lstat_time

ilmari:


zhuomingliang:
Since we have `lstat` and `stat` op, what's the reason that we have to add this new op? Is it tv_nsec necessary for Perl 6 core setting? if not, a module for Perl 6 may be better.

ilmari:
@zhuomingliang This is to enable IO::Path.modified and friends to return Instants with subsecond precision, which libuv already gives us, see rakudo/rakudo#676.

Issue #335, opened 2016-01-21
VS2015 on Windows 10:  Cannot open include file: 'msinttypes/inttypes.h'


nanis:
I am trying to build from a fresh git clone (last commit is 775271df). I get this when I run `nmake test` after `perl Configure.pl ...`

Looking at:

    #include <msinttypes/inttypes.h>

the angle-bracket seems to indicate this file is presumed to be either bundled by some VC version, or it should be in the include paths added by the rakudo build process.

AFAIK, full `inttypes.h` support was added in VS2013. Therefore, it seems to me, unless there is some other reason for the gymnastics, the `#ifdef` in `/include/moar\platform/inttypes.h` needs to be replaced with a version check.

If I can figure out the original reason for this, I'll put together a patch, but, in the mean time, FYI.

I am confused by the fact that this does not seem to be a recent addition, but it never caused problems with my previous VS2015 builds.

On the other hand, now I have found why I am getting warnings for using `int` formats to print `size_t` values when building a 64-bit version on Windows.

nanis:
As far as I can see, the only relevant change is in 940850e6 -- which may or may not have something to do with some of the *unrecognized option* messages that scrolled by during the build. It would be nice if the build process took advantage of `File::Spec`.

FROGGS:
I bet your clone is not up to date, since that revision is unkown to git. Also, the referenced commit is supposed to fix what you are experiencing now, so I think your codebase does not include that commit.

FROGGS:
Umm, actually, I'm not sure what the problem report is about. :o)

you say " I get this when I run nmake test after perl Configure.pl ...", but what *do* you get?

nanis:
Actually ... I was confused and pasted the commit hash from rakudo, not MoarVM. You are right, the fresh rakudo clone had pulled in 8079ca54aa.

So, I guess that takes care of the install problem. However, the unconditional delegation to `msinttypes` and definition of the `size_t` format are still wrong for VS2013 and later. I hope my PR fixes this.

nanis:
I did a fresh clone of `rakudo`, ran `perl Configure.pl ...`, and then did an `nmake test` which gave me *** Cannot open include file: 'msinttypes/inttypes.h'*** At that point, looking around in the MoarVM that was pulled in by the rakudo build process, I did not realize I was not looking at the most recent MoarVM source, and, to make matters worse, ran `git log` in the wrong subdirectory.

[...]

PR #336, opened 2016-01-21
According to [MS][1], VS2013 and later provide C99 inttypes.h

nanis:
This is related to #335. 

[1]: https://msdn.microsoft.com/library/hh409293%28v=vs.120%29.aspx

### DIFF
--- a/src/platform/inttypes.h
+++ b/src/platform/inttypes.h
@@ -1,4 +1,4 @@
-#ifdef _MSC_VER
+#if _MSC_VER < 1800
 #include <msinttypes/inttypes.h>
 /* Print size_t values. */
 #define MVM_PRSz "Iu"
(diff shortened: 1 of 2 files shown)


nanis:
Issue #335 was due to my confusion. However, it remains true that VS2013 and later provide `inttypes.h` and `stdint.h` which should be used in preference. Also, this avoids the incorrect definition of the `size_t` `printf` format specifier in those cases.

nanis:
It looks like I botched this ... Measure twice, cut once and all. Apologies for the noise.

I'll close this, and submit a new one when I can actually focus, instead of switching among *N* unrelated tasks.

If anyone else wants to fix my horrible fix, we need a defined guard for `_MSC_VER` before checking its value. *sigh*

PR #340, opened 2016-02-11
build/probe.pm: Use system Perl 5 for some values instead of GCC

flussence:
I came across [a bizarre failure](https://irclog.perlgeek.de/moarvm/2016-02-10#i_12021848);
it turned out to be unfortunate interaction of toolchain pieces on one machine,
so I can work around it for now, but it got me wondering why moar's
Configure.pl is trying so hard to reinvent the ugly parts of autotools.

This PR just asks the Perl 5 that's actually running the Configure script for
the answers. If it turns out to break something, it's easily reverted.

### DIFF
--- a/build/probe.pm
+++ b/build/probe.pm
@@ -310,30 +310,12 @@ EOT
 
 sub C_type_bool {
     my ($config) = @_;
-    my $restore = _to_probe_dir();
-    my $template = <<'EOT';
-#include <stdio.h>
-#include <stdlib.h>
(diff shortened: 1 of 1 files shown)


FROGGS:
I clearly vote against this PR.
We intentionally do not use Perl 5's Config, because this does not always report correct values.

The only way to check for certain system details is actually to check them. And that's what we do.
If our probes are not working correctly, we should fix them in the right way.

PR #347, opened 2016-03-22
A proposed fix for RT#127756

salortiz:
See https://rt.perl.org/Ticket/Display.html?id=127756


salortiz:
If the slots contains objects or strings, they are gc marked, so is naive to simply zeroed them.
 :-1:  to me.

PR #356, opened 2016-04-08
adjust byte count when UTF-16 BOM marker consumed

dwarring:
Something I noticed in Rakudo, when experimenting with
UTF-16 BOM markers. For example:

    % perl6 -e'my $s = Buf.new([255, 254, 72, 0, 101, 0]).decode("utf-16"); say $s.perl'
    "He\0"

The 2-byte "\xFF\FE" BOM marker is consumed, but the byte count hasn't been adjusted, so
it's running past the end of the string.


jnthn:
Added a test in S32-str/encode.t. Thanks! :cake:

Issue #56, opened 2013-09-11
Change `MVMString*` to `char*` where it makes sense


gerdr:
Some immutable things like `MVMREPROps` could be compile-time constants (as in C static variables) if they did not contain `MVMString*` members (which are also less convenient for debugging).

It would make sense to change those to `const char*` and look up the runtime strings in a hash that hangs off instance. They probably could even be hashed by their address instead of their character data, making lookup reasonably fast.

The REPR names for REPR `Xxx` could be added to the hash by the corresponding `MVMXxx_initalize()` or lazily on access.

The name of an `MVMContainerSpec` is another candidate, and there are probably others.

PR #357, opened 2016-04-08
Workaround tommath issue #56, affects random bigint numbers > 32bits

skids:
See rakudo RT#109586


skids:
Totally untested (up way past bedtime) and needs review.

skids:
Managed to squeeze off a rakudo and nqp test run before bed; they were OK.  Only tested on 64-bit linux.  Needs testing on 32-bit.

jnthn:
I don't have a 32-bit thingy handy to try it out on, but I'm happy enough with the patch otherwise.

jnthn:
Any reports from 32-bit on this?

FROGGS:
The tests pass before and after applying the changes on x86.

PR #361, opened 2016-04-17
Prohibit adding a non-serializable object to an SC

niner:
This would blow up later when trying to serialize the SC. By catching this
early, we give e.g. try_add_to_sc the chance to catch this error early
enough.


niner:
Retracting as it seems to cause regressions.

PR #362, opened 2016-04-22
Fix segfault when composing an uncomposed P6opaque repr.

pmurias:


vendethiel:
shouldn't you check for both?

pmurias:
No, if REPR_data is set (it can happen by either composing or deserialization) the name_to_index_mapping is always set.

vendethiel:
okay, thanks.

PR #364, opened 2016-05-04
	Implement loadbytecodebuffer OP

niner:


src/core/loadbytecode.c
```diff
@@ -7,6 +7,43 @@ static void run_load(MVMThreadContext *tc, void *sr_data);
 static void mark_sr_data(MVMThreadContext *tc, MVMFrame *frame, MVMGCWorklist *worklist) {
     MVM_gc_worklist_add(tc, worklist, &frame->special_return_data);
 }
+void MVM_run_comp_unit(MVMThreadContext *tc, MVMCompUnit *cu) {
```
jnthn: If this is just for internal use (within this file), it should be marked `static`, and the convention is such things don't get an `MVM_` prefix.

src/core/loadbytecode.c
```diff
@@ -7,6 +7,43 @@ static void run_load(MVMThreadContext *tc, void *sr_data);
...
+    MVMuint32    data_size;
+
+    /* Ensure the source is in the correct form. */
+    if (!IS_CONCRETE(buf) || REPR(buf)->ID != MVM_REPR_ID_MVMArray)
+        MVM_exception_throw_adhoc(tc, "loadbytecodebuffer requires a native array to read from");
+
+    /* MVMCompUnit expects the data to be non-GC managed as it usually comes straight from a file */
+    data_size = ((MVMArray *)buf)->body.elems;
+    data_start = MVM_malloc(data_size);
+    memcpy(data_start, (MVMuint8 *)(((MVMArray *)buf)->body.slots.i8 + ((MVMArray *)buf)->body.start), data_size);
```
jnthn: Should also check that the the buffer's element type is 8-bit integers. It could easily be something else.

niner:
Updated the PR with a new commit bringing back the mmap optimization

src/core/compunit.c
```diff
@@ -67,6 +67,35 @@ MVMCompUnit * MVM_cu_map_from_file(MVMThreadContext *tc, const char *filename) {
...
+    }
+
+    size = req.statbuf.st_size;
+
+    if ((block = MVM_platform_map_file(fd, &handle, (size_t)size, 0)) == NULL) {
+        /* FIXME: check errno or GetLastError() */
+        MVM_exception_throw_adhoc(tc, "Could not map file into memory: %s", "FIXME");
+    }
+
+    block += pos;
```
jnthn: Do we somewhere validate that `pos` is in range?

src/core/loadbytecode.c
```diff
@@ -7,6 +7,47 @@ static void run_load(MVMThreadContext *tc, void *sr_data);
...
+void MVM_load_bytecode_buffer(MVMThreadContext *tc, MVMObject *buf) {
+    MVMCompUnit *cu;
+    MVMuint8    *data_start;
+    MVMuint32    data_size;
+
+    /* Ensure the source is in the correct form. */
+    if (
+        !IS_CONCRETE(buf)
+        || REPR(buf)->ID != MVM_REPR_ID_MVMArray
+        || ((MVMArrayREPRData *)STABLE(buf)->REPR_data)->slot_type != MVM_ARRAY_U8
```
jnthn: I guess I8 would be permissible too.

src/core/loadbytecode.c
```diff
@@ -7,6 +7,47 @@ static void run_load(MVMThreadContext *tc, void *sr_data);
...
+    )
+        MVM_exception_throw_adhoc(tc, "loadbytecodebuffer requires a native array to read from");
```
jnthn: Maybe mention the 8-bit requirement in the error

src/core/loadbytecode.c
```diff
@@ -60,6 +85,26 @@ void MVM_load_bytecode(MVMThreadContext *tc, MVMString *filename) {
...
+
+    MVMROOT(tc, filename, {
+        MVMuint64 pos = MVM_io_tell(tc, oshandle);
+        cu = MVM_cu_map_from_file_handle(tc, MVM_io_fileno(tc, oshandle), pos);
+        cu->body.filename = filename;
+
+        run_comp_unit(tc, cu);
+    });
+
+LEAVE:
```
jnthn: unused label?

src/core/compunit.c
jnthn: Ah, I see we get it from `tell` below.

PR #368, opened 2016-05-14
Fix temp root stack underflow with exceptions in NC callbacks

niner:
MVM_gc_root_temp_mark marks the temporary root stack at its current height as
the limit for removing all roots. At this point all MVMROOTs have to be in
place. Otherwise a MVM_gc_root_temp_pop_all called by
MVM_exception_throw_adhoc_free_va will remove not only the temp roots installed
by the nested run loop, but also frame roots leading to a possible temp root
underflow later on.


jnthn:
Nice catch! Note that `nativecall_libffi.c` also needs a patch like this too. (Which probably means we need to factor stuff out a bit better...)

niner:
Thanks for the reminder! Added a second commit with a ported fix. And yes, that's much copy&pasted code...

PR #369, opened 2016-05-15
Update more serialization from int32 to varint.

zhuomingliang:
This reduces rakudo core setting size about 512 KB.

Note: This is broken somewhere and I don't know where I should take a look or fix it.


zhuomingliang:
Merged.

PR #370, opened 2016-05-18
Serialize hll role.

pmurias:


src/6model/serialization.c
```diff
@@ -10,7 +10,7 @@
 
 /* Version of the serialization format that we are currently at and lowest
  * version we support. */
-#define CURRENT_VERSION 19
+#define CURRENT_VERSION 20
```
zhuomingliang: I  think it don't need bump it since you didn't do any version check, and looks like it is safe.

pmurias: Isn't the purpose of the serialization format to prevent stuff being
deserialized by old versions of MoarVM?

zhuomingliang: Yes, but only need when you added version check code. like this: https://github.com/pmurias/MoarVM/blob/ae8d3838635a8443f1fba0e95b2ade541cd6902c/src/6model/serialization.c#L1657

pmurias: The policy on versioning seems to be: "When making any change that would render a version of MoarVM that doesn't know about a change unable to successfully/completely deserialize things".

Code from before the pull request won't understand the newly added flag and as such fail to deserialize properly.

zhuomingliang: I don't think it will fail to deserialize properly since you didn't add any version check code, and no STABLE_HAS_HLL_ROLE flag in the old bytecode files.

zhuomingliang: Well, it is easy to test. If you don't bump the version, and you can still build nqp, then it's fine.

jnthn:
The version bump is required. Consider a serialized STable with a HLL role, serialized after this change. An older version of MoarVM that didn't recognize the new flag would end up reading the HLL role value as it if was, for example, part of a parametric type descriptor (simply because that is what comes next).

PR #376, opened 2016-06-11
Fix for RT #128306

MasterDuke17:
Converts Unicode codepoints to their numeric value


jnthn:
Thanks for looking into the bug. While this patch may indeed fix the issue in question, it's the wrong place to fix it. The ASCII encoder should not be magically transforming input characters. A more likely fix is for the code processing named arguments in `Actions.nqp` in Rakudo to use a way of processing numbers that can properly handle other Unicode digits, such as the `nqp::radix` op.

PR #383, opened 2016-07-07
Smoke me/spaceybuild

coke:
Fix the issue with a space in the path during build.

Old branch, smoked but never merged, making pull request so we don't lose it.


coke:
Not quite baked yet: If I build in ~/sandbox/Moar\ VM/ , make install still makes a ~/sandbox/Moar directory and then installs into ./VM

PR #388, opened 2016-08-02
Fix wrong EOF detect for /proc file.

Gnouc:
Fixes RT #127370


jnthn:
Looks reasonable; thanks!

PR #390, opened 2016-08-17
Remove keep_caller from MVMFrame

LemonBoy:
Fixes some cases (RT #128803) where the backtraces would be abruptly
truncated due to an optimization introduced in cecf5729.
As suggested by @jnthn [1] the best solution is to simply revert the
optimization as it's not worth it anymore.

Please keep in mind that at the moment this change makes the spectest suite fail when doing the `t/spec/integration/error-reporting.rakudo.moar` file as it contains a test for the [RT #12745](https://rt.perl.org/Public/Bug/Display.html?id=127425).
This change obsoletes the [https://github.com/rakudo/rakudo/commit/67b6544e48](commit referenced in the ticket) and renders the test useless as @zoffixznet pointed out.

[1] http://irclog.perlgeek.de/moarvm/2016-08-17#i_13040630


bdw:
Looks good enough to me. I'll merge it. The test will have to be removed or changed.

PR #392, opened 2016-08-18
Correctly NULL-terminate the buffer in MVM_vm_dump_file.

LemonBoy:


jnthn:
I read that as `o[1]` instead of `o[l]` for a moment and was thinking "how could that possibly work?!" :-) Looks good; thanks!

PR #397, opened 2016-08-23
Fix mvm_tell for files that haven't been read from.

LemonBoy:
Not 100% sure about the exception message though,


zhuomingliang:
Not sure about it, didn't it return 0 for files that haven't been read from?

LemonBoy:
The user could've only written something to the file and in that case `ds` would still be NULL

jnthn:
Seems reasonable; can see the hesitation on the error, but it mentions the operation the user was trying to do (a tell) that failed, so I think it's OK.

PR #401, opened 2016-09-06
Remove unused variables

paultcochrane:
Remove as many unused variables as found by the `-Wunused-variable` GCC compiler flag as possible.  This doesn't remove all such warnings since the `copy_to()` function in `CUnion.c` is currently NYI, and the automatically generated output in `unicode.c` contains in one place an unused `size` variable.

I've tried to slice the commits as thinly as possible so that they can be cherry-picked if so desired.

If I can improve upon the patches please let me know and I'll resubmit the PR.


jnthn:
All seem reasonable to me; thanks.

PR #402, opened 2016-09-06
Remove parentheses warnings in gcc

paultcochrane:
This pull request removes the parentheses issues highlighted by the `-Wparentheses` gcc warning option (on by default in `-Wall`).  It adds the `-Wparentheses` option to the list of gcc warnings flags so that such issues don't repeat themselves in the future.

If I can improve upon the PR please let me know and I'll update it and resubmit.

### DIFF
--- a/src/6model/reprs/MVMStaticFrame.c
+++ b/src/6model/reprs/MVMStaticFrame.c
@@ -361,7 +361,8 @@ static void describe_refs(MVMThreadContext *tc, MVMHeapSnapshotState *ss, MVMSTa
             if (body->spesh_candidates[i].sg) {
                 MVMCollectable **c_ptr;
                 MVM_spesh_graph_mark(tc, body->spesh_candidates[i].sg, ss->gcwl);
-                while (c_ptr = MVM_gc_worklist_get(tc, ss->gcwl)) {
+                c_ptr = MVM_gc_worklist_get(tc, ss->gcwl);
+                while (c_ptr) {
                     MVMCollectable *c = *c_ptr;
                     MVM_profile_heap_add_collectable_rel_const_cstr(tc, ss, c,
                         "Object held by spesh graph");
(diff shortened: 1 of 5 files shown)


src/6model/reprs/MVMStaticFrame.c
```diff
@@ -361,7 +361,8 @@ static void describe_refs(MVMThreadContext *tc, MVMHeapSnapshotState *ss, MVMSTa
             if (body->spesh_candidates[i].sg) {
                 MVMCollectable **c_ptr;
                 MVM_spesh_graph_mark(tc, body->spesh_candidates[i].sg, ss->gcwl);
-                while (c_ptr = MVM_gc_worklist_get(tc, ss->gcwl)) {
+                c_ptr = MVM_gc_worklist_get(tc, ss->gcwl);
+                while (c_ptr) {
```
vendethiel: those completly change the meaning of those `while` (executed once vs executed at each iteration).

I think the correct way to fix the assignment warning is to add a pair of parentheses `while ((a = 1))` for GCC & clang.

jnthn:
Argh, I explicitly disabled this warning on clang where it's a default one. Knowing that && is tighter than || is up there with knowing that * is tighter than +. And anyone feeling it ain't is in the worst place to submit patches changing it. :-)

Short of obviously OK stuff like removing dead code, I'm generally not fond of patches submitted purely for "cleanup". Folks fixing bugs, adding features, or doing optimizations are welcome and encouraged to leave the code in that area cleaner than they found it. They're doing so off the back of having gained an understanding of what's going on. Patches for the sake of cleanup are risky. Had this one got in, we'd have had two bugs in heap profiling, and they would not have come to light easily. Given the generally good code quality of MoarVM, the risk outweighs the gain to me.

In summary: thanks for trying to help, and sorry there was a misleading todo list hanging around, I'll reject this one, because it does harm, without doing good.

paultcochrane:
That's good to know.  Thanks for the clear explanation and comment about policy!

PR #405, opened 2016-09-09
Optimize the check for negative bignums.

LemonBoy:
We save a function call for every check.

### DIFF
--- a/src/math/bigintops.c
+++ b/src/math/bigintops.c
@@ -185,7 +185,7 @@ static void store_int64_result(MVMP6bigintBody *body, MVMint64 result) {
 static void store_bigint_result(MVMP6bigintBody *body, mp_int *i) {
     if (can_be_smallint(i)) {
         body->u.smallint.flag = MVM_BIGINT_32_FLAG;
-        body->u.smallint.value = SIGN(i) ? -DIGIT(i, 0) : DIGIT(i, 0);
+        body->u.smallint.value = SIGN(i) == MP_NEG ? -DIGIT(i, 0) : DIGIT(i, 0);
         mp_clear(i);
         MVM_free(i);
     }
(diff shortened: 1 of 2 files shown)


src/math/bigintops.c
```diff
@@ -543,13 +543,13 @@ MVMObject *MVM_bigint_div(MVMThreadContext *tc, MVMObject *result_type, MVMObjec
     bc = get_bigint_body(tc, result);
 
     if (MVM_BIGINT_IS_BIG(ba)) {
-        cmp_a = mp_cmp_d(ba->u.bigint, 0);
+        cmp_a = SIGN(ba->u.bigint) == MP_NEG ? MP_LT : MP_GT;
     } else {
         // we only care about MP_LT or !MP_LT, so we give MP_GT even for 0.
         cmp_a = ba->u.smallint.value < 0 ? MP_LT : MP_GT;
```
LemonBoy: We could probably make this code use `MP_NEG` and `MP_ZPOS` instead

jnthn:
@LemonBoy any reason to close this besides "too slow to get review"? Finally got around to looking at it now, and it looks fine.

LemonBoy:
None in particular, it got killed in a housekeeping spree :)
You can easily grab the [patch](https://patch-diff.githubusercontent.com/raw/MoarVM/MoarVM/pull/405.patch) and apply it using `git` though, I'd be happy to see this applied.

MasterDuke17:
I seem to recall this was related to https://rt.perl.org/Ticket/Display.html?id=128035, was the patch actually applied?

jnthn:
D'oh, I forgot. It's applied now, though it was a welcome optimization rather than a behavior change, so no impact on the RT in question.

PR #406, opened 2016-09-10
Make MVM_file_open_fh() throw if the file we opened was a directory

cygx:
This means `nqp::open()` will now throw if we pass in the path to a directory, making the check in `src/core/IO/Handle.pm` superfluous.

Getting rid of that check is the Right Thing To Do if we care for atomicity of file system operations, but getting an exception of type `X::IO::Directory` thrown would need some additional fiddling...

### DIFF
--- a/src/io/syncfile.c
+++ b/src/io/syncfile.c
@@ -470,27 +470,31 @@ static int resolve_open_mode(int *flag, const char *cp) {
 
 /* Opens a file, returning a synchronous file handle. */
 MVMObject * MVM_file_open_fh(MVMThreadContext *tc, MVMString *filename, MVMString *mode) {
-    char          * const fname  = MVM_string_utf8_c8_encode_C_string(tc, filename);
-    char          * const fmode  = MVM_string_utf8_encode_C_string(tc, mode);
-    MVMOSHandle   * const result = (MVMOSHandle *)MVM_repr_alloc_init(tc, tc->instance->boot_types.BOOTIO);
-    MVMIOFileData * const data   = MVM_calloc(1, sizeof(MVMIOFileData));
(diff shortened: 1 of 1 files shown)


src/io/syncfile.c
```diff
@@ -492,6 +492,18 @@ MVMObject * MVM_file_open_fh(MVMThreadContext *tc, MVMString *filename, MVMStrin
         MVM_exception_throw_adhoc_free(tc, waste, "Failed to open file %s: %s", fname, uv_strerror(req.result));
     }
 
+    /* Check that we didn't open a directory by accident. If stat fails, just move on. */
```
niner: Why is it safe to move on if stat fails?

niner:
This would not only make open more atomic, but also remove a rather costly check on each file open.

src/io/syncfile.c
cygx: We don't really have a way to check if a failure to call `fstat()` indicates an inability to read from/write to the file. `ENOMEM` (not enough kernel memoy) and `EOVERFLOW` aside, as far as I can see, the documented failure modes on linux should already have triggered when opening the file. So if such a failure after a successful call to `open()` happens at all, we're dealing with a truly exceptional situation. As we can't really do anything about it, we just move on and wait for the next IO operation to fail or succeed as it might.

cygx:
Also note that various code paths leak memory (`fmode`, and `data` need to be freed and we need to call `uv_fs_req_cleanup()` on req).

src/io/syncfile.c
niner: That's what I'd just like to read in the comment :)

cygx: I also managed to introduce a bug by trying to close `data->fd` instead of `fd` :(

I'll add the comment and fix the other stuff after lunch.

PR #408, opened 2016-09-17
Add error message for likely macOS build failure

labster:
On a recent Rakudo install, I encountered a build failure that is due to this happening after an Xcode upgrade:

```shell
brent@ragnar ~/c/rakudo (nom)> gcc


Agreeing to the Xcode/iOS license requires admin privileges, please re-run as root via sudo.


brent@ragnar ~/c/rakudo (nom)>
```

So, in the pursuit of awesome error messages, this.


jnthn:
The check fail is just that one of the builds timed out while doing a git clone.

PR #416, opened 2016-09-30
Die on CStruct without any fields.

peschwa:
I'm personally not quite happy with info_alloc now existing as purely an alias to num_attrs, but it's probably semantically clearer that way.

In any case, purpose of this patch is to stop cheating when the user does something that doesn't really make sense. The equivalent patch on nqp-j locally also has the class name, but I haven't figured out how we do that on MoarVM.

### DIFF
--- a/src/6model/reprs/CStruct.c
+++ b/src/6model/reprs/CStruct.c
@@ -133,7 +133,7 @@ static void compute_allocation_strategy(MVMThreadContext *tc, MVMObject *repr_in
         MVMint32 i;
 
         if (info_alloc == 0)
-            MVM_exception_throw_adhoc(tc, "Class has repr CStruct but no fields");
+            MVM_exception_throw_adhoc(tc, "Class %s has no attributes, which is illegal with the CStruct representation.", st->debug_name);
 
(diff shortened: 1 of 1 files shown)


src/6model/reprs/CStruct.c
```diff
@@ -127,11 +127,14 @@ static void compute_allocation_strategy(MVMThreadContext *tc, MVMObject *repr_in
...
         /* Get number of attributes and set up various counters. */
         MVMint32 num_attrs        = MVM_repr_elems(tc, flat_list);
-        MVMint32 info_alloc       = num_attrs == 0 ? 1 : num_attrs;
+        MVMint32 info_alloc       = num_attrs;
         MVMint32 cur_obj_attr     = 0;
         MVMint32 cur_init_slot    = 0;
         MVMint32 i;
 
+        if (info_alloc == 0)
+            MVM_exception_throw_adhoc(tc, "Class has repr CStruct but no fields");
```
jnthn: Which class? :-) `st->debug_name` or so should be useful for including in such an error, though you may need to pass that in from whatever calls this code. Maybe something like "Class %s has no attributes, which is illegal with the CStruct representation." would be a tad clearer also.

PR #434, opened 2016-10-22
Fix the nativecall attribute error messages

titsuki:
The nativecall attribute error messages are missing `num64` in their list.


titsuki:
Thanks !

PR #455, opened 2016-12-15
Fix missing mvmroot and arg_flags allocation

zhuomingliang:


src/6model/6model.c
```diff
@@ -86,14 +91,19 @@ void MVM_6model_find_method(MVMThreadContext *tc, MVMObject *obj, MVMString *nam
 
     /* First try to find it in the cache. If we find it, we have a result.
      * If we don't find it, but the cache is authoritative, then error. */
+    MVM_gc_root_temp_push(tc, (MVMCollectable **)&obj);
+    MVM_gc_root_temp_push(tc, (MVMCollectable **)&name);
```
jnthn: I'd much rather use `MVMROOT` instead.

src/6model/6model.c
```diff
@@ -86,14 +91,19 @@ void MVM_6model_find_method(MVMThreadContext *tc, MVMObject *obj, MVMString *nam
...
     cache = get_method_cache(tc, STABLE(obj));
     if (cache && IS_CONCRETE(cache)) {
         MVMObject *meth = MVM_repr_at_key_o(tc, cache, name);
         if (!MVM_is_null(tc, meth)) {
+            MVM_gc_root_temp_pop_n(tc, 2);
             res->o = meth;
             return;
         }
         if (STABLE(obj)->mode_flags & MVM_METHOD_CACHE_AUTHORITATIVE) {
+            MVM_gc_root_temp_pop_n(tc, 2);
```
jnthn: When we throw an exception and return back to the interpreter in the correct position, then we always pop all temporary roots, so this isn't needed.

src/core/frame.c
```diff
@@ -1070,7 +1070,9 @@ void MVM_frame_capture_inner(MVMThreadContext *tc, MVMObject *code) {
         MVMStaticFrame *sf_outer = code_obj->body.sf->body.outer;
         outer = create_context_only(tc, sf_outer, (MVMObject *)sf_outer->body.static_code, 1);
         MVMROOT(tc, outer, {
-            MVMFrame *outer_outer = autoclose(tc, sf_outer->body.outer);
+            MVMFrame *outer_outer;
+            sf_outer = code_obj->body.sf->body.outer; /* get the new sf_outer again since it misses MVMROOT */
```
jnthn: Why not just MVMROOT it?

Issue #451, opened 2016-12-13
getuniprop_str returns a string of the hex code for the Bidi_Mirroring_Glyph


samcv:
We should either return an int for getuniprop_int or return the actual codepoint character, not a string containing the hex code.

samcv:
This has been fixed since https://github.com/MoarVM/MoarVM/commit/1b779d61

Closing

Issue #465, opened 2016-12-24
Unicode: Bidi_Paired_Bracket_Type and Bidi_Paired_Bracket NYI


samcv:
Neither of these return anything but 0.

PR #468, opened 2016-12-25
[needs rework ] Add 8+ properties and fix several that were working incorrectly

samcv:
Emoji properties
Bidi_Paired_Bracket
Bidi_Paired_Bracket_Type
Bidi_Mirroring_Glyph
Indic_Syllabic_Category
East_Asian_Width
Line_Break
Indic_Positional_Category
Jamo_Short_Name

* Add in all Emoji properties (except Emoji_All which is a derived property) and add Indic_Syllabic_Category
* With this change, Bidi_Mirroring_Glyph now returns a string instead of a
string which contains a hex number.
* The Bidi_Paired_Bracket and Bidi_Paired_Bracket_Type properties have now been added.
* Scrap the old LineBreak sub written in 2012 and use the one we use for enumerated values. The old one had hardcoded a set number of values for this property. Making that change fixes the tests we were failing for uniprop Line_Break in roast

With this change, rakudo.moar now passes all the tests for these three properties, as well as all the ones it was previously passing (aside from the failures for returning a 1 or 0 instead of True or False, which need to be handled on the Rakudo side).


samcv:
This fixes #451 and half of #465

samcv:
Most of these have been added in other commits, but I am going to leave this open as reference until the rest of the values are added.

samcv:
Closing. Stale and not needed anymore.

PR #469, opened 2016-12-27
Fix RT #122471 and #122470 return <control-0000> for \0 and other controls

samcv:
RT: https://rt.perl.org/Ticket/Display.html?id=122471
https://rt.perl.org/Ticket/Display.html?id=122470
We now pass several tests we were not passing before in uniname.t


samcv:
@jnthn you should probably accept this pull before #468, and I will rebase the other one to this.

jnthn:
Looks sensible to me. Thanks! :-)

zoffixznet:
This work appears to have broken some tests: https://gist.github.com/zoffixznet/29301168d1fbbfeb2f003d3c7f515802

Including failures in 6.c-errata: https://gist.github.com/zoffixznet/29f443ead147b43382756278a9892b79

PR #474, opened 2016-12-29
Add unicmp_s op implementing Unicode Collation Algorithm (UCA)

samcv:
Op: `unicmp_s w(int64) r(str) r(str) r(int64) r(int64) r(int64)`

Takes 5 Parameters:
stringa, # string a to compare
stringb, # string b to compare
collationmode (integer), # bitmask
ISO 639 Language code (integer),
ISO 3166 Country code (integer)

The default value for the collation mode is 7 which specifies to sort using Primary, Secondary, and Tertiary. This will sort as most users would expect text to be sorted visually.
The bitmask is as follows:
1 => Unicode Collation Primary
2 => Unicode Collation Secondary
4 => Unicode Collation Tertiary

While the Primary, Secondary and Tertiary mean different things for
different scripts, for the Latin script used in English they mostly
correspond with Primary being Alphabetic, Secondary being Diacritics
and Tertiary being Case.

# Language and Country Codes
The default language and country code is 0, meaning to sort without
respect to country or language.

A setting 0 for language, and country will collate all scripts according to
their own scripts distinctions for Primary, Secondary, and Tertiary, although it
will not take into account certain languages or countries differences in sorting.

For example, some language based differences in collation
“…include ch as in traditional Spanish, ä as in traditional German,
and å as in Danish” ― See http://unicode.org/reports/tr10/ for more details on the UCA

## Country and Language codes details:

   - ISO 3166 defines country numbers which do not change even when a
     country's 2 or 3 letter identifier changes, for example if the country
     split into more than one country, or its name changes. The
     country code will use these numbers and will in the future also provide
     a way to look up numeric codes from three letter identifiers.
   - ISO 639-5 language identifiers do not have a number assigned to them,
     but they are presumed never to change. When implemented an op will be
     added to get MVM's internal language identifier from 3 letter identifiers,
     similar to looking up a MVM Unicode property code. These internal property codes are not
     guaranteed nor expected to remain the same between releases.

## Note
 - Currently only language and country insensitive sorting methods are implemented.
 - We have implemented the language and country insensitive sorting
 using only the Primary level, and if they evaluate as the same on the
 primary level it will fall back to comparing by codepoint. For Latin
 script this will take case into account in most cases. In worst case it
 will not do any worse than the current cmp_s op
- The weight values are accessible by getting the MVM_COLLATION_PRIMARY, MVM_COLLATION_SECONDARY and MVM_COLLATION_TERTIARY properties of a grapheme.
- At the moment we only implement collation by grapheme. The UCA has weights for multiple graphemes, which take precedence over the weights for single graphemes.


jnthn:
I think I'd rather we add the Unicode collation algorithm based comparison a new op, rather than modifying the semantics of the existing `cmp_s` op.

samcv:
Sounds good. What do you think it should be called?

jnthn:
Well, the `uni` prefix we have on various of the Unicode data-based ops is probably worth having on the start of it. `unicmp` ain't bad, though `unicolcmp` makes clearer it's using the collation algorithm. More than just the name, however, we should think about what arguments it should take. So far as I understand it, the Unicode collation algorithm works at a number of different levels, and then there are further ways to customize it. It would be good to get the op interface worked out so we can support those various bits later (or those that we feel belong in the VM), even if for now they just throw due to being NYI.

samcv:
@jnthn pull request is ready for merge as long as everything looks fine.

src/strings/unicode_ops.c
```diff
@@ -1,3 +1,56 @@
+/* Compares two strings, returning -1, 0 or 1 to indicate less than,
+ * equal or greater than based on Unicode Collation Algorithm */
+MVMint64 MVM_unicode_string_compare(MVMThreadContext *tc, MVMString *a, MVMString *b) {
+    MVMStringIndex alen, blen, i, scanlen;
+    // Interation variables
+    MVMGrapheme32 ai;
+    MVMGrapheme32 bi;
+    // Colation order numbers
```
jnthn: Typo: collation

src/strings/unicode_ops.c
```diff
@@ -1,3 +1,56 @@
+/* Compares two strings, returning -1, 0 or 1 to indicate less than,
+ * equal or greater than based on Unicode Collation Algorithm */
+MVMint64 MVM_unicode_string_compare(MVMThreadContext *tc, MVMString *a, MVMString *b) {
+    MVMStringIndex alen, blen, i, scanlen;
+    // Interation variables
```
jnthn: `Int`eration is a good pun given they're integers, but maybe not quite what was meant :-D

src/strings/unicode_ops.c
```diff
@@ -1,3 +1,56 @@
...
+    if (alen == 0)
+        return blen == 0 ? 0 : -1;
+    if (blen == 0)
+        return 1;
+
+    /* Otherwise, need to scan them. */
+    scanlen = alen > blen ? blen : alen;
+    for (i = 0; i < scanlen; i++) {
+        ai = MVM_string_get_grapheme_at_nocheck(tc, a, i);
+        bi = MVM_string_get_grapheme_at_nocheck(tc, b, i);
```
jnthn: This works, though it would be faster on any stand-y string to use the GraphemeIterator instead of indexing.

tools/ucd2c.pl
```diff
@@ -1,12 +1,15 @@
...
 $Data::Dumper::Maxdepth = 1;
 # Make C versions of the Unicode tables.
 
-# Before running, download zip files from http://www.unicode.org/Public/zipped/
-# and extract them in UNIDATA
+# Before running, downloading UNIDATA.zip from http://www.unicode.org/Public/zipped/
+# and extract them into UNIDATA in the root of this repository.
+
+# Download allkeys.txt from http://www.unicode.org/Public/UCA/latest/
+# and place into a folder named UCA under the UNIDATA directory.
```
jnthn: Not important for this PR, but wonder if we should stick a shell script under tools to wget/curl and unzip them...

jnthn: I'm good with that assumption; it's a time-saver for a handful of people, not something that needs to be portable.

src/strings/unicode_ops.c
samcv: We use `MVM_string_get_grapheme_at_nocheck` for the normal string comparing function. Would it be proper for them both to use GraphemeIter instead?

PR #476, opened 2016-12-31
Implement More Unicode 9.0 GCB properties, Prepend, Emoji, and Bidi_Mirroring_Glyph

samcv:
Add support for Grapheme_Cluster_Break=Prepend from Unicode 9.0
These marks were added in Unicode 9.0 which occur *before* the
codepoints they combine with. We had assumed that these controlling
characters would not appear before others and only *after*, but with
this change we now pass all the Prepend tests from GraphemeBreakTest.txt
except for a few odd ones ( line Prepend + a space or newline which
shouldn't occur in real life).

We now check these properties when breaking for graphemes:
Emoji_Modifier
Emoji_Base
E_Base_GAZ
Glue_After_ZWJ

Emoji properties except Emoji_All have been implemented as well.


src/strings/normalize.c
```diff
@@ -507,6 +521,8 @@ static MVMint32 is_spacing_mark(MVMThreadContext *tc, MVMCodepoint cp) {
     }
 }
 static MVMint32 should_break(MVMThreadContext *tc, MVMCodepoint a, MVMCodepoint b) {
+    int GCB_a = MVM_unicode_codepoint_get_property_int(tc, a, MVM_UNICODE_PROPERTY_GRAPHEME_CLUSTER_BREAK);
+    int GCB_b = MVM_unicode_codepoint_get_property_int(tc, b, MVM_UNICODE_PROPERTY_GRAPHEME_CLUSTER_BREAK);
```
zhuomingliang: I would like to call `MVM_unicode_codepoint_get_property_int` before `switch (GCB_a) ` and `switch (GCB_b)` until it is used, which may avoids two calls for fast-path check.

samcv: That can't be done, because both switches rely on the values of GCB_b and GCB_a (1/2 of them do).
I can also remove the hangul check once I add it to the switches, since we're already going to be checking GCB property so we can do it all at once.
I can also move `is_spacing_mark`, and that `is_grapheme_prepend`, since it's tested in the switches though for some reason would not work probably because of when we decide to call `should_break` is optimized for NFG_QC=Yes.

src/strings/normalize.c
```diff
@@ -1,4 +1,14 @@
...
+#define MVM_UNICODE_PROPERTY_GCB_OTHER              0
+#define MVM_UNICODE_PROPERTY_GCB_PREPEND            1
+#define MVM_UNICODE_PROPERTY_GCB_EXTEND             5
+#define MVM_UNICODE_PROPERTY_GCB_REGIONAL_INDICATOR 6
+#define MVM_UNICODE_PROPERTY_GCB_SPACINGMARK        7
+#define MVM_UNICODE_PROPERTY_GCB_E_MODIFIER        14
+#define MVM_UNICODE_PROPERTY_GCB_E_BASE            13
+#define MVM_UNICODE_PROPERTY_GCB_E_BASE_GAZ        17
+#define MVM_UNICODE_PROPERTY_GCB_ZWJ               15
+#define MVM_UNICODE_PROPERTY_GCB_GLUE_AFTER_ZWJ    16
```
jnthn: I'd much rather these were emitted by `ucd2c.pl` into `unicode_gen.h`.

src/strings/normalize.c
```diff
@@ -341,21 +351,24 @@ static MVMint32 is_control_beyond_latin1(MVMThreadContext *tc, MVMCodepoint in)
...
-
-            /* Unassigned is, but only for Default_Ignorable_Code_Point. */
-            if (genprop[1] == 'n') {
-                return MVM_unicode_codepoint_get_property_int(tc, in,
-                    MVM_UNICODE_PROPERTY_DEFAULT_IGNORABLE_CODE_POINT) != 0;
-            }
+        /* Unassigned is, but only for Default_Ignorable_Code_Point. */
+        if (genprop[1] == 'n') {
+            return MVM_unicode_codepoint_get_property_int(tc, in,
+                MVM_UNICODE_PROPERTY_DEFAULT_IGNORABLE_CODE_POINT) != 0;
```
jnthn: I think this code belongs under the `case 'C'` branch of the switch statement? Otherwise we thing that any general category whose second char is an 'n' counts, when originally this only cared for `Cn`?

src/strings/normalize.c
```diff
@@ -528,19 +548,53 @@ static MVMint32 should_break(MVMThreadContext *tc, MVMCodepoint a, MVMCodepoint
...
 
-    /* Don't break between regional indicators. */
-    if (is_regional_indicator(a) && is_regional_indicator(b))
-        return 0;
+    switch (GCB_a) {
+        case MVM_UNICODE_PROPERTY_GCB_REGIONAL_INDICATOR:
+            if ( GCB_b == MVM_UNICODE_PROPERTY_GCB_REGIONAL_INDICATOR )
+                return 0;
+            break;
+        // Don't break after Prepend Grapheme_Cluster_Break=Prepend
```
jnthn: We tend to stick with `/* ... */` style comments.

src/strings/normalize.c
```diff
@@ -578,11 +632,12 @@ static void grapheme_composition(MVMThreadContext *tc, MVMNormalizer *n, MVMint3
  * compute the normalization. */
 MVMint32 MVM_unicode_normalizer_process_codepoint_full(MVMThreadContext *tc, MVMNormalizer *n, MVMCodepoint in, MVMCodepoint *out) {
     MVMint64 qc_in, ccc_in;
-
+    int is_prepend = is_grapheme_prepend(tc, in);
     /* If it's a control character (outside of the range we checked in the
      * fast path) then it's a normalization terminator. */
-    if (in > 0xFF && is_control_beyond_latin1(tc, in))
+    if ( ( in > 0xFF && is_control_beyond_latin1(tc, in) ) && ( !is_prepend ) ) {
```
jnthn: What are the cases when `is_control_beyond_latin1` can return something that is a prepend, out of curiosity?

src/strings/normalize.c
```diff
@@ -619,11 +674,12 @@ MVMint32 MVM_unicode_normalizer_process_codepoint_full(MVMThreadContext *tc, MVM
 
     /* If we didn't pass quick check... */
     if (!qc_in) {
+
         /* If we're composing, then decompose the last thing placed in the
          * buffer, if any. We need to do this since it may have passed
          * quickcheck, but having seen some character that does pass then we
          * must make sure we decomposed the prior passing one too. */
-        if (MVM_NORMALIZE_COMPOSE(n->form) && n->buffer_end != n->buffer_norm_end) {
+        if ((MVM_NORMALIZE_COMPOSE(n->form) && n->buffer_end != n->buffer_norm_end) && !is_prepend) {
```
jnthn: Not sure we need the extra parens? :-)

src/strings/normalize.h
```diff
@@ -75,9 +75,10 @@ MVM_STATIC_INLINE MVMint32 MVM_unicode_normalizer_process_codepoint(MVMThreadCon
      * far in normalized form without having to consider them into the
      * normalization process. The exception is if we're computing NFG, and
      * we got \r, which can form a grapheme in the case of \r\n. */
-    if (in < 0x20 || (in >= 0x7F && in <= 0x9F) || in == 0xAD)
-        if (!(MVM_NORMALIZE_GRAPHEME(n->form) && in == 0x0D))
+    if (in < 0x20 || (in >= 0x7F && in <= 0x9F) || in == 0xAD )
+        if (!(MVM_NORMALIZE_GRAPHEME(n->form) && in == 0x0D)) {
```
jnthn: Just started at this for a while trying to spot what changed... The answer seems to be "nothing worth changing".

tools/ucd2c.pl
```diff
@@ -242,6 +243,24 @@ sub break_property {
...
+            Other => 0,
+            # Make sure we specify the values of Prepend and Extend
+            # We need these consistent since they are used in MoarVM code
+            #Prepend => 1,
+            #Extend => 2,
+            #E_Base => 3,
+            #E_Base_GAZ => 4,
+            #ZWJ        => 5,
+            #E_Modifier => 6,
+            #Glue_After_Zwj => 7
```
jnthn: Looks like this didn't end up happening?

tools/ucd2c.pl
```diff
@@ -697,6 +716,7 @@ sub emit_property_value_lookup {
     my $allocated = shift;
     my $enumtables = "\n\n";
     my $hout = "typedef enum {\n";
+    say Dumper($prop_names);
```
jnthn: Leftover debug output

tools/ucd2c.pl
```diff
@@ -1581,6 +1601,24 @@ sub Jamo {
         $points_by_hex->{$code_str}->{Jamo_Short_Name} = $name;
     });
 }
+sub BidiMirroring {
+    #my ( $file, $propname ) = @_;
```
jnthn: Remove

src/strings/normalize.c
samcv: I can probably move that back under `case 'C'`. Should not interfere with what this section was changed to fix (Prepend marks are Cf).

src/strings/normalize.c
```diff
@@ -341,21 +351,24 @@ static MVMint32 is_control_beyond_latin1(MVMThreadContext *tc, MVMCodepoint in)
...
+                    return 1;
+                }
+                if (genprop[1] == 'f' ) {
+                    /* Format can have special properties (not control) */
+                    return 0;
+                }
         }
-        if (genprop[0] == 'C') {
-            /* Control, Surrogate, and Format are controls. */
-            if (genprop[1] == 'c' || genprop[1] == 's' || genprop[1] == 'f') {
```
samcv: Notice how the 'f' is removed. Since GCB='Prepend' are Cf

src/strings/normalize.c
samcv: Okay, will fix that.

src/strings/normalize.c
samcv: Never. Due to removing 'Cf' matching.

tools/ucd2c.pl
samcv: It would cause inconsistent results for the values of the enum. And may have totally broken it by changing the return values to some other number. (It would assign the wrong 'name' to whatever happened in that enum)

samcv:
Made those changes you wanted, and ran spectest again. Passing all but `'  ' ~~ /<:space>/` which is from possibly some bug in nqp.

PR #477, opened 2017-01-01
Implement in unicmp_s checking of secondary and tetriary collation weights

samcv:
An issue is also fixed where ucd2c.pl would not add up the total
   weighting for characters which decompose into multiple characters.

   We also return distinct values for the last level which decided the
   result.

   * 0   The strings are identical, includes codepoints
   * -1/1   We used the primary collation values to decide the result
   * -2/2   We used the secondary, meaning the primary weights were equal
   * -3/3   We used the tetriary, meaning the primary and also the secondary
       weights (if we used that option, which is the default) were equal.
   * -4/4   We used codepoints to decide because primary, secondary and/or tetriary
       were equal (depending if secondary and tetriary were requested).
       If the codepoints are all the same, we will decide based on
       length.
   * -10/10 means the collation algorithm was not able to be applied and so they
     were compared based on codepoints

Currently in MoarVM this should almost always be because of
collation weights not found or NYI for the string.
Currently this happens for graphemes made up of more than one codepoint in composed form.


src/strings/unicode_ops.c
```diff
@@ -1,6 +1,17 @@
...
  * equal or greater than based on Unicode Collation Algorithm */
-MVMint64 MVM_unicode_string_compare(MVMThreadContext *tc, MVMString *a, MVMString *b) {
+MVMint32 MVM_unicode_collation_primary (MVMThreadContext *tc, MVMint32 codepoint) {
+     MVM_unicode_codepoint_get_property_int(tc, codepoint, MVM_UNICODE_PROPERTY_MVM_COLLATION_PRIMARY);
+}
+MVMint32 MVM_unicode_collation_secondary (MVMThreadContext *tc, MVMint32 codepoint) {
+     MVM_unicode_codepoint_get_property_int(tc, codepoint, MVM_UNICODE_PROPERTY_MVM_COLLATION_SECONDARY);
+}
+MVMint32 MVM_unicode_collation_tertiary (MVMThreadContext *tc, MVMint32 codepoint) {
+     MVM_unicode_codepoint_get_property_int(tc, codepoint, MVM_UNICODE_PROPERTY_MVM_COLLATION_TERTIARY);
```
jnthn: These three seem to be missing `return` statements?

src/strings/unicode_ops.c
```diff
@@ -1,6 +1,17 @@
 /* Compares two strings, returning -1, 0 or 1 to indicate less than,
  * equal or greater than based on Unicode Collation Algorithm */
-MVMint64 MVM_unicode_string_compare(MVMThreadContext *tc, MVMString *a, MVMString *b) {
```
jnthn: Think the comment above wants changing to reflect the new range of return values.

src/strings/unicode_ops.c
```diff
@@ -25,23 +36,47 @@ MVMint64 MVM_unicode_string_compare(MVMThreadContext *tc, MVMString *a, MVMStrin
         bi = MVM_string_get_grapheme_at_nocheck(tc, b, i);
         /* If they are the same grapheme */
         if (ai != bi) {
-            /* only try and get a property if the value is zero or more */
+            /* only try and get a property if the value it is isn't a synthetic
+             * grapheme. we should change from get_grapheme to something to get the
+             * NFC form */
```
jnthn: If `CodePointIter` was used instead of indexing, then you'd be getting NFC.

jnthn: (But that can happen in a future change.)

tools/ucd2c.pl
```diff
@@ -1656,30 +1660,37 @@ sub collation {
...
             # don't put their values into the data structure.
-            if ( $temp =~ m/ (?: \[ (.) (\p{AHex}+) (.) (\p{AHex}+) (.) (\p{AHex}+) \] )+ /xms ) {
-                $weight1 = hex $2;
-                $weight2 = hex $4;
-                $weight3 = hex $6;
+
+            # When multiple tables are specified for a character, it is because those
+            # are the composite values for the decomposed character. Since we compare
+            # in NFC form not NFD, let's add these together.
+            print " $line_no ";
```
jnthn: I think this and other `$line_no` usages are left-behind debugging code?

PR #479, opened 2017-01-03
Add a script to download the latest version of all of the Unicode data

samcv:
Downloads UCD.zip, collation data (allkeys.txt) and emoji-data.txt using this script.

Since there is no 'latest' folder on the ftp for Emoji, we check if the ReadMe.txt has draft or preliminary in the file. All the other draft specs on the Unicode ftp server use one or more of those words in the draft, while do not use them at all in the final verision.


jnthn:
Nice! :-)

PR #485, opened 2017-01-04
Get property int property #define's in ucd2c.pl for Unicode

samcv:
Generate Decomposition_Type Unicode prop. #define's in ucd2c.pl

Decompose_Type, Unicode: use int lookup instead of str for better perf.
Also generate Canonical_Combining_Class #define's in ucd2c.pl.

This will reduce our need to compare strings.


samcv:
All tests pass.

PR #486, opened 2017-01-06
Don't break after ZWJ or for MALE SIGN, FEMALE SIGN or Glue_After_ZWJ

samcv:
Passes another ~700 more tests from emoji-test.t, also passes 2 more tests from GraphemeBreakTest.t

### DIFF
--- a/src/strings/normalize.c
+++ b/src/strings/normalize.c
@@ -561,9 +561,20 @@ static MVMint32 should_break(MVMThreadContext *tc, MVMCodepoint a, MVMCodepoint 
                 return 0;
             if ( GCB_b == MVM_UNICODE_PVALUE_GCB_ZWJ )
                 return 0;
+            if ( GCB_b == MVM_UNICODE_PVALUE_GCB_GLUE_AFTER_ZWJ )
+                return 0;
+            if ( b == 0x2640 || b == 0x2642 )
+                return 0;
+        case MVM_UNICODE_PVALUE_GCB_E_MODIFIER:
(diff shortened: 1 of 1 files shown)


src/strings/normalize.c
```diff
@@ -561,9 +561,20 @@ static MVMint32 should_break(MVMThreadContext *tc, MVMCodepoint a, MVMCodepoint
                 return 0;
             if ( GCB_b == MVM_UNICODE_PVALUE_GCB_ZWJ )
                 return 0;
+            if ( GCB_b == MVM_UNICODE_PVALUE_GCB_GLUE_AFTER_ZWJ )
```
jnthn: I wonder a little if a `switch` would be faster here, but maybe the compiler will figure that out anyway...

src/strings/normalize.c
```diff
@@ -561,9 +561,20 @@ static MVMint32 should_break(MVMThreadContext *tc, MVMCodepoint a, MVMCodepoint
...
+                return 0;
+            if ( b == 0x2640 || b == 0x2642 )
```
jnthn: Maybe we can `#define` constants for these rather than magic numbers in the code?

src/strings/normalize.c
samcv: I was thinking similar. I will do that.

src/strings/normalize.c
samcv: What file do you want them define’d in? Is this file alright?

jnthn: Yes, in this one is fine.

PR #487, opened 2017-01-06
Clean up frame->work lifetime

jnthn:
This work ensures that the lifetime of `->work` (a frame's working space) is strictly that of its lifetime on the call stack, modulo the case where the frame is not on the stack but its working storage lives on due to it having been captured in a continuation.

To achieve this, two VM features unused by Perl 6 have been eliminated:

* Taking register references (unused; intended for `localref` which was also unused and never implemented by any other backend; this will likely be a relief to those implementing said other backends)
* The partial (but already decidedly broken) bits of work towards supporting multi-shot continuations

With the lifetime clarified, it was possible to simplify a number of code paths, some of them hot. This also means memory can be freed sooner, reducing overhead and memory pressure.

This work leads to [around 20% less time and 10% less memory in the CORE.setting build](https://gist.github.com/jnthn/de0c1b39fd17e1ce27fa725f93541cfb).


src/core/frame.c
```diff
@@ -88,17 +88,19 @@ static void instrumentation_level_barrier(MVMThreadContext *tc, MVMStaticFrame *
...
     if (frame->work) {
         MVM_args_proc_cleanup(tc, &frame->params);
         MVM_fixed_size_free(tc, tc->instance->fsa, frame->allocd_work,
             frame->work);
     }
-    if (frame->env)
-        MVM_fixed_size_free(tc, tc->instance->fsa, frame->allocd_env, frame->env);
     if (frame->continuation_tags)
         MVM_continuation_free_tags(tc, frame);
+    if (frame->env)
```
timo: could make the diff a tiny bit smaller (easier to digest?) by moving this back up before the continuation_free_tags part.

PR #488, opened 2017-01-06
Take into account actual allocated of I/O buffers.

jnthn:
It seems libuv suggest we allocate 64KB sometimes, even when the
input we get is tiny. While I'm not sure second-guessing it is wise,
we should at least be honest internally about what's allocated. By
storing the actual allocated size, the GC can track it as part of
the gen2 promotion statistics, and be smarter about triggering full
collections. This reduces memory overhead.


jnthn:
No objections; merging.

PR #490, opened 2017-01-06
Free up spesh log slots after specialization.

jnthn:
Spesh logging keeps values alive, preventing the GC from collecting
them. It logs values to sample what types show up, which is fine, but
we should not hang on to them beyond the point the specializer has
used them in its analysis. This reduces memory overhead, perhaps
quite notably in some applications that have large objects (for
example, RT #130494 leaked many objects in this way). On CORE.setting
compilation it saves ~3MB - not much in the scheme of things, but nice
to win.


jnthn:
No objections; merged.

PR #492, opened 2017-01-10
Implement getstrbyname/MVM_unicode_string_from_name to retrieve grapheme clusters by name

samcv:
Returns a string from its Unicode name. Unlike MVM_unicode_lookup_by_name, this works
for multiple codepoints. This allows us to support Emoji sequences and other potential
names which are sequences of codepoints.

Working example: ```nqp-m -e "say(nqp::getstrfromname('person golfing: medium-light skin tone'))" #> 🏌🏼```


src/core/oplist
```diff
@@ -800,6 +800,7 @@ indexingoptimized   w(str) r(str) :pure
 captureinnerlex     r(obj) :noinline
 unicmp_s            w(int64) r(str) r(str) r(int64) r(int64) r(int64) :pure
 setdispatcherfor    r(obj) r(obj)
+getstrbyname         w(str) r(str) :pure
```
jnthn: The `w(str)` seems a space out of line with all those above it.

src/strings/unicode_ops.c
```diff
@@ -108,13 +108,41 @@ MVMGrapheme32 MVM_unicode_lookup_by_name(MVMThreadContext *tc, MVMString *name)
...
+
     if (!codepoints_by_name) {
         generate_codepoints_by_name(tc);
     }
     HASH_FIND(hash_handle, codepoints_by_name, cname, strlen((const char *)cname), result);
     MVM_free(cname);
+
     return result ? result->codepoint : -1;
 }
+/* XXX move this so it's included properly */
```
jnthn: Either should be done, or shouldn't but the comment not left behind.

src/strings/unicode_ops.c
```diff
@@ -108,13 +108,41 @@ MVMGrapheme32 MVM_unicode_lookup_by_name(MVMThreadContext *tc, MVMString *name)
...
+MVMString * MVM_string_from_grapheme(MVMThreadContext *tc, MVMCodepoint cp) {
```
jnthn: This is called `string_from_grapheme`, but actually is taking a codepoint, which is not always a grapheme.

src/strings/unicode_ops.c
```diff
@@ -108,13 +108,41 @@ MVMGrapheme32 MVM_unicode_lookup_by_name(MVMThreadContext *tc, MVMString *name)
...
+    MVMString *s;
+    MVMGrapheme32 g;
+    MVMNormalizer norm;
+
+    MVM_unicode_normalizer_init(tc, &norm, MVM_NORMALIZE_NFG);
+    if (!MVM_unicode_normalizer_process_codepoint_to_grapheme(tc, &norm, cp, &g)) {
+        MVM_unicode_normalizer_eof(tc, &norm);
+        g = MVM_unicode_normalizer_get_grapheme(tc, &norm);
```
jnthn: This code makes the (so far as I can tell, correct in Unicode 9) assumption that a single codepoint will always result in a single grapheme under NFG. I'm just a little nervous about this assumption as it's not true for NFC. I guess it's unlikely that such a case will arise in the future too. But it caught my eye, and thought it was worth mentioning.

src/strings/unicode.h
```diff
@@ -16,6 +16,11 @@ struct MVMUnicodeNameRegistry {
     MVMGrapheme32 codepoint;
     UT_hash_handle hash_handle;
 };
+struct MVMUnicodeGraphemeNameRegistry {
+    char *name;
+    MVMint32 structitem;
```
jnthn: Not entirely keen on `structitem` as a name...it seems more of an index to the codepoint sequence?

src/strings/unicode_ops.c
```diff
@@ -400,3 +440,42 @@ void MVM_unicode_release(MVMThreadContext *tc)
...
+    if (!result)
+        return tc->instance->str_consts.empty;
+
+    uni_seq = uni_seq_enum[result->structitem];
+    /* The first element is the number of codepoints in the sequence */
+    array_size = uni_seq[0];
+    MVM_unicode_normalizer_init(tc, &norm, MVM_NORMALIZE_NFG);
+    for (int i = 1; i < array_size; i++) {
+        cp = uni_seq[i];
+        MVM_unicode_normalizer_process_codepoint_full(tc, &norm, cp, &cp);
```
jnthn: This discards the return value. I guess at the moment we assume that all sequences are going to produce a single grapheme. That's reasonable, but we should at the very least stick in an `assert` that we really get `0` back from this.

jnthn: Also, re-using `cp` for the out parameter is a bit confusing.

src/strings/unicode_ops.c
```diff
@@ -400,3 +440,42 @@ void MVM_unicode_release(MVMThreadContext *tc)
...
+    MVM_free(cname);
+    /* If we can't find a result return an empty string */
+    if (!result)
+        return tc->instance->str_consts.empty;
+
+    uni_seq = uni_seq_enum[result->structitem];
+    /* The first element is the number of codepoints in the sequence */
+    array_size = uni_seq[0];
+    MVM_unicode_normalizer_init(tc, &norm, MVM_NORMALIZE_NFG);
+    for (int i = 1; i < array_size; i++) {
```
jnthn: If the first element is the number of codepoints in the sequence, but then we start at `1`, will we not miss the last codepoint?

src/strings/unicode_ops.c
```diff
@@ -400,3 +440,42 @@ void MVM_unicode_release(MVMThreadContext *tc)
...
+        cp = uni_seq[i];
+        MVM_unicode_normalizer_process_codepoint_full(tc, &norm, cp, &cp);
+    }
+    MVM_unicode_normalizer_process_codepoint_norm_terminator(tc, &norm, cp, &result_graph);
```
jnthn: I don't understand `cp` being fed into the normalizer here; either it's what we already put in, or it's whatever was already written into `cp` by the normalizer the last time around the loop. Should this just be calling the `eof`` function?

src/strings/unicode_ops.c
```diff
@@ -400,3 +440,42 @@ void MVM_unicode_release(MVMThreadContext *tc)
...
+    MVM_unicode_normalizer_cleanup(tc, &norm);
+
+    return MVM_string_from_grapheme(tc, result_graph);
```
jnthn: By this point we have an NFG grapheme. But then `MVM_string_from_grapheme` is running it through the normalizer again. That's a bit wasteful, but also the normalizer was designed to be fed codepoints, not graphemes, so there's a risk of oddness going on from that.

src/strings/unicode_ops.c
jnthn: Though as mentioned below, this whole pass through the normalizer may not make sense.

src/strings/unicode_ops.c
jnthn: fwiw, I think this particular routine would be better in `src/string/ops.c` and with a decl in `src/string/ops.h`.

src/strings/unicode_ops.c
samcv: Yes I think it should. Will make that change.

src/strings/unicode.h
samcv: Yeah it's the index. Name changed.

src/core/oplist
```diff
@@ -800,6 +800,7 @@ indexingoptimized   w(str) r(str) :pure
 captureinnerlex     r(obj) :noinline
 unicmp_s            w(int64) r(str) r(str) r(int64) r(int64) r(int64) :pure
 setdispatcherfor    r(obj) r(obj)
+getstrfromname        w(str) r(str) :pure
```
jnthn: Whitespace is a bit off again.

src/strings/unicode_ops.c
```diff
@@ -400,3 +412,35 @@ void MVM_unicode_release(MVMThreadContext *tc)
...
+    MVMGrapheme32 result_graph;
+    const MVMint32 * uni_seq;
+    int array_size;
+    MVMint32 cp;
+    MVMNormalizer norm;
+
+    result_graph = MVM_unicode_lookup_by_name(tc, name);
+    /* If it's just a codepoint, return that */
+    if (result_graph >= 0)
+        return MVM_string_from_grapheme(tc, result_graph);
```
jnthn: We could eliminate `MVM_string_from_grapheme` and just use `MVM_string_chr` here - which will take care of singleton decomposition, which this doesn't.

samcv:
Should have made all the changes you requested.

src/strings/unicode_ops.c
```diff
@@ -400,3 +412,35 @@ void MVM_unicode_release(MVMThreadContext *tc)
...
+MVMString * MVM_unicode_string_from_name(MVMThreadContext *tc, MVMString *name) {
+    MVMuint64 size;
+    char *cname = MVM_string_ascii_encode(tc, name, &size, 0);
+    MVMUnicodeGraphemeNameRegistry *result;
+
+    MVMGrapheme32 result_graph;
+    const MVMint32 * uni_seq;
+    int array_size;
+    MVMint32 cp;
+    MVMNormalizer norm;
```
jnthn: The variables `cp` and `norm` are unused.

src/strings/unicode_ops.c
```diff
@@ -400,3 +412,33 @@ void MVM_unicode_release(MVMThreadContext *tc)
...
+    MVMUnicodeGraphemeNameRegistry *result;
+
+    MVMGrapheme32 result_graph;
+    const MVMint32 * uni_seq;
+    int array_size;
+
+    result_graph = MVM_unicode_lookup_by_name(tc, name);
+    /* If it's just a codepoint, return that */
+    if (result_graph >= 0)
+        return MVM_string_chr(tc, result_graph);
```
jnthn: It seems if we take this path, then we leak `cname` (because we never hit the `MVM_free` below).

samcv: Fixed.

PR #493, opened 2017-01-10
Refactor MVM_unicode_codepoints_to_nfg_string

samcv:
Seperate out the section which involves MVMObject so we can re-use this
function in other places with native c data structures.

### DIFF
--- a/src/strings/normalize.c
+++ b/src/strings/normalize.c
@@ -81,36 +81,27 @@ void MVM_unicode_normalize_codepoints(MVMThreadContext *tc, const MVMObject *in,
     ((MVMArray *)out)->body.start     = 0;
     ((MVMArray *)out)->body.elems     = result_pos;
 }
-
-/* Takes an object, which must be of VMArray representation and holding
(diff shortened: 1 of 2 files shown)


src/strings/normalize.c
```diff
@@ -134,6 +125,21 @@ MVMString * MVM_unicode_codepoints_to_nfg_string(MVMThreadContext *tc, const MVM
...
 }
 
+/* Takes an object, which must be of VMArray representation and holding
+ * 32-bit integers. Treats them as Unicode codepoints, normalizes them at
+ * Grapheme level, and returns the resulting NFG string. */
+MVMString * MVM_unicode_codepoints_to_nfg_string(MVMThreadContext *tc, const MVMObject *codes) {
+    MVMCodepoint  *input;
+    MVMint64       input_codes;
+
+    /* Get input array; if it's empty, we're done already. */
```
jnthn: The `if it's empty` bit no longer applies; can just kill off this comment though as it doesn't add much. :-)

samcv: Fixed now.

Issue #482, opened 2017-01-03
`@*ARGS` utf-8 decoding broke


flussence:
I've bisected this down to rakudo/rakudo@ee38721ca58c, which was a moarvm version bump (9a9f2d46..fe110d6):
```
 ~/perl6/rakudo $ ./perl6 -MTest -e 'put @*ARGS.perl; is-deeply @*ARGS[0].chars, 1' '“'
["􏿽xE2􏿽x80􏿽x9C"]
not ok 1 -

# Failed test at -e line 1
# expected: 1
#      got: 3
```
Note that those aren't whitespace before each "x", they're `U+10FFFD`.

Expected result:
```
 ~ $ /usr/bin/perl6 --version
This is Rakudo version 2016.12 built on MoarVM version 2016.12
implementing Perl 6.c.
 ~ $ /usr/bin/perl6 -MTest -e 'is-deeply @*ARGS[0].chars, 1' '“'
ok 1 -
```

jnthn:
Fixed in 9475d8db4c. Note that command line arguments are not decoded by a standard UTF-8 decoder, but rather as one that seeks to preserve the bytes as they were provided and not choking on non-UTF-8. This is because command line arguments often contain file names, and file names may be any byte sequence on POSIX platforms. The intent was that anything that will not have its round-trip broken by being stored in NFG will be decoded as UTF-8; the private use sequence is used to encode bytes that would be transformed, so they can be reproduced. The commit that broke things was a bit too overzealous on fixing a round-tripping failure; the fix should address this and other similar cases.

PR #494, opened 2017-01-10
Decode (hopefully) all NFC UTF8 to NFG in UTF8-C8

jnthn:
In the last round of tweaks to UTF8-C8, we fixed some sequences that
would not round-trip properly due to being mis-represented in UTF8.
The fix dealt with those cases, but was a bit too sweeping. UTF8-C8
aims to decode everything that's both valid UTF8 and in NFC as the
UTF8 decoder would, and express everything else as synthetics that
will ensure round-tripping. This fix deals with the issue raised in
MoarVM Issue #482, while not regressing any of the UTF8-C8 roundtrip
tests.


jnthn:
@arnsholt took a look at this and mentioned it looks fine; will take that as sufficient review. :-)

Issue #478, opened 2017-01-01
Must not GC when in the specializer/JIT while running code example from RT #127987


dogbert17:
Running the code example from https://rt.perl.org/Public/Bug/Display.html?id=127987 with a small nursery, i.e. 64k and MVM_GC_DEBUG 2 the above message appears most of the time.

dogbert@dogbert-VirtualBox ~/repos/rakudo $ cat ~/scratch.pl6 
for (^100).race(batch=>1) { {1 + $^a}.assuming(10)() }

dogbert@dogbert-VirtualBox ~/repos/rakudo $ gdb --args /home/dogbert/repos/rakudo/install/bin/moar --libpath="/home/dogbert/repos/rakudo/install/share/nqp/lib" --libpath="." /home/dogbert/repos/rakudo/perl6.moarvm -I lib ~/scratch.pl6
GNU gdb (Ubuntu 7.7.1-0ubuntu5~14.04.2) 7.7.1
Copyright (C) 2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "i686-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /home/dogbert/repos/rakudo/install/bin/moar...done.
(gdb) break MVM_panic
Function "MVM_panic" not defined.
Make breakpoint pending on future shared library load? (y or [n]) y
Breakpoint 1 (MVM_panic) pending.
(gdb) r
Starting program: /home/dogbert/repos/rakudo/install/bin/moar --libpath=/home/dogbert/repos/rakudo/install/share/nqp/lib --libpath=. /home/dogbert/repos/rakudo/perl6.moarvm -I lib /home/dogbert/scratch.pl6
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/i386-linux-gnu/libthread_db.so.1".
[New Thread 0xb657eb40 (LWP 27134)]
[New Thread 0xb5bffb40 (LWP 27135)]
[...]

PR #495, opened 2017-01-11
Fix bugs in interaction between inlining, GC, and threads

jnthn:
This PR addresses:

* Issue #478 by avoiding a bunch of codepaths that could end up with us running GC during specialization, which were triggered by cross-compilation-unit inlining
* Three abuses of `realloc` that could lead to other threads reading freed memory (this accounts for the ASAN noise in spectest6, for example)


timo:
looked though the diff and it looked good to me.

PR #498, opened 2017-01-12
Convert MVM_malloc+memset to MVM_calloc

MasterDuke17:
In cases of `foo = MVM_malloc(...); memset(foo, 0, ...);`, convert them
to `foo = MVM_calloc(...);`

NQP builds and passes `make m-test` and Rakudo build and passes `make m-spectest`.

### DIFF
--- a/src/6model/reprs/CUnion.c
+++ b/src/6model/reprs/CUnion.c
@@ -331,8 +331,7 @@ static void initialize(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, voi
 
     /* Allocate object body. */
     MVMCUnionBody *body = (MVMCUnionBody *)data;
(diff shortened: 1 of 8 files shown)


src/core/validation.c
```diff
@@ -671,8 +671,7 @@ void MVM_validate_static_frame(MVMThreadContext *tc,
 
 #ifdef MVM_BIGENDIAN
     assert(fb->bytecode == fb->orig_bytecode);
-    val->bc_start = MVM_malloc(fb->bytecode_size);
-    memset(val->bc_start, 0xDB, fb->bytecode_size);
```
jnthn: Oops, you can't do it here because we aren't zeroing.

MasterDuke17:
Ugh, can't believe I didn't see that. Anyway, updated to remove that one. jnthn++

PR #501, opened 2017-01-15
Fix warnings during compile

MasterDuke17:
I get the format string warnings on my Kubuntu 16.10 machine (gcc 6.2.0), but not on my Arch Linux machine (gcc 6.3.1).

### DIFF
--- a/src/6model/reprs/NFA.c
+++ b/src/6model/reprs/NFA.c
@@ -454,7 +454,7 @@ static MVMint64 * nqp_nfa_run(MVMThreadContext *tc, MVMNFABody *nfa, MVMString *
         if (nfadeb) {
             if (offset < eos) {
                 MVMGrapheme32 cp = MVM_string_get_grapheme_at_nocheck(tc, target, offset);
-                fprintf(stderr,"%c with %ds target %lx offset %lld\n",cp,(int)numcur, (long)target, offset);
+                fprintf(stderr,"%c with %ds target %lx offset %ld\n",cp,(int)numcur, (long)target, offset);
             }
             else {
                 fprintf(stderr,"EOS with %ds\n",(int)numcur);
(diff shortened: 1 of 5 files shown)


src/strings/unicode_ops.c
```diff
@@ -257,7 +257,7 @@ static void generate_property_codes_by_seq_names(MVMThreadContext *tc) {
         entry->name = (char *)uni_seq_pairs[num_names].name;
         entry->structindex = uni_seq_pairs[num_names].value;
         HASH_ADD_KEYPTR(hash_handle, property_codes_by_seq_names,
-            entry->name, strlen(entry->name), entry);
+            entry->name, strlen(entry->name), (MVMUnicodeNameRegistry *)entry);
```
jnthn: Wouldn't a more correct fix here be to change the type of `*property_codes_by_seq_names` above to `MVMUnicodeGraphemeNameRegistry *`?

MasterDuke17: Not sure. @samcv what's your thought?

samcv: I don't think it's a good idea to cast it as something that it is not, since MVMUnicodeNameRegistry holds differently typed items than MVMUnicodeGraphemeNameRegistry, the former holds a grapheme and the latter holds an index for another struct.

Maybe there is some way without changing the types?

jnthn: I took care of it in 696883728b.

MasterDuke17:
@jnthn fixed the warning that the second commit in this PR addressed. Should I remove that commit so the rest of the PR can be merged? Or are those changes not wanted and I should just close it?

jnthn:
Remove the commit; rest of the PR looked good.

MasterDuke17:
Done.

PR #503, opened 2017-01-16
Save a tiny bit of space in 6 of our structs

timo:
Since the struct members used to be grouped together, feel free to reject any of the individual patches. Of course, the comments can also be changed to make things easier to figure out.

### DIFF
--- a/src/jit/compile.h
+++ b/src/jit/compile.h
@@ -11,22 +11,19 @@ struct MVMJitCode {
      * runtime (because we need a second dereference to figure the
      * labels out), but very simple for me now, and super-easy to
      * optimise at a later date */
-    MVMint32   num_labels;
-    void     **labels;
-
+    MVMint32       num_labels;
     MVMint32       num_bbs;
+    void         **labels;
     MVMint32      *bb_labels;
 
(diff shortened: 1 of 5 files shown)


src/jit/compile.h
```diff
@@ -11,22 +11,19 @@ struct MVMJitCode {
...
+    MVMint32       num_labels;
     MVMint32       num_bbs;
+    void         **labels;
     MVMint32      *bb_labels;
 
     MVMint32       num_deopts;
-    MVMJitDeopt    *deopts;
-
     MVMint32       num_inlines;
+    MVMJitDeopt    *deopts;
```
niner: Would be lovely if you added some comments to retain the information, which of the fields are related as you did in src/6model/reprs/MVMStaticFrame.h

timo:
Here's the rest that could be optimized. The last column is how much can be saved by shuffling things around.

```MVMInstance	1960	1952	8
MVMSpeshGraph	216	208	8
MVMNormalizer	48	40	8
MVMCArrayREPRData	24	16	8
MVMJitGraph	88	80	8```

PR #504, opened 2017-01-16
Multi cache no segfault on null

timo:
liz' recent changes to how X works somehow managed to sneak a null pointer into these code paths, and i think it's much better to not segfault when that happens.

not 100% sure if the handling of null pointers in the cache is correct. I tried to make it "ignore the cache completely" when a null pointer arrives.


bdw:
@timo is this still relevant?

PR #505, opened 2017-01-18
Add some new nqp:: ops for swapping array entries

MasterDuke17:
nqp::swappos[_i, _n, _s](@a, int $index1, int $index2)

See https://irclog.perlgeek.de/perl6-dev/2017-01-18#i_13940214 and https://irclog.perlgeek.de/perl6-dev/2017-01-18#i_13941033


jnthn:
Thanks, but I don't think this deserves to be a VM-level op. It may be that there's value in Perl 6 defining a desugar ([like this](https://github.com/rakudo/rakudo/blob/nom/src/vm/moar/Perl6/Ops.nqp#L101)) or a more likely code-gen function for a new op that turns into multiple VM ops ([like here](https://github.com/rakudo/rakudo/blob/nom/src/vm/moar/Perl6/Ops.nqp#L115)).

In general, we try to avoid adding complex VM ops that are easily made up out of smaller ones, because the only thing they save is interpreter overhead, which disappears under JIT compilation anyway. The cost of having them is fairly high in the long run, because they are harder to analyze, so in the end are liable to end up missing out on optimizations that the simpler operations they are made up of would get.

PR #506, opened 2017-01-18
Inactivate async tasks

jnthn:
This fixes a leak of async task handles, which was responsible for a managed memory leak (e.g. GC-managed objects that never became collectable). The mechanism to blame exists in order to safely have libuv callbacks able to obtain GC-managed references. We can't have the GC traverse libuv internals, so we stash the async task handles in an array, and store that index. This works well, but we needed to clear those array entries in the future, to avoid leaking AsyncTask handles. This is achieved by this work.

This gets rid of most of the leakage, but there is still a far more minor one: the task handles array itself will grow uboundedly. This will be fixed by re-using the indexes; that will happen in a future PR, once this one has gone in and had some real-world testing.

### DIFF
--- a/src/io/asyncsocket.c
+++ b/src/io/asyncsocket.c
@@ -82,6 +82,7 @@ static void on_read(uv_stream_t *handle, ssize_t nread, const uv_buf_t *buf) {
         if (buf->base)
             MVM_free(buf->base);
         uv_read_stop(handle);
+        MVM_io_eventloop_remove_active_work(tc, &(ri->work_idx));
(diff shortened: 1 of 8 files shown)


niner:
All changes reviewed, understood and approved. Please merge :)

jnthn:
Decided to do it after the monthly release; just a tiny bit fearful one of the new panics may be hit accidentally (can't think of or produce a case, but would kick myself if I shoved it in the day before release, and we find such a case the day after...)

PR #507, opened 2017-01-18
Fix decode stream leaks

jnthn:
This fixes a couple of leaks related to decode streams, and also fixes a typo.


niner:
The only thing I don't like about this PR is that you first introduce a call to a misspelled function and then afterwards change the new call along with the others. A bit of unnecessary churn. Other than that this looks quite straight forward.

jnthn:
I didn't notice the typo until afterwards, and even then pondered whether to just leave it be for somebody else to fix. :-) Will merge as is, rather than do a rebase dance etc.

PR #508, opened 2017-01-18
Fix typo in comment

MasterDuke17:
Final destory->destroy?


jnthn:
Heh, forgot `git grep` is case-sensitive. :-)

PR #511, opened 2017-01-25
Make getting Uni seq/cp's case insensitive and add seq from NamedSequences.txt

samcv:
This makes looking up codepoints or sequences by name case insensitive. We store them
all in uppercase when generating the Unicode database and then uppercase any requests we
receive before trying to look them up.

This also adds the sequences from NamedSequences.txt, previously only the Emoji Sequences
and Emoji ZWJ Sequences were in our database.


src/strings/unicode_ops.c
```diff
@@ -417,7 +417,7 @@ void MVM_unicode_release(MVMThreadContext *tc)
 /* Looks up a grapheme by name. Lazily constructs a hash. */
 MVMString * MVM_unicode_string_from_name(MVMThreadContext *tc, MVMString *name) {
     MVMuint64 size;
-    char *cname = MVM_string_ascii_encode(tc, name, &size, 0);
+    char *cname = MVM_string_ascii_encode(tc, MVM_string_uc(tc, name), &size, 0);
```
jnthn: `MVM_string_uc` will allocate a new `MVMString`, which in turn may cause the pointer to `name` to become outdated. Thus it is not safe to use it again, however it is used again below in the call to `MVM_unicode_lookup_by_name`. This may cause occasional reading of junk memory and SEGVs.

One solution would be to do an `MVMROOT` of name. However, I note that we currently will end up doing the `uc` and ASCII encoding twice, which is a bit wasteful. A neater and more efficient fix might be to extract the part of the looking up of codepoint names by a C string into a function, and call that from here, so that we care only doing the uc/encode one time.

samcv: I had thought this when I was making these changes as well. Sounds like a plan.

src/strings/unicode_ops.c
```diff
@@ -106,7 +106,7 @@ MVMint64 MVM_unicode_string_compare
 /* Looks up a codepoint by name. Lazily constructs a hash. */
 MVMGrapheme32 MVM_unicode_lookup_by_name(MVMThreadContext *tc, MVMString *name) {
     MVMuint64 size;
-    char *cname = MVM_string_ascii_encode(tc, MVM_string_uc(tc, name), &size, 0);
+    char *cname = MVM_string_ascii_encode(tc, name, &size, 0);
```
jnthn: Hm, doesn't this lose us case-insensitivity on codepoint names? Is this an intentional consequence? (Or is it just that we never notice since we don't expose the old op in Perl 6 space any more and use the new one for all `\c[...]` things?)

samcv: We will never notice yes. We use the new op in Rakudo

samcv: I don't want to write it into another function until I decide what to do about the Emoji Seq names which have diacritics in them as well.

jnthn: Well, encoding to ASCII will blow up on those for sure. We could encode to UTF-8. Provided we construct the hash using UTF-8 keys we'll be fine.

jnthn: This seems like something for a latter PR, however?

samcv: Yeah. I think so.

PR #512, opened 2017-01-26
Line based coverage pr

timo:
we've had the line based coverage thing in a bunch of often-rebased branches so far. It's time to merge it into master so the oplist and interp.c and such no longer give merge conflicts :)

### DIFF
--- a/build/Makefile.in
+++ b/build/Makefile.in
@@ -205,6 +205,7 @@ OBJECTS = src/core/callsite@obj@ \
           src/profiler/profile@obj@ \
           src/profiler/heapsnapshot@obj@ \
           src/instrument/crossthreadwrite@obj@ \
+          src/instrument/line_coverage@obj@ \
           src/moar@obj@ \
           @platform@ \
           @jit@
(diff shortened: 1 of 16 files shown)


timo:
annoyingly the line based profiler still causes lines that were definitely executed to not be marked as executed.

I think that's because the code so far only takes the first annotation that belongs to each BB, but we probably have BBs that span multiple lines, and those only get the first line reported.

timo:
Also, the procedure to actually get a report is annoyingly roundabout; you have to moar --dump the file you want to have reported, you should collect all the output files and to make things faster run it through sort and uniq, and then you can finally put those files together with the source file/files that gave the .moarvm file you --dump'ed and get a folder full of html out of the whole ordeal ...

jnthn:
A current problem we have is that we end up with wrong line numbers etc. when reporting on specialized code. The solution for that would partly consist of inserting spesh annotations onto nodes where the original bytecode stream had a line/file number annotation. We could then use those to generate specialized annotations. It strikes me that implementing the first half of that would get you more accurate reporting when a basic block contains multiple lines of code, which is likely not rare.

timo:
even though this branch still needs work to be really good, we'll merge it so that the oplist doesn't cause merge conflicts every time a new op gets added.

PR #515, opened 2017-01-29
Make radix 50% faster with Unicode Nd, add East_Asian_Width prop.

samcv:
Don't use the general category which is slow and requires string operations.
Check that Numeric_Type=Decimal

This should also speed up radix in general when it is given non A-Z, a-z or 0-9 codepoints

This also adds support for the East_Asian_Width property.


samcv:
Full spectest pass, plus some todo'd

MasterDuke17:
Could this same change be done for MVM_bigint_radix() also?

samcv:
@MasterDuke17 also going to mention I applied this to MVM_bigint_radix as well

PR #516, opened 2017-02-01
Add data section for JIT compiled code

bdw:
Adding a data section allows us to store large data blobs in code rather than loading them as constants or using externally allocated memory. This is expected to plug a memory leak of malloc()'ed extop 'fake registers', and it is used to store constant pointers to functions, avoiding emitting a large 'movabs' instruction.


jnthn:
Reviewed this, looks like a really nice improvement that we'll be able to use for various things. Thanks!

Issue #59, opened 2013-10-05
Consistent naming for REPRs


gerdr:
The REPR name occurs in various places, among those the prefix of C structures and functions, file names, the `MVM_REPR_ID_*` constant and the actual user-exposed `name` member of the `MVMREPROps` structure.

Some names have a `VM` prefix for vm-specific but multi-backend types and some purely MoarVM-specific types have a `MVM` prefix.

For readability, both C prefixes `MVMVM` and `MVMMVM` get shortened to `MVM`. However, we also lost the distinction between `VM` and `MVM` REPRs where we need not do so, in particular the file names and REPR-ID constants.

For consistency, these should match their authorative, user-exposed names - eg `MVMString` should be kept as-is, but `MVMArray.c`, `MVMArray.h` and `MVM_REPR_ID_MVMArray` should be renamed to `VMArray*`.

jnthn:
Trouble with this is that we'd then be exposing a bunch of symbols to the world without the MVM prefix, which doesn't seem desirable. I think we can live with the public REPR-name mismatch with the type name in the C code as the lesser evil.

PR #522, opened 2017-02-05
renamed MVMArray to VMArray

jeffythedragonslayer:
I fixed https://github.com/MoarVM/MoarVM/issues/59

FYI, this is my first time ever contributing to an open source project


src/6model/reprs.c
```diff
@@ -196,7 +196,7 @@ void MVM_repr_initialize_registry(MVMThreadContext *tc) {
 
     /* Add all core representations. */
     register_core_repr(String);
-    register_core_repr(Array);
+    register_repr(tc, VMArray_initialize(tc), NULL);
     register_core_repr(Hash);
```
zoffixznet: I'm not familiar with this codebase.... but I notice all the other changes are merely renamings, but this particular change seems to replace code line with something entirely different. Is that meant to be that way?

jeffythedragonslayer: Check line 188:

#define register_core_repr(name) \
    register_repr(tc, MVM##name##_initialize(tc), NULL)

I expanded the macro manually because it forces MVM to be in there

zoffixznet:
Comments from IRC:
> &lt;jnthn&gt;  Hmm...looking at that one for a moment between cooking tasks...I'm not quite convinced it's a desirable change. :-( We try to prefix everything that we export MVMFoo (types) or MVM_foo (functions)

jnthn:
I just took a look at #59 and while it was well-intended, it would also cause us to expose non-MVM-prefixed things, which would make for a potential source of conflicts for anyone embedding. Generally, it's polite for libraries that may be embedded to have a common prefix on things; we use MVM, as MVMFoo for types and MVM_foo for functions. So, I'd say we can live with the inconsistency as the less bad thing.

What we do seem to do in most cases (but not this one) is to name the constant and the filename without MVM in the cases of public REPRS. So, the rename of MVMArray.c to VMArray.c is OK, as is the constant `MVM_REPR_ID_MVMArray` changing to `MVM_REPR_ID_VMArray`; that is consistent with what we do elsewhere. The typename changes aren't so desirable, however. If you fancy backing those parts out, the remaining bits (constant name change, file name change) would be fine to merge.

Sorry for the confusion, and thanks for contributing.

jeffythedragonslayer:
Ok, I'll try again with just those changes.  Hopefully soon.

jnthn:
Closing as replaced by #532.

PR #523, opened 2017-02-05
Implement support for synthetic graphemes in MVM_unicode_string_compare

samcv:
We iterate by grapheme, and if we find one which is synthetic, we add all of the
collation weights for the codepoints which make it up together and compare that way.


samcv:
Commited this directly. Closing.

PR #530, opened 2017-02-09
Enforce consistent commenting style

samcv:
Use /* */ instead of //

// commenting is not part of C standards before C99


jnthn:
I was sorta expecting "enforce" to mean "add compiler flag to make it happen". :-) Anyway, this is nice for consistency, so will merge.

PR #532, opened 2017-02-11
VMArray2

jeffythedragonslayer:
Second attempt at the rename!


jnthn:
This appears to be missing a couple of things. First, the Makefile.in needs updating for the changed filename; the automated build reports:

```
gcc: error: src/6model/reprs/MVMArray.o: No such file or directory
make: *** [libmoar.so] Error 1
```

Secondly, it'd be good to rename the header file `MVMArray.h` to `VMArray.h` also. (Note this will also need tweaking in the Makefile, and probably in something like reprs.h also).

jnthn:
Looks good now; thanks!

Issue #533, opened 2017-02-11
Cygwin build fails


nanis:
This is with an up-to-date Cygwin install on Windows 10, fresh `git clone` (01570074025f52ede3633c8129c44fb7a41d9e1a).

```text
$ perl Configure.pl
Welcome to MoarVM!

Updating submodules .................................... OK
Configuring native build environment ...................
Testing pkgconfig ... 0.29.1
OK
    trying to compile a simple C program ............... YES
    probing whether your compiler thinks that it is gcc  YES
    probing how your compiler does static inline ....... static __inline__
    your CPU can read unaligned values for all of int32 int64 num64
    probing the size of pointers ....................... 8
JIT isn't supported on cygwin-thread-multi yet.
    probing C type support for: _Bool, bool ............ YES: _Bool,bool
    probing computed goto support ...................... YES
    probing pthread_yield support ...................... YES

        make: make
     compile: gcc -Wdeclaration-after-statement -Werror=declaration-after-statement -O3 -DNDEBUG  -DWIN32 -DAO_ASSUME_WINDOWS98 -fPIC
    includes:  -I3rdparty/libuv/include -I3rdparty/libuv/src -I3rdparty/libatomic_ops/src -I3rdparty/libtommath -I3rdparty/dynasm -I3rdparty/dyncall/dynload -I3rdparty/dyncall/dyncall -I3rdparty/dyncall/dyncallback
        link: gcc  -O3 -DNDEBUG -Wl,-rpath,"/@libdir@"
        libs: -lshell32 -lws2_32 -lmswsock -lrpcrt4 -ladvapi32 -lpsapi -liphlpapi -luserenv

  byte order: little endian

Configuring 3rdparty libs .............................. OK

  3rdparty: 3rdparty/dyncall/dyncall/libdyncall_s.a
            3rdparty/dyncall/dyncallback/libdyncallback_s.a
            3rdparty/dyncall/dynload/libdynload_s.a
            3rdparty/libtommath/libtommath.a
            3rdparty/libuv/libuv.a
            3rdparty/sha1/libsha1.a
            3rdparty/tinymt/libtinymt.a

Generating src/gen/config.c ............................ OK
Generating src/gen/config.h ............................ OK
Generating Makefile .................................... OK
Generating tools/check.mk .............................. OK
Generating build/mk-moar-pc.pl ......................... OK

Configuration SUCCESS.
```

Output from `make`:

```text
compiling src/main.o
[...]

PR #534, opened 2017-02-11
Issue #533: MoarVM does not build on `cygwin`.

nanis:
Remove mention of `cygwin` support from `Configure.pl`.


jnthn:
Wow, didn't even realize that was there. Thanks!

PR #539, opened 2017-02-21
Streamline some switches and try to suppress Coverity warnings

samcv:
Streamline switches, try to suppress Coverity warnings in MVM_string_compute_hash_code


jnthn:
Took me a bit to be comfortable with this, since it seems to flip the order we put stuff into the buffer for hashing. However, that appears not to matter since we add the bytes into the hash in the correct order.

PR #546, opened 2017-03-05
Check result of repeat/concat fit in an MVMString

MasterDuke17:

Fixes https://rt.perl.org/Ticket/Display.html?id=127971

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -397,6 +397,7 @@ static MVMint32 final_strand_matches(MVMThreadContext *tc, MVMString *a, MVMStri
 MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString *b) {
     MVMString *result;
     MVMuint32  agraphs, bgraphs;
+    MVMuint64  total_graphs;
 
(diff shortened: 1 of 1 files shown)


src/strings/ops.c
```diff
@@ -509,6 +518,13 @@ MVMString * MVM_string_repeat(MVMThreadContext *tc, MVMString *a, MVMint64 count
     if (agraphs == 0)
         return tc->instance->str_consts.empty;
 
+    /* Total size of the resulting string can't be bigger than an MVMString is allowed to be. */
+    total_graphs = (MVMuint64)agraphs * (MVMuint64)count;
```
jnthn: Since count is a 64-bit integer, it's possible that this bounds check itself could overflow.

MasterDuke17: agraphs can't be greater than 0xFFFFFFFF (since num_graphs is a MVMuint32) and we already check that count is not greater than (1 << 30). Should be fine, right?

jnthn: Ah, I missed the earlier check on `count`. Then yeah, this is safe. :+1:

PR #548, opened 2017-03-08
Fix string_equal_at_ignore_case when string `a` changes length

samcv:
This makes nqp tests using `nqp::eqatic` that will soon to be in the nqp
repo pass.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -583,6 +583,8 @@ MVMint64 MVM_string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *haysta
     /* Store the grapheme as we iterate through the haystack */
     MVMGrapheme32 h_g, n_g;
     MVMint64 i, j;
+    MVMuint32 h_fc_cps;
+
     /* An additional needle offset which is used only when codepoints expand
      * when casefolded. The offset is the number of additional codepoints that
      * have been seen so haystack and needle stay aligned */
@@ -604,9 +606,9 @@ MVMint64 MVM_string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *haysta
     });
 
     for (i = 0; i + h_offset < h_graphs && i + n_offset < n_graphs; i++) {
-        const MVMCodepoint *h_result_cps;
-        MVMuint32 h_fc_cps = MVM_unicode_get_case_change(tc, h_g, MVM_unicode_case_change_type_fold, &h_result_cps);
+        const MVMCodepoint* h_result_cps;
         h_g = MVM_string_get_grapheme_at_nocheck(tc, haystack, h_offset + i);
+        h_fc_cps = MVM_unicode_get_case_change(tc, h_g, MVM_unicode_case_change_type_fold, &h_result_cps);
         /* If we get 0 for the number that means the cp doesn't change when casefolded */
         if (h_fc_cps == 0) {
             n_g = MVM_string_get_grapheme_at_nocheck(tc, needle_fc, i + n_offset);
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -647,29 +647,6 @@ MVMint64 MVM_string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *haysta
         }
     }
     return return_val;
-    /*
-    change_a = MVM_string_graphs(tc, haystack_fc) - h_graphs;
-
-    if (change_a == 0 && n_change == 0) {
-        return MVM_string_equal_at(tc, haystack_fc, needle_fc, h_offset);
-    }
-    if (change_a > 0) {
-        MVMint64 counter = 0;
-        /* Find the point where the case-folded string and the original string
-         * contain different characters
-         * Although it shouldn't be possible to overrun the string unless the
-         * last cp of `a` expands to itself + another cp, ensure we don't go past
-         * the length of the original string */
-        /*while (MVM_string_substrings_equal_nocheck(tc, a, counter, 1, haystack_fc, 0) && counter < h_graphs) {
-            counter++;
-        }
-        /* If the string changes before the offset we're checking, we have
-         * to adjust our offset. If the string changes *after* the offset we
-         * are checking, then the offset is unchanged */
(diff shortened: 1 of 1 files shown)


src/strings/ops.c
```diff
@@ -575,12 +575,30 @@ MVMint64 MVM_string_equal_at(MVMThreadContext *tc, MVMString *a, MVMString *b, M
...
     });
+    change_a = MVM_string_graphs(tc, fca) - len_a;
+    change_b = MVM_string_graphs(tc, fcb) - len_b;
+
+    if (change_a == 0 && change_b == 0)
+        return MVM_string_equal_at(tc, fca, fcb, offset);
+
+    if (change_a > 0) {
+        counter = 0;
+        while (MVM_string_substrings_equal_nocheck(tc, a, counter, 1, fca, 0)) {
```
jnthn: This access to `a` is unreliable because it was not `MVMROOT`'d above.

src/strings/ops.c
```diff
@@ -575,12 +575,30 @@ MVMint64 MVM_string_equal_at(MVMThreadContext *tc, MVMString *a, MVMString *b, M
...
+            counter++;
+        }
+        if (offset > counter)
+            offset += change_a;
```
jnthn: What if there are multiple changes in `a` -> `fca`, some before and some after the counter? I'm tired and may be mis-reading, but it seems we assume all of them are ahead of the offset we're looking at.

src/strings/ops.c
jnthn: Is there any way that if the change is at the end of `a` then we read off the end of it? I think "not quite" but we might to explain it with a comment or add an explicit bounds check, because I don't find it totally easy to reason about. I think the reasoning is something like "the foldcase if it expands to can never start off with the same first codepoint"?

samcv:
Found a major flaw, since I compare the string to find where they changed case, if there's a capital letter before the special character then it completely misses it.

I am working on writing new code that will only fold case from the offset, instead of the entire string, which will avoid this issue.

src/strings/ops.c
```diff
@@ -572,16 +572,104 @@ MVMint64 MVM_string_equal_at(MVMThreadContext *tc, MVMString *a, MVMString *b, M
...
+    MVMStringIndex h_change, n_change;
+    MVMGrapheme32 h_g, n_g;
+    MVMint64 i, j;
+    MVMint64 return_val = 1;
+    MVMint64 n_offset = 0;
+    fprintf(stderr, "starting function\n");
+    if (h_offset < 0) {
+        h_offset += h_graphs;
+        if (h_offset < 0)
+            h_offset = 0; /* XXX I think this is the right behavior here */
```
jnthn: We tend to throw on such out-of-range things.

src/strings/ops.c
```diff
@@ -572,16 +572,77 @@ MVMint64 MVM_string_equal_at(MVMThreadContext *tc, MVMString *a, MVMString *b, M
...
+        fprintf(stderr, "h_fc_cps=%i first loop i=%li\n", h_fc_cps, i);
+        /* If we get 0 for the number that means the cp doesn't change when casefolded */
+        if (h_fc_cps == 0) {
+            n_g = MVM_string_get_grapheme_at_nocheck(tc, needle_fc, i + n_offset);
+            if (h_g != n_g) {
+                fprintf(stderr, "614 in first for loop h_g[%i] != n_g[%i] j=%li\n", h_g, n_g, j);
+                return_val = 0;
+                break;
+            }
+        }/*
```
samcv: Do you think the code below, which programs for the case when it folds to only one character should be its own codepath, or we should use the one further down 629-640 with the inner for loop? Code works both ways.

jnthn: I think I'd go with less code, and just handle it in the path below. I doubt there'll be much of a performance difference; sure we avoid the branching associated with a loop, but only because we add a branch checking for that specific case. So it probably breaks even.

samcv:
Version 2.0 is done and pushed to my repo. Full spectest pass. @jnthn let me know if it is fine and if so I will squash it a bit and then merge. Thanks :)

src/strings/ops.c
```diff
@@ -571,17 +571,60 @@ MVMint64 MVM_string_equal_at(MVMThreadContext *tc, MVMString *a, MVMString *b, M
...
+        MVMROOT(tc, needle_fc, {
+            needle_fc = MVM_string_fc(tc, needle);
         });
     });
-    return MVM_string_equal_at(tc, fca, fcb, offset);
+
+    for (i = 0; i + h_offset < h_graphs && i + n_offset < n_graphs; i++) {
+        const MVMCodepoint *h_result_cps;
+        h_g = MVM_string_get_grapheme_at_nocheck(tc, haystack, h_offset + i);
+        MVMuint32 h_fc_cps = MVM_unicode_get_case_change(tc, h_g, MVM_unicode_case_change_type_fold, &h_result_cps);
```
jnthn: This will break on some compilers, because the line above is not a decl. However, since `h_g` is only used inside of this loop, then we can simply declare it here instead.

src/strings/ops.c
```diff
@@ -571,17 +571,60 @@ MVMint64 MVM_string_equal_at(MVMThreadContext *tc, MVMString *a, MVMString *b, M
...
+ * Sometimes there is a difference in length of a string before and after foldcase,
+ * because of this we must compare this differently than just foldcasing both
+ * strings to ensure the offset is correct */
+MVMint64 MVM_string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *haystack, MVMString *needle, MVMint64 h_offset) {
+    /* foldcase version of needle */
+    MVMString *needle_fc;
+    MVMStringIndex h_graphs = MVM_string_graphs(tc, haystack);
+    MVMStringIndex n_graphs = MVM_string_graphs(tc, needle);
+    /* Store the grapheme as we iterate through the haystack */
+    MVMGrapheme32 h_g, n_g;
```
jnthn: Aside from declaring `h_g` further down, `n_g` could also have narrower scope, and also `j`.

PR #550, opened 2017-03-14
Add case insensitive string index op indexic_s

samcv:
And corresponding MVM C function MVM_string_index_ignore_case.

For now, calls MVM_string_equal_at_ignore_case along the string, which ends up with the needle being foldcased each time.


src/core/interp.c
```diff
@@ -1510,6 +1510,11 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                     GET_REG(cur_op, 2).s, GET_REG(cur_op, 4).s, GET_REG(cur_op, 6).i64);
                 cur_op += 8;
                 goto NEXT;
+            OP(indexic_s):
+                GET_REG(cur_op, 0).i64 = MVM_string_index(tc,
```
MasterDuke17: Shouldn't this be MVM_string_index_ignore_case()?

jnthn: Looks like.

src/strings/ops.c
```diff
@@ -644,7 +642,34 @@ MVMint64 MVM_string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *haysta
     }
     return 1;
 }
+MVMint64 MVM_string_index_ignore_case(MVMThreadContext *tc, MVMString *haystack, MVMString *needle, MVMint64 start) {
+    size_t index           = (size_t)start;
+    MVMStringIndex hgraphs = MVM_string_graphs(tc, haystack), ngraphs = MVM_string_graphs(tc, needle);
```
jnthn: Should not do these assignments until after the `MVM_string_check_arg` calls below, otherwise may dereference a NULL pointer or read out of bounds.

src/strings/ops.c
```diff
@@ -644,7 +642,34 @@ MVMint64 MVM_string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *haysta
...
+
+    if (start < 0 || start >= hgraphs)
+        return -1;
+
+    if (ngraphs > hgraphs || ngraphs < 1)
+        return -1;
 
+    /* brute force for now. horrible, yes. halp. */
+    while (index <= hgraphs - ngraphs) {
+        if (MVM_string_equal_at_ignore_case(tc, haystack, needle, index)) {
```
jnthn: I believe `MVM_string_equal_at_ignore_case` allocates due to doing an `fc` operation on `needle`. Thus `MVMROOT` of both haystack and needed are required. The most efficient way is to place them around the whole loop, however the loop should then not `return`, but rather set a result and break, otherwise we'd not clear the rooted things from the temporary stack.

src/core/interp.c
samcv: Fixed

src/strings/ops.c
samcv: Should be fixed now.

src/strings/ops.c
samcv: This should be fixed now.

jnthn:
The one Travis failure was just a connection error, so all goods good.

Issue #545, opened 2017-03-05
creating lots of really big numbers uses a ton of ram


MasterDuke17:
E.g., `my ($x,$y,$i) = (0,1,1); while ($i < 200000) {$y += $x ; $x = $y - $x ; $i += 1 } ; say $y;` allocates 3.6G. Some discussion with @timo here:
https://irclog.perlgeek.de/moarvm/2017-03-05#i_14206954

timo:
executive summary: a bigint takes only very little space in the nursery, but it can hold on to a gigantic buffer.

we probably want a solution similar to the "bytes promoted since last major collection" but instead of turning a minor into a major collection, we'd do a minor collection before the nursery has hit its limit.

of course, if we come up with something smarter, that'd be fantastic as well!

jnthn:
The cheapest way would just be to subtract some from the nursery alloc limit - that is, shorten the nursery on large allocations.

MasterDuke17:
You mean in src/core/alloc.h?

MasterDuke17:
For future reference, "I think when we produce a new bigint [in src/math/bigintops.c] (that's actually big), before we hand it back to the user, we [should] take the number of digits and [carefully] use that to influence the nursery size" - jnthn

MasterDuke17:
Fixed in 63430ac

PR #551, opened 2017-03-14
Shorten the nursery when creating large bigints

MasterDuke17:
A bigint takes only very little space in the nursery, but it can hold
onto a large buffer. When creating a bigint that's larger than a native,
decrease the nursery_alloc_limit if possible. The decreased nursery
means a garbage collection will happen sooner, cleaning up any unused
bigints and their buffers.

An initial stab at MoarVM issue #545 and RT #126450.

For this code `my ($x,$y,$i) = (0,1,1); while ($i < 200000) {$y += $x ; $x = $y - $x ; $i += 1 } ; say $y;`, `/usr/bin/time` reports:
`1129284maxresident` - before
`844512maxresident` - after

### DIFF
--- a/src/math/bigintops.c
+++ b/src/math/bigintops.c
@@ -359,6 +359,11 @@ MVMObject * MVM_bigint_##opname(MVMThreadContext *tc, MVMObject *result_type, MV
         mp_##opname(ia, ib, ic); \
         store_bigint_result(bc, ic); \
         clear_temp_bigints(tmp, 2); \
+        /*if (mp_count_bits(ic) > 65 && \
+            ((tc->nursery_alloc_limit - sizeof(mp_int)) > (tc->nursery_alloc + sizeof(mp_int)))) \
+        { \
+            tc->nursery_alloc_limit -= sizeof(mp_int); \
+        }*/ \
     } \
     else { \
         MVMint64 sc; \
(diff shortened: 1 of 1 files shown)


src/math/bigintops.c
```diff
@@ -359,6 +359,11 @@ MVMObject * MVM_bigint_##opname(MVMThreadContext *tc, MVMObject *result_type, MV
         mp_##opname(ia, ib, ic); \
         store_bigint_result(bc, ic); \
         clear_temp_bigints(tmp, 2); \
+        if (mp_count_bits(ic) > 65 && \
+            ((tc->nursery_alloc_limit - sizeof(mp_int)) > (tc->nursery_alloc + sizeof(mp_int)))) \
+        { \
+            tc->nursery_alloc_limit -= sizeof(mp_int); \
```
jnthn: This only adds a constant amount of pressure, rather than being proportional to the size of the bigint itself (which `mp_count_bits` can advise on since we already call it - however, I suspect there's a faster way, like just reading the `->used` or whatever it's called of the `mp_int`, which will give a number in bytes). I'd suggest a scheme like:

* Apply memory pressure equivalent to the size of the bigint storage, and mask the lower three bits so we don't lose 8-byte alignment (this freely handles bigints that use less than 8 bytes not causing any added pressure, which seems fair enough). Note the loss of alignment can cause crashes on some platforms, though on x86/x64 would just result in a slowdown.
* Put put an upper limit on the pressure of something like 32KB so we won't just GC madly on such values

MasterDuke17:
`108668maxresident` now

jnthn:
Looks good to me now; thanks! :-)

PR #556, opened 2017-03-19
Special case "\r\n" in MVM_nfg_is_concat_stable

MasterDuke17:
"\r\n" can never combine with anything else, so we can special-case it
in the check whether the relevant graphemes are synthetic.


samcv:
I'm probably going to change a lot of this section to more narrowly decide if normalization is needed;  but this change seems like it may be fine because of it being a possibly very common case.

How does this improve speed/memory with that JSON file?

MasterDuke17:
.5s faster, .3g less ram

samcv:
I'm probably going to change a lot of this section to more narrowly decide if normalization is needed;  but this change seems like it may be fine because of it be.  .x.   ing a possibly very common case.

How does this improve speed/memory with that JSON file?

Issue #513, opened 2017-01-28
Unboxing bigint to native has a problem with detecting "overflow to negative" boundary


MasterDuke17:
Some discussion here: http://irclog.perlgeek.de/perl6-dev/2016-09-08#i_13173656 and here: http://irclog.perlgeek.de/perl6-dev/2017-02-25#i_14164696 and here: http://irclog.perlgeek.de/moarvm/2017-02-26#i_14168048

Related to: https://github.com/rakudo/rakudo/pull/867

MasterDuke17:
Fixed in ad44958

PR #557, opened 2017-03-19
Correctly detect+handle overflow in mp_get_int64

MasterDuke17:
Handles the fact that for 64-bit 2's complement numbers the maximum
positive value is `2**63 - 1`, but the maximum negative value is `2**63`.

Uses the mp_get_long_long now available with the newer version of
libtommath. Also, instead of counting the bits every time, only do it
during the too-large-value error path.

Fixes issue #513. NQP passes `make m-test` and Rakudo passes `make m-spectest`, with two todo's now passing in t/spec/S02-types/int-uint.t and one todo now passing in t/spec/S03-operators/repeat.t.

### DIFF
--- a/src/6model/reprs/P6bigint.c
+++ b/src/6model/reprs/P6bigint.c
@@ -5,7 +5,7 @@
 #endif
 
 /* Handles both int and uint cases. */
-static MVMuint64 mp_get_int64(MVMThreadContext *tc, mp_int * a, int sign) {
+static MVMuint64 mp_get_int64(MVMThreadContext *tc, mp_int * a, int is_signed) {
     MVMuint64 res;
     MVMuint64 signed_max = 9223372036854775807ULL;
 
@@ -15,12 +15,12 @@ static MVMuint64 mp_get_int64(MVMThreadContext *tc, mp_int * a, int sign) {
 
     /* For 64-bit 2's complement numbers the positive max is 2**63-1
      * but the negative max is 2**63. */
-    if (sign && MP_NEG == SIGN(a)) {
+    if (is_signed && MP_NEG == SIGN(a)) {
         ++signed_max;
     }
 
     res = mp_get_long_long(a);
-    if (res == 0 || (sign && res > signed_max)) {
+    if (res == 0 || (is_signed && res > signed_max)) {
         /* The mp_int was either so big it overflowed a MVMuint64 or was bigger than a signed result could be. */
         MVM_exception_throw_adhoc(tc, "Cannot unbox %d bit wide bigint into native integer", mp_count_bits(a));
     }
--- a/src/6model/reprs/P6bigint.c
+++ b/src/6model/reprs/P6bigint.c
@@ -4,8 +4,8 @@
    #define MIN(x,y) ((x)<(y)?(x):(y))
 #endif
 
-/* Handles both int and uint cases. */
-static MVMuint64 mp_get_int64(MVMThreadContext *tc, mp_int * a, int is_signed) {
+/* Get a native int64 from an mp_int. */
+static MVMint64 mp_get_int64(MVMThreadContext *tc, mp_int * a) {
     MVMuint64 res;
     MVMuint64 signed_max = 9223372036854775807ULL;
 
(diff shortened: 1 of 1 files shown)


MasterDuke17:
```
loop (my int $i = 1; $i <= 10_000_000; $i = $i + 1) {
    my uint64 $a = 2**64-1; my int64 $b = 2**62; my int64 $c = -2**63
}
say now - INIT now
```
reported an average of about 3.5s before, 3.4s after

src/6model/reprs/P6bigint.c
```diff
@@ -92,10 +89,10 @@ static MVMint64 get_int(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, vo
     if (MVM_BIGINT_IS_BIG(body)) {
         mp_int *i = body->u.bigint;
         if (MP_NEG == SIGN(i)) {
-            return -mp_get_int64(tc, i);
+            return -mp_get_int64(tc, i, 1);
```
samcv: It would look better if there were a space between the `-` and the function to make it more visible.

MasterDuke17: I think I have a slight preference for as is, but don't mind changing if there's consensus. But FWIW, my branch that attempts to figure out a solution to the more general uint problem ends up re-writing this so there's no `-` at all.

src/6model/reprs/P6bigint.c
```diff
@@ -4,30 +4,27 @@
    #define MIN(x,y) ((x)<(y)?(x):(y))
 #endif
 
-/* A forced 64-bit version of mp_get_long, since on some platforms long is
- * not all that long. */
-static MVMuint64 mp_get_int64(MVMThreadContext *tc, mp_int * a) {
-    int i, bits;
+/* Handles both int and uint cases. */
+static MVMuint64 mp_get_int64(MVMThreadContext *tc, mp_int * a, int is_signed) {
```
jnthn: Maybe call this `mp_get_uint64` to match the return type?

src/6model/reprs/P6bigint.c
```diff
@@ -4,30 +4,27 @@
...
-    /* get number of digits of the lsb we have to read */
-    i = MIN(a->used,(int)((sizeof(MVMuint64)*CHAR_BIT+DIGIT_BIT-1)/DIGIT_BIT))-1;
-
-    /* get most significant digit of result */
-    res = DIGIT(a,i);
-
-    while (--i >= 0) {
-        res = (res << DIGIT_BIT) | DIGIT(a,i);
+    res = mp_get_long_long(a);
+    if (res == 0 || (is_signed && res > signed_max)) {
```
jnthn: What is the `== 0` part of this here for? Does `mp_get_long_long` reliably return 0 on overflow?

src/6model/reprs/P6bigint.c
MasterDuke17: The difference between name and return type did confuse me at first, since it is currently used for both.

I have some follow-on plans to break it out into a `mp_get_uint64` and `mp_get_int64`, should I also rename it now?

src/6model/reprs/P6bigint.c
MasterDuke17: It passed all my testing (ad-hoc + spectest), but I can't remember exactly what I did for the ad-hoc testing.

MasterDuke17: I just did some more testing, with values from `2**64` up to `2**99999999`, and `mp_get_long_long` always returned 0.

jnthn: OK, good.

src/6model/reprs/P6bigint.c
jnthn: Is there a reason not to break it out now? I think that'd be my preference.

MasterDuke17: I did test that, but I believe the problem was because of the lack of some other *_u ops. Some things end up calling the *_int version when they really mean/want the *_uint version, and then you run into the exact same problem that they can have different numbers of bits allowed.

MasterDuke17:
NQP and Rakudo still pass their respective tests, with two todo's now passing in t/spec/S02-types/int-uint.t and one todo now passing in t/spec/S03-operators/repeat.t.

PR #558, opened 2017-03-22
Optimize MVM_nfg_is_concat_stable to speed up concatenation of strings

samcv:
Here we special case '\r' ~ '\n' to ensure it will be normalized, since
it is a common possible case.

All other characters that have Grapheme_Cluster_Break=Control will
always break and do not change under normalization.

This speeds up concatenation greatly in a testcase with JSON::Fast
processing a file with all \r\n line endings. This speeds up JSON
parsing 39%, so the savings to the concatenation function itself are likely
greatly in excess of that for concatenation of strings whose borders
may be control characters.


samcv:
Happy with this. Going to merge.

PR #563, opened 2017-03-25
UTF-8 Provide a better error when failing to encode surrogates or out of bounds characters

samcv:
Check the general category is "Cs" (Control, Surrogates) and if so show
a more useful error message.

Share the exception code between both utf8 and utf8-c8
Have a new function MVM_exception_throw_utf8 so code does not have to
be repeated between utf8.c and utf8_c8.c implementations.

In addition add a seperate exception message if the codepoint that
is attempted to be encoded is > 0x10FFFF which is the legal limit for
UTF8

Roast has a [test](https://github.com/perl6/roast/blob/master/S29-conversions/ord_and_chr.t#L155) todo'd for [RT #124834](https://rt.perl.org/Ticket/Display.html?id=124834)

### DIFF
--- a/src/strings/utf8.c
+++ b/src/strings/utf8.c
@@ -468,11 +468,19 @@ char * MVM_string_utf8_encode_substr(MVMThreadContext *tc,
             result_pos += repl_length;
         }
         else {
+            const char *gencat = MVM_unicode_codepoint_get_property_cstr(tc, cp, MVM_UNICODE_PROPERTY_GENERAL_CATEGORY);
             MVM_free(result);
             MVM_free(repl_bytes);
-            MVM_exception_throw_adhoc(tc,
-                "Error encoding UTF-8 string: could not encode codepoint %d",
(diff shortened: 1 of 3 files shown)


samcv:
@jnthn what are your thoughts on this?

src/strings/utf8.c
```diff
@@ -470,9 +470,7 @@ char * MVM_string_utf8_encode_substr(MVMThreadContext *tc,
         else {
             MVM_free(result);
             MVM_free(repl_bytes);
-            MVM_exception_throw_adhoc(tc,
-                "Error encoding UTF-8 string: could not encode codepoint %d",
-                cp);
+            MVM_utf8_throw_encoding_exception(tc, cp);
```
jnthn: `MVM_string_utf8_throw_encoding_exception` would follow the naming convention of other things in the file better.

samcv:
Merging.

PR #565, opened 2017-03-28
Refactor eqatic and indexic and make Perl 6 :i regex 1.8x faster

samcv:
1.8x factor is dependent on not yet merged nqp code where it switches
MVM to use indexic instead of index + fc for locating sections of a string.

MVM_string_equal_at_ignore_case_INTERNAL_loop is the internal loop that
will be used be MVM_string_index_ignore_case. This will allow sharing
code between MVM_string_index_ignore_case and
MVM_string_equal_at_ignore_case and prevent the index function from getting
too messy.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -688,6 +688,10 @@ MVMint64 MVM_string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *haysta
     return 1;
 }
 MVMint64 MVM_string_index_ignore_case(MVMThreadContext *tc, MVMString *haystack, MVMString *needle, MVMint64 start) {
+    /* foldcase version of needle */
+    MVMString *needle_fc;
+    MVMStringIndex n_fc_graphs;
+
     size_t index           = (size_t)start;
     MVMStringIndex hgraphs, ngraphs;
     MVMint64 return_val = -1;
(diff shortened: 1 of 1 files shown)


src/strings/ops.c
```diff
@@ -619,71 +657,50 @@ MVMint64 MVM_string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *haysta
...
-        MVMROOT(tc, needle, {
-            /* brute force for now. horrible, yes. halp. */
-            while (index <= hgraphs) {
-                if (MVM_string_equal_at_ignore_case(tc, haystack, needle, index)) {
-                    return_val = (MVMint64)index;
-                    break;
-                }
-                index++;
-            }
+        MVMROOT(tc, needle_fc, {
```
zhuomingliang: `needle_fc = MVM_string_fc(tc, needle);` should move into here.

src/strings/ops.c
```diff
@@ -585,6 +585,44 @@ MVMint64 MVM_string_equal_at(MVMThreadContext *tc, MVMString *a, MVMString *b, M
         return 0;
     return MVM_string_substrings_equal_nocheck(tc, a, offset, bgraphs, b, 0);
 }
+MVM_STATIC_INLINE MVMint32 MVM_string_equal_at_ignore_case_INTERNAL_loop(MVMThreadContext *tc, MVMString *haystack, MVMString *needle_fc, MVMint64 h_start, MVMint64 h_graphs, MVMint64 n_graphs) {
```
zhuomingliang: Only global function starts with ```MVM_``` prefix

src/strings/ops.c
samcv: Thanks. Was not sure whether I needed to put needle_fc in MVMROOT, though I might not since I get the string as the result of the foldcase operation.

samcv:
@zhuomingliang made those two changes. Otherwise does it seem to be good? I have full spectest passing + some todo.

samcv:
@zhuomingliang Looks like the change of removing the MVMROOT caused all the indexic tests to fail. Removing it and putting that back the way it was.

src/strings/ops.c
zhuomingliang: Well, you are going to using `needle_fc` var again, but since `MVM_string_fc` may trigger a GC(which may moves address of objects), you need `MVMROOT` it(to get a new address of the object) , otherwise the address of `needle_fc` will be outdated, see https://github.com/samcv/MoarVM/blob/b9104537ce89bdee1d33905bb3de9e5a60457875/src/strings/ops.c#L654-L660 for example.

src/strings/ops.c
```diff
@@ -619,71 +657,50 @@ MVMint64 MVM_string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *haysta
...
-                return 0;
-        }
-        else if (h_fc_cps >= 1) {
-            for (j = 0; j < h_fc_cps; j++) {
-                n_g = MVM_string_get_grapheme_at_nocheck(tc, needle_fc, i + n_offset);
-                h_g = h_result_cps[j];
-                if (h_g != n_g)
-                    return 0;
-                n_offset++;
-            }
```
zhuomingliang: ```MVMROOT(tc, needle_fc, {`` is needless, since you already updated `needle_fc` by `MVM_string_fc`

zhuomingliang:
It should looks like 
``` 
    MVMROOT(tc, haystack, {
            needle_fc = MVM_string_fc(tc, needle);
    });
```

samcv:
@zhuomingliang yes,that is what I tried but then most of the tests failed which use these functions.

samcv:
Trying again. Seems to be passing NQP tests, going to make sure it passes roast too.

samcv:
Got passing on roast. just going to wait for Travis to check

samcv:
Got spectest pass. Any other concerns @zhuomingliang ? I also removed a couple of unused variables that were left over from the refactor.

zhuomingliang:
@samcv No other concerns from me now. 👍

src/strings/ops.c
```diff
@@ -649,41 +623,73 @@ MVMint64 MVM_string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *haysta
...
-            /* brute force for now. horrible, yes. halp. */
-            while (index <= hgraphs) {
-                if (MVM_string_equal_at_ignore_case(tc, haystack, needle, index)) {
-                    return_val = (MVMint64)index;
-                    break;
-                }
-                index++;
-            }
-        });
+            needle_fc = MVM_string_fc(tc, needle);
```
jnthn: An indentation level more than needed here? :-)

PR #567, opened 2017-03-28
Use PRIu64 in fprintf's to fix compiler warning RE incorrect types

samcv:
PR just to get Travis CI to run, will commit directly if all is fine.


samcv:
Reopening a new one with extraneous commits removed

Issue #561, opened 2017-03-24
SEGV from `perl6 -e 'run(:out, "bash").out.t.say'`


MasterDuke17:
Valgrind
```
================================================================================================
This is Rakudo Perl 6 running in valgrind, a tool for debugging and profiling programs.
Running a program in valgrind usually takes *a lot* more time than running it directly,
so please be patient.
This Rakudo version is 2017.03.28.ga.01.d.6794.d built on MoarVM version 2017.03.23.gc.4.de.954.b,
running on arch (4.9.3.1.arch) / linux (4.10.2.1.ARCH)
------------------------------------------------------------------------------------------------
==490== Memcheck, a memory error detector
==490== Copyright (C) 2002-2015, and GNU GPL'd, by Julian Seward et al.
==490== Using Valgrind-3.12.0 and LibVEX; rerun with -h for copyright info
==490== Command: /home/dan/Source/perl6/install/bin/moar --full-cleanup --execname=./perl6-valgrind-m --libpath=/home/dan/Source/perl6/install/share/nqp/lib --libpath=/home/dan/Source/perl6/install/share/nqp/lib --libpath=. /home/dan/Source/perl6/rakudo/perl6.moarvm --nqp-lib=blib -e run(:out,\ "bash").out.t.say
==490== 
==490== Jump to the invalid address stated on the next line
==490==    at 0x0: ???
==490==    by 0x4FE3095: MVM_io_is_tty (io.c:40)
==490==    by 0x4FB6584: MVM_interp_run (interp.c:4808)
==490==    by 0x5072E48: MVM_vm_run_file (moar.c:310)
==490==    by 0x400F02: main (main.c:212)
==490==  Address 0x0 is not stack'd, malloc'd or (recently) free'd
==490== 
==490== 
==490== Process terminating with default action of signal 11 (SIGSEGV): dumping core
==490==  Bad permissions for mapped region at address 0x0
==490==    at 0x0: ???
==490==    by 0x4FE3095: MVM_io_is_tty (io.c:40)
==490==    by 0x4FB6584: MVM_interp_run (interp.c:4808)
==490==    by 0x5072E48: MVM_vm_run_file (moar.c:310)
==490==    by 0x400F02: main (main.c:212)
==490== 
==490== HEAP SUMMARY:
==490==     in use at exit: 46,134,447 bytes in 196,803 blocks
==490==   total heap usage: 211,664 allocs, 14,861 frees, 55,595,692 bytes allocated
==490== 
[...]

PR #568, opened 2017-03-29
Add a check that introspection->is_tty is not NULL

MasterDuke17:
Since it is for pipes. Fixes MoarVM issue #561 

`perl6 -e 'run(:out, "bash").out.t.say'` now returns `False`.


jnthn:
It mighta been easier to just implement it on pipes. On the other hand, I/O will no doubt be in for a notable refactor when we move synchronous I/O away from libuv, so will just merge this.

MasterDuke17:
I meant to include this in the first comment, but here's some discussion with @timo and geekosaur about that very idea (implementing for pipes vs the check in this PR) https://irclog.perlgeek.de/moarvm/2017-03-28#i_14340977

PR #576, opened 2017-04-08
join: extra check for empty string + empty sep

timo:
this could cause \r and \n to be kept as two graphemes
when they were joined with an empty string in between and
the empty string as the separator.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -1222,11 +1222,10 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
             if (i > 0) {
                 /* If there's no separator and one piece is The Empty String we
                  * have to be extra careful about concat stability */
-                if (sgraphs == 0 && MVM_string_graphs(tc, piece) == 0) {
-                    if (!concats_stable)
-                        ;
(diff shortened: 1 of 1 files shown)


src/strings/ops.c
```diff
@@ -1220,9 +1220,18 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
 
             /* Add separator if needed. */
             if (i > 0) {
+                /* If there's no separator and one piece is The Empty String we
+                 * have to be extra careful about concat stability */
+                if (sgraphs == 0 && MVM_string_graphs(tc, piece) == 0) {
```
samcv: You can use MVM_string_graphs_nocheck here since MVM_string_graphs is already run on each piece in the loop above.

src/strings/ops.c
```diff
@@ -1220,9 +1220,18 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
...
+                    if (!concats_stable)
+                        ;
```
samcv: What is this empty block for? I think it would be more clear if the if statement on L1225  checked concat_stable and the `else if loop` below become just a plain if loop

PR #578, opened 2017-04-11
Fix mp_get_int64 and mp_get_uint64 for realz

MasterDuke17:
mp_get_long_long does in fact overflow/wrap around when the mp_int is
too large. So bring back checking the number of bits in the mp_int
before calling it. Also, remove the redundant check for whether the
mp_int is 0, since the exact same check is done in mp_get_long_long.

Before https://github.com/MoarVM/MoarVM/pull/557, `sub (int $x) {dd $x}(999999999999999999999999999999999999999999999)` gave `Cannot unbox 150 bit wide bigint into native integer`. After that PR was merged, it then gave `802379605485813759`. This PR corrects it back to  `Cannot unbox 150 bit wide bigint into native integer`, while still keeping the other fixes from https://github.com/MoarVM/MoarVM/pull/557.

I'm not exactly sure how it happened, since I did do lots of testing before, but I guess I just used values that were multiples of the size needed to get the correct behavior I was testing. MasterDuke17--

### DIFF
--- a/src/6model/reprs/P6bigint.c
+++ b/src/6model/reprs/P6bigint.c
@@ -8,21 +8,27 @@
 static MVMint64 mp_get_int64(MVMThreadContext *tc, mp_int * a) {
     MVMuint64 res;
     MVMuint64 signed_max = 9223372036854775807ULL;
-
-    if (a->used == 0) {
-        return 0;
-    }
+    int bits = mp_count_bits(a);
 
     /* For 64-bit 2's complement numbers the positive max is 2**63-1
      * but the negative max is 2**63. */
     if (MP_NEG == SIGN(a)) {
+        if (bits > 64) {
+            MVM_exception_throw_adhoc(tc, "Cannot unbox %d bit wide bigint into native integer", bits);
+        }
         ++signed_max;
+    }
+	else {
+        if (bits > 63) {
+            MVM_exception_throw_adhoc(tc, "Cannot unbox %d bit wide bigint into native integer", bits);
+        }
     }
 
     res = mp_get_long_long(a);
-    if (res == 0 || res > signed_max) {
+
+    if (res > signed_max) {
         /* The mp_int was bigger than a signed result could be. */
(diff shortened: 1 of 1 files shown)


src/6model/reprs/P6bigint.c
```diff
@@ -8,41 +8,41 @@
...
-
-    if (a->used == 0) {
-        return 0;
-    }
+    int bits = mp_count_bits(a);
 
     /* For 64-bit 2's complement numbers the positive max is 2**63-1
      * but the negative max is 2**63. */
     if (MP_NEG == SIGN(a)) {
+        if (bits > 64) {
```
niner: Why do we accept a 64 bit number for a signed int64? Don't we always lose a bit of storage space to the sign? The comment also says that the max is 2**63-1, i.e. 63 bits.

src/6model/reprs/P6bigint.c
```diff
@@ -8,41 +8,41 @@
...
 }
 
 /* Get a native uint64 from an mp_int. */
 static MVMuint64 mp_get_uint64(MVMThreadContext *tc, mp_int * a) {
-    MVMuint64 res;
-
-    if (a->used == 0) {
-        return 0;
-    }
+    int bits = mp_count_bits(a);
```
niner: The compiler probably detects this already, but I'm still a fan of const int there.

src/6model/reprs/P6bigint.c
MasterDuke17: The positive max is 2**63-1, but the negative max is -(2**63), 64 bits.

src/6model/reprs/P6bigint.c
MasterDuke17: Ah, good idea, I'll make that change.

src/6model/reprs/P6bigint.c
niner: Ok. Since it confused me, it will probably confuse other causal readers, so can you please extend the comment to make clear that bits will be 64 for -2^63?

MasterDuke17: Sure, will do.

PR #580, opened 2017-04-15
Merge coverage into master

samcv:


samcv:
This should be ready for merge now. If @jnthn can make sure the scripts don't infringe on any licensing issues. Since they're not a main part of MoarVM and all sources where the `make-index.sh` (CC-by-SA) and `update-gh-pages.sh` (Creative Commons Zero v1.0 Universal [aka public domain dedication]).

PR #585, opened 2017-04-21
Fix broken multi when annotations file isn't given

zoffixznet:


zoffixznet:
Fixed in another commit

PR #588, opened 2017-04-30
Coerce s n

pmurias:
Make coercing a str to a num behave more like the .Num method.
Make it no longer os dependent.

### DIFF
--- a/src/strings/parse_num.h
+++ b/src/strings/parse_num.h
@@ -1,0 +1,1 @@
+MVMnum64 MVM_coerce_s_n_new(MVMThreadContext *tc, MVMString *s);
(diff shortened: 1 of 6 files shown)


src/strings/parse_num.h
```diff
@@ -0,0 +1 @@
+MVMnum64 MVM_coerce_s_n_new(MVMThreadContext *tc, MVMString *s);
```
zhuomingliang: `MVM_coerce_s_n` me thinks?

pmurias:
Fixed.

pmurias:
NQP tests where doing some incorrect conversions that where uncovered by this PR.
I fixed those tests.

jnthn:
Finally got around to reviewing this (sorry it took a while), and it looks good overall to me. Release is very soon, so suggest merging this just after that.

PR #589, opened 2017-05-03
Zero slot when popping an array

MasterDuke17:
Otherwise ghost elements are left.

Fixes https://rt.perl.org/Ticket/Display.html?id=131245

Previously `perl6 -e 'my @a=(1,2,3,4,5); @a.pop; @a.pop, @a[4]="hi"; say @a.join(",");'` would print `1,2,3,4,hi`, now it prints `1,2,3,,hi`.


src/6model/reprs/VMArray.c
```diff
@@ -599,6 +599,7 @@ static void pop(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *data
         default:
             MVM_exception_throw_adhoc(tc, "MVMArray: Unhandled slot type");
     }
+    zero_slots(tc, body, body->start + body->elems, body->start + body->elems + 1, repr_data->slot_type);
```
niner: Would you mind introducing a const MVMuint64 slot = body->start + body->elems? Probably won't change anything for performance, but it would reduce code duplication for sure :)

MasterDuke17: @niner sure, just updated.

PR #590, opened 2017-05-04
Remove arbitrary string repeat limit

MasterDuke17:
Instead make the limit the same as the max number of graphemes possible
in a string. This is ok because there's a later check that the size of
the resulting string (i.e., size of original string * count) is also
less than the max number of graphemes possible.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -464,7 +464,7 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
     total_graphs = (MVMuint64)agraphs + (MVMuint64)bgraphs;
     if (total_graphs > MAX_GRAPHEMES)
         MVM_exception_throw_adhoc(tc,
-            "Can't concatenate strings, required number of graphemes %"PRIu64" > max allowed of %u",
+            "Can't concatenate strings, required number of graphemes %"PRIu64" > max allowed of %lld",
              total_graphs, MAX_GRAPHEMES);
 
     /* Otherwise, we'll assemble a result string. */
@@ -561,7 +561,7 @@ MVMString * MVM_string_repeat(MVMThreadContext *tc, MVMString *a, MVMint64 count
     if (count < 0)
         MVM_exception_throw_adhoc(tc, "Repeat count (%"PRId64") cannot be negative", count);
     if (count > MAX_GRAPHEMES)
-        MVM_exception_throw_adhoc(tc, "Repeat count (%"PRId64") cannot be greater than max allowed number of graphemes %u", count, MAX_GRAPHEMES);
+        MVM_exception_throw_adhoc(tc, "Repeat count (%"PRId64") cannot be greater than max allowed number of graphemes %lld", count, MAX_GRAPHEMES);
 
     /* If input string is empty, repeating it is empty. */
     agraphs = MVM_string_graphs_nocheck(tc, a);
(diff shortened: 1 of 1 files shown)


src/strings/ops.c
```diff
@@ -557,8 +557,8 @@ MVMString * MVM_string_repeat(MVMThreadContext *tc, MVMString *a, MVMint64 count
         return a;
     if (count < 0)
         MVM_exception_throw_adhoc(tc, "repeat count (%"PRId64") cannot be negative", count);
-    if (count > (1 << 30))
-        MVM_exception_throw_adhoc(tc, "repeat count > %d arbitrarily unsupported...", (1 << 30));
+    if (count > 0xFFFFFFFF)
```
niner: Would be nice to have a named constant for this magic number. Especially when it's used in multiple places.

MasterDuke17: Good idea, just did.

MasterDuke17:
How crazy would it be to use bigints to calculate the number of graphemes requested in the error branch so it reports to correct value even if it would be bigger than a MVMuint64?

src/strings/ops.c
```diff
@@ -557,8 +558,8 @@ MVMString * MVM_string_repeat(MVMThreadContext *tc, MVMString *a, MVMint64 count
         return a;
     if (count < 0)
         MVM_exception_throw_adhoc(tc, "repeat count (%"PRId64") cannot be negative", count);
-    if (count > (1 << 30))
-        MVM_exception_throw_adhoc(tc, "repeat count > %d arbitrarily unsupported...", (1 << 30));
+    if (count > MAX_GRAPHEMES)
+        MVM_exception_throw_adhoc(tc, "repeat count (%"PRId64") cannot be > than max allowed number of graphemes %u", count, MAX_GRAPHEMES);
```
samcv: Can you use "greater than" instead of ">"? Thanks!

src/strings/ops.c
```diff
@@ -1,6 +1,7 @@
 #include "platform/memmem.h"
 #include "moar.h"
 #define MVM_DEBUG_STRANDS 0
+#define MAX_GRAPHEMES     0xFFFFFFFF
```
samcv: Can you add a comment here that 0xFFFFFFFF is the largest value that can be stored in MVMStringIndex (32bit-uint)

samcv: Also how do you feel about naming this MAX_STRING_INDEX instead of MAX_GRAPHEMES?

MasterDuke17: @samcv I'd rather keep it as MAX_GRAPHEMES. It's not being used for indexing in any way, but it is being used to ensure the resulting string has an allowed/valid value for `MVMStringBody.num_graphs`.

pmurias:
Instead of using bignum math maybe just mentioning the size of the original string and the count would be better?

MasterDuke17:
I could just do something `"Can't repeat string, required number of graphemes (%lu * %lu) is greater than max allowed of %u",`, so we don't have to worry about actually doing the multiplication and handling overflow.

samcv:
@MasterDuke17 that sounds like an alright idea. We could use the unicode `×` maybe? Would look nicer than a star. Though maybe people don't like that. Either way just throwing that out there.

Go ahead and have some variation on what you said in your most recent comment.

samcv:
@MasterDuke17 is this commit ready for merge now or do you still need to make changes?

MasterDuke17:
@samcv should be good to go.

PR #591, opened 2017-05-05
Add can_fit_into_8bit funct, put logic used many places into one funct

samcv:
There are 7 places we use checks for this within src/strings/ops.c
Writing out the conditionals which sometimes use || and sometimes
use &&. They also sometimes use >= sign and other times > signs.
Make it more consistent for ease of reading and clarity.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -38,6 +38,12 @@ static void copy_strands(MVMThreadContext *tc, const MVMString *from, MVMuint16 
         num_strands * sizeof(MVMStringStrand));
 }
 
+MVM_STATIC_INLINE int can_fit_into_8bit (MVMGrapheme32 g) {
+    return -128 <= g && g <= 127;
+}
+MVM_STATIC_INLINE int can_fit_into_ascii (MVMGrapheme32 g) {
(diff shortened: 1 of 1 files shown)


src/strings/ops.h
```diff
@@ -96,3 +96,5 @@ MVMint64 MVM_string_find_not_cclass(MVMThreadContext *tc, MVMint64 cclass, MVMSt
 MVMuint8 MVM_string_find_encoding(MVMThreadContext *tc, MVMString *name);
 MVMString * MVM_string_chr(MVMThreadContext *tc, MVMCodepoint cp);
 void MVM_string_compute_hash_code(MVMThreadContext *tc, MVMString *s);
+MVM_STATIC_INLINE int can_fit_into_8bit (MVMGrapheme32 g);
+MVM_STATIC_INLINE int can_fit_into_ascii (MVMGrapheme32 g);
```
zhuomingliang: local static function should not put into *.h

samcv:
Okay, @zhuomingliang, removed the new functions from the .h file. Good to go with this change?

zhuomingliang:
As long as test is passed, I think so, yes 👍

PR #592, opened 2017-05-08
Add at_pos support for VMHash

MasterDuke17:
With this, `nqp-m -e 'my %h := nqp::hash("a", "b", "c", "d");
say(nqp::atpos(%h, 0))'` returns "b".

NQP passes `make m-test`, haven't run a Rakudo spectest yet.

### DIFF
--- a/src/6model/reprs/MVMHash.c
+++ b/src/6model/reprs/MVMHash.c
@@ -68,6 +68,33 @@ static void gc_free(MVMThreadContext *tc, MVMObject *obj) {
         MVM_fixed_size_free(tc, tc->instance->fsa, sizeof(MVMHashEntry), tmp);
 }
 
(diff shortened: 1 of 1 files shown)


zhuomingliang:
-1, Because hash is not ordered.

lizmat:
@zhuomingliang: indeed, and that is the point!  If you want to take a random element from a VMhash with N elements, you need to run an NQP iterator for the rand(N) in NQP code.  This merely shifts it to C-land, where it can be faster.

MasterDuke17:
Closing after discussion here https://irclog.perlgeek.de/moarvm/2017-05-09#i_14554429 and here https://irclog.perlgeek.de/moarvm/2017-05-09#i_14555813

PR #595, opened 2017-05-11
Remove workaround for tommath issue $56

lizmat:
As fas as I know, we've upgraded tommath and the issue there is fixed.

If we didn't upgrade yet, maybe we should also have a libtommath upgrade?

This would make nqp::rand_I a lot faster again, so it can actually be used in many cases.


jnthn:
Indeed, it's [fixed there](https://github.com/libtom/libtommath/pull/56) and we have that commit in our libtommath fork also.

Issue #554, opened 2017-03-15
SEGV in t/spec/S17-promise/nonblocking-await.t


MasterDuke17:
Full output of all the stuff below is at https://gist.github.com/MasterDuke17/c1620e8aebe05c51dfa85e3489223d70.
It was in test 11.
```
    my $c = Channel.new;
    my @proms = (1..100).map: -> $i {
        start {
            await($c) + await($c)
        }
    };
    for 1..100 {
        $c.send($_);
    }
    $c.close;
    #?rakudo.jvm skip 'hangs on JVM'
    throws-like { await @proms }, X::Channel::ReceiveOnClosed,
        'Hundred of outstanding awaits on channels that gets closed works';
```
There seem to be two types of errors. The first is:
```
==31703== Process terminating with default action of signal 6 (SIGABRT): dumping core
==31703==    at 0x55DFA10: raise (in /usr/lib/libc-2.25.so)
==31703==    by 0x55E1139: abort (in /usr/lib/libc-2.25.so)
==31703==    by 0x50913A6: uv_mutex_destroy (thread.c:114)
==31703==    by 0x5011BE1: gc_free (ConcBlockingQueue.c:72)
==31703==    by 0x4FE1518: MVM_gc_collect_free_nursery_uncopied (collect.c:580)
==31703==    by 0x4FDDDC7: MVM_gc_global_destruction (orchestrate.c:530)
==31703==    by 0x50728E1: MVM_vm_destroy_instance (moar.c:397)
==31703== by 0x400D5C: main (main.c:215)
```
The second is:
```
==32198== Thread 17:
==32198== Invalid read of size 8
==32198==    at 0x4FBF215: MVM_interp_run (interp.c:300)
==32198==    by 0x4FCC63D: start_thread (threads.c:80)
==32198==    by 0x50912A6: uv__thread_start (thread.c:49)
==32198==    by 0x5C6A2E6: start_thread (in /usr/lib/libpthread-2.25.so)
==32198==  Address 0x93cec48 is 808 bytes inside a block of size 5,808 free'd
==32198==    at 0x4C2C14B: free (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==32198==    by 0x4FC4DF2: MVM_frame_destroy (frame.c:101)
==32198==    by 0x4FE14DA: MVM_gc_collect_free_nursery_uncopied (collect.c:572)
[...]

PR #599, opened 2017-05-19
protect the parameterization lookup list with a mutex

timo:
addresses #554 
but perhaps we can do better than a mutex?


jnthn:
This is a step in the right direction, but not sufficient; there's still two races:

* Reading the parameterizations array when it may be being written to
* Ending up with the same parameterization key in the array twice, when we need to have interning semantics, which is what the original comment was about

jnthn:
Cherry-picked this and fixed the conflict; will now do the remaining bits to make it fully safe.

jnthn:
Cherry pick commit ID is 86407034d0

PR #600, opened 2017-05-22
Add ability to get port from already bind-ed socket

cono:
# Overview
This change trying to simplify S32-io/IO-Socket-INET.t and add ability to bind on first un-used port.

# Description
There is a possibility to find first unused port. You just making a bind on port = 0 and with ability of getsockname call you can get what port is in use. Here is an example in C:

```C
#include <stdio.h>
#include <errno.h>
#include <sys/types.h>
#include <sys/socket.h>
#include <strings.h>
#include <arpa/inet.h>
#include <unistd.h>

int
main(int argc, char **argv) {
	int s;
	struct sockaddr_in addr;
	socklen_t s_size;

	s = socket(AF_INET, SOCK_STREAM | SOCK_NONBLOCK, 0);

	if (s < 0) {
		perror("Can't create socket");
		return errno;
	}

	printf("AF_INET: %d, AF_INET6: %d\n", AF_INET, AF_INET6);
	bzero(&addr, sizeof(addr));
	addr.sin_family = AF_INET;
	addr.sin_addr.s_addr = INADDR_ANY;
	addr.sin_port = 0;

	if (bind(s, (struct sockaddr *) &addr, sizeof(addr)) < 0) {
		perror("Can't bind to address");
		return errno;
	}

	s_size = sizeof(addr);
	if (getsockname(s, (struct sockaddr *) &addr, &s_size) < 0) {
		perror("can't getsockname");
		return errno;
	}

	if (close(s) < 0) {
		perror("Can't close socket");
		return errno;
	}

	printf("port: %d\n", ntohs(addr.sin_port));

	return 0;
}
```
# Working example

IPv6:
```
% perl6 -e 'my $l = IO::Socket::INET.new(:listen, :localhost<localhost>, :localport(0)); qx{netstat -lnp 2>/dev/null| grep moar}.chomp.say; $l.localport.say'
tcp6       0      0 ::1:45073               :::*                    LISTEN      25558/moar          
45073
```

IPv4:
```
% perl6 -e 'my $l = IO::Socket::INET.new(:listen, :localhost<127.0.0.1>, :localport(0)); qx{netstat -lnp 2>/dev/null| grep moar}.chomp.say; $l.localport.say'
tcp        0      0 127.0.0.1:46149         0.0.0.0:*               LISTEN      25573/moar          
46149
```

# Reference
Pull-request of NQP: https://github.com/perl6/nqp/pull/360
Pull-request of Rakudo: https://github.com/rakudo/rakudo/pull/1086
Pull-request of spec-tests: https://github.com/perl6/roast/pull/270


src/io/io.c
```diff
@@ -374,6 +374,18 @@ void MVM_io_bind(MVMThreadContext *tc, MVMObject *oshandle, MVMString *host, MVM
...
+MVMint64 MVM_io_get_port(MVMThreadContext *tc, MVMObject *oshandle) {
+    MVMOSHandle *handle = verify_is_handle(tc, oshandle, "bind");
+    if (handle->body.ops->sockety) {
+        uv_mutex_t *mutex = acquire_mutex(tc, handle);
+        MVMint64 result = handle->body.ops->sockety->get_port(tc, handle);
+        release_mutex(tc, mutex);
+        return result;
+    }
+    else
+        MVM_exception_throw_adhoc(tc, "Cannot bind this kind of handle");
```
niner: Error message looks copy&pasted

cono: uh, thanks. Indeed! Let me fix this.

src/core/oplist
```diff
@@ -800,6 +800,8 @@ unicmp_s            w(int64) r(str) r(str) r(int64) r(int64) r(int64) :pure
 setdispatcherfor    r(obj) r(obj)
 getstrfromname       w(str) r(str) :pure
 indexic_s            w(int64) r(str) r(str) r(int64) :pure
+get_port_sk          w(int64) r(obj)
```
jnthn: We don't put `_` inside opnames except (a) a prefix to specify they are not normal ops (for example `sp_` for specialized ops) or a suffix to indicate type (for example `_i`). So this should be renamed to `getport_sk`.

src/io/syncsocket.c
```diff
@@ -353,6 +353,39 @@ static void socket_bind(MVMThreadContext *tc, MVMOSHandle *h, MVMString *host, M
...
+    uv_os_fd_t fd;
+
+    struct sockaddr_storage name;
+    int error, len = sizeof(struct sockaddr_storage);
+    MVMint64 port = 0;
+
+    error = uv_fileno((uv_handle_t *) socket, &fd);
+
+    if (error != 0)
+        MVM_exception_throw_adhoc(tc, "Failed to get fileno from uv handle: %s", uv_strerror(error));
```
jnthn: After the merge of the `sync-socket-no-uv` then most of the above goes away, because `->handle` becomes (on UNXI-y things) a file descriptor; on Windows it's a `SOCKET`, and the type `Socket` is typedef's to the appropriate thing. So this should be able to be simplified after the merge.

PR #604, opened 2017-05-26
new "hardware_concurrency" op

robertlemmen:
among other things, this can be used in rakudo to auto-size the default
thread pool


robertlemmen:
related: https://github.com/perl6/nqp/pull/362, https://github.com/rakudo/rakudo/pull/1088

src/core/oplist
```diff
@@ -927,3 +927,5 @@ prof_allocated   .s r(obj)
 ctw_check        .s r(obj) int16
 
 coverage_log     .s str int32 int32 int64
+
+hardware_concurrency w(int64) :pure
```
jnthn: Please read the instructions at the top of `oplist` to see how to correctly place new ops in this file. Also, the name should not contain a `_`, we only use that for type suffixes and dynamic op prefixes. I'd prefer a name like `cpucores` or so, which makes clearer what it is.

src/core/threads.c
```diff
@@ -262,3 +263,13 @@ void MVM_thread_join_foreground(MVMThreadContext *tc) {
         }
     }
 }
+
+MVMint64 MVM_cpucores() {
```
zhuomingliang: Don't we have a  `MVM_platform_cpu_count`?https://github.com/MoarVM/MoarVM/blob/2eedba88d2b0c08f94dd4426a60d0bc7f8effa13/src/platform/sys.h

robertlemmen: right, I did not see MVM_platform_cpu_count. it appears unused in moar, and does not map to an op. but could of course be used by some other libmoar client. The implementation within libuv seems more portable and complete, than the one in MVM_platform_cpu_count.

how about I use MVM_platform_cpu_count for the "cpucores" op, but switch the implementation to the libuv-based one? comments?

zhuomingliang: +1 from me.

jnthn: Sounds sensible.

src/core/interp.c
```diff
@@ -5632,6 +5633,11 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                 cur_op += 20;
                 goto NEXT;
             }
+            OP(cpucores): {
```
zhuomingliang: Here is a wrong position, please follow the order in `src/core/oplist`

robertlemmen: fixing, there seem to be a few other cases like this. do you guys usually clean up cosmetic stuff  as part of another commit when you come across it, or do it separately? also: do you squash your PRs?

zhuomingliang: yeah, like https://github.com/MoarVM/MoarVM/commit/cc59c5e505c0fd6e0a44d4a7cd8e17a094f9942c

zhuomingliang: We don't squash any commit.

jnthn: It's fine if people wish to squash/rebase/tidy their commits before they're merged, but yeah, we don't insist on it. :-)

Issue #610, opened 2017-07-06
Do not set -Werror=declaration-after-statement for 3rd-party libs


zoffixznet:
Moved from https://rt.perl.org/Ticket/Display.html?id=131657#ticket-history

> the problem here is that we have a flag to ensure gcc explodes at us when we write code on linux that'd immediately explode on windows due to MSVC being somewhat archaic, but that flag is also set for third-party code, and since the arm dyncall stuff only gets tested once every few months at best, we never noticed. this most likely wants a fix that throws out the flag for third party code in moarvm's repository until then you could just remove all occurences of -Werror=declaration-after-statement from all "Makefile" files and re-build

stmuk:
I don't think this is strictly a moar issue since I'm able to build on build moar *standalone* as of 4e29e4c7 fine on a Raspberry PI 3 (gcc version 4.9.2/Raspbian 8/armv7l).

I can reproduce the original issue  by trying to build moarvm via nqp (as well as rakudo).

Continuing to investigate.

stmuk:
Ignore last post. I was using "make install" standalone on ARM which oddly builds fine since CFLAGS isn't passed to dyncall  in this case but *can* reproduce the original issue via "Configure --make-install" .

I can't reproduce this (probably broken) behaviour on amd64 and have no explanation other than wondering if  the ARM toolchain (make?) is buggy (the original GCC shipped with raspbian couldn't even build moar).

There is a explicit fix for the ARM issue in https://github.com/MoarVM/MoarVM/pull/613

jnthn:
Sounds like #613 resolves the issue, so closing this.

PR #613, opened 2017-07-10
Fix ARM build

stmuk:
Don't pass CFLAGS (which includes -Werror=declaration-after-statement)
to dyncall. Just pass -fPIC. Otherwise we see a fatal warning with
dyncall_callvm_arm32_arm_armhf.c


stmuk:
Also see comments on https://github.com/MoarVM/MoarVM/issues/610

PR #615, opened 2017-07-18
Don't re_nfg string a + string b when concatenating in many cases, instead add only the section needing renormalization in as another strand

samcv:
PR is ready for review, spectest passes, was wondering if people wanted to take a look and make there's nothing code quality wise that needs changing. Thanks.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -17,7 +17,7 @@ static void check_strand_sanity(MVMThreadContext *tc, MVMString *s) {
     }
     if (len != MVM_string_graphs(tc, s))
         MVM_exception_throw_adhoc(tc,
-            "Strand sanity check failed (stand length %d != num_graphs %d)",
+            "Strand sanity check failed (strand length %d != num_graphs %d)",
             len, MVM_string_graphs(tc, s));
 }
 #define STRAND_CHECK(tc, s) check_strand_sanity(tc, s);
@@ -38,6 +38,16 @@ static void copy_strands(MVMThreadContext *tc, const MVMString *from, MVMuint16 
     assert(from->body.storage_type == MVM_STRING_STRAND);
     assert(to->body.storage_type == MVM_STRING_STRAND);
     memcpy(
+        to->body.storage.strands + to_offset,
+        from->body.storage.strands + from_offset,
+        num_strands * sizeof(MVMStringStrand));
+}
+/* Move strands inside the same strand string. */
+static void move_strands(MVMThreadContext *tc, const MVMString *from, MVMuint16 from_offset,
+        MVMString *to, MVMuint16 to_offset, MVMuint16 num_strands) {
+    assert(from->body.storage_type == MVM_STRING_STRAND);
+    assert(to->body.storage_type == MVM_STRING_STRAND);
+    memmove(
         to->body.storage.strands + to_offset,
         from->body.storage.strands + from_offset,
         num_strands * sizeof(MVMStringStrand));
@@ -453,10 +463,13 @@ static MVMint32 final_strand_matches(MVMThreadContext *tc, MVMString *a, MVMStri
     return 0;
 }
 MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString *b) {
-    MVMString *result;
+    MVMString *result = NULL, *renormalized_section = NULL;
+    int renormalized_section_graphs = 0, consumed_a = 0, consumed_b = 0;
     MVMuint32  agraphs, bgraphs;
     MVMuint64  total_graphs;
-
+    int lost_strands          = 0;
+    int is_concat_stable      = 0;
+    int index_ss_b;
     MVM_string_check_arg(tc, a, "concatenate");
     MVM_string_check_arg(tc, b, "concatenate");
 
@@ -468,6 +481,25 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
     if (bgraphs == 0)
         return a;
 
+    is_concat_stable = MVM_nfg_is_concat_stable(tc, a, b);
+    if (is_concat_stable == 0) {
+        MVMCodepoint last_a_first_b[2] = {
+            MVM_string_get_grapheme_at_nocheck(tc, a, a->body.num_graphs - 1),
+            MVM_string_get_grapheme_at_nocheck(tc, b, 0)
+        };
+        /* Needing both to be CCC = 0 can probably be relaxed some, but be careful optimizing */
+        if (0 <= last_a_first_b[0] && 0 <= last_a_first_b[1] && MVM_unicode_relative_ccc(tc, last_a_first_b[0]) == 0 && MVM_unicode_relative_ccc(tc, last_a_first_b[1]) == 0) {
+            MVMROOT(tc, a, {
+            MVMROOT(tc, b, {
+                renormalized_section = MVM_unicode_codepoints_c_array_to_nfg_string(tc, last_a_first_b, 2);
+                consumed_a = 1; consumed_b = 1;
+            });
+            });
+            if (agraphs == consumed_a && bgraphs == consumed_b)
+                return renormalized_section;
+            renormalized_section_graphs = MVM_string_graphs_nocheck(tc, renormalized_section);
+        }
+    }
     /* Total size of the resulting string can't be bigger than an MVMString is allowed to be. */
     total_graphs = (MVMuint64)agraphs + (MVMuint64)bgraphs;
     if (total_graphs > MAX_GRAPHEMES)
@@ -478,6 +510,8 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
     /* Otherwise, we'll assemble a result string. */
     MVMROOT(tc, a, {
     MVMROOT(tc, b, {
+    MVMROOT(tc, renormalized_section, {
+
         /* Allocate it. */
         result = (MVMString *)MVM_repr_alloc_init(tc, tc->instance->VMString);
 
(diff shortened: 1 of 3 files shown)


src/strings/ops.c
```diff
@@ -468,16 +471,36 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
     if (bgraphs == 0)
         return a;
 
+    is_concat_stable = MVM_nfg_is_concat_stable(tc, a, b);
+    if (is_concat_stable == 0) {
+        MVMCodepoint last_a_first_b[2];
+        last_a_first_b[0] = MVM_string_get_grapheme_at_nocheck(tc, a, a->body.num_graphs - 1);;
```
vendethiel: Minor: extra `;`

src/strings/ops.c
```diff
@@ -522,36 +545,93 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
             }
+            if (renormalized_section) {
+                int index_ss_re;
+                int index_ss_a = strands_a - 1;
+                /* Tweak the end of the last strand of string a. Since if a is made up of multiple strands, we can't just refer to index 0 and instead erfer to strands_a - 1 */
+                MVMStringStrand *ss_a = &(result->body.storage.strands[index_ss_a]);
+                MVMStringStrand *ss_re = NULL;
+                ss_a->end -= consumed_a;
+                /* If the strands ends up to be zero length we need to reduce the number of strand_index and also incease lost_strands so the next operation writes over it */
+                if (ss_a->start == ss_a->end)
```
vendethiel: The comment says "we need to reduce the number of strand_index", does this happen also by changing lost_strands?

src/strings/ops.c
```diff
@@ -522,36 +545,93 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
+    }
+    /* If it's regional indicator */
+    else if (is_concat_stable == 2) {
+        return re_nfg(tc, result);
+    }
+    else if (is_concat_stable == 0 && renormalized_section) {
+        return result;
+    }
+    /* We should have returned by now, but if we did return re_nfg */
+    return re_nfg(tc, result);
```
vendethiel: typo: didn't

src/strings/ops.c
```diff
@@ -522,36 +545,93 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
+
+                /* Add the renormalized section in as a strand */
+                ss_re->blob_string = renormalized_section;
+                ss_re->start       = 0;
+                ss_re->end         = renormalized_section->body.num_graphs;
+                ss_re->repetitions = 0;
+                if (ss_re->start == ss_re->end) {
+                    MVM_gc_root_temp_pop(tc);
+                    MVM_exception_throw_adhoc(tc, "Unexpected error in concatenation: renormalized_section is 0 graphemes.\n");
+                    lost_strands++;
```
vendethiel: Seems like dead code?

src/strings/ops.c
```diff
@@ -468,16 +471,36 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
         return a;
 
+    is_concat_stable = MVM_nfg_is_concat_stable(tc, a, b);
+    if (is_concat_stable == 0) {
+        MVMCodepoint last_a_first_b[2];
+        last_a_first_b[0] = MVM_string_get_grapheme_at_nocheck(tc, a, a->body.num_graphs - 1);;
+        last_a_first_b[1] = MVM_string_get_grapheme_at_nocheck(tc, b, 0);
+        /* Needing both to be CCC = 0 can probably be relaxed some, but be careful optimizing */
+        if (0 <= last_a_first_b[0] && 0 <= last_a_first_b[1] && MVM_unicode_relative_ccc(tc, last_a_first_b[0]) == 0 && MVM_unicode_relative_ccc(tc, last_a_first_b[1]) == 0) {
+            renormalized_section = MVM_unicode_codepoints_c_array_to_nfg_string(tc, last_a_first_b, 2);
```
jnthn: This can allocate, and so may invalidate `a` and `b`, which may be read after.

src/strings/ops.c
samcv: Yes it happens by changing lost_strands. Which records the strands we lost so we can correct the indexes further on in the code.

src/strings/ops.c
```diff
@@ -468,16 +471,36 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
+            MVM_gc_root_temp_push(tc, (MVMCollectable **)&(renormalized_section));
```
jnthn: It's not clear what is being protected with the push/pop pair here. `MVM_string_graphs` is just an accessor for the number of graphemes, so doesn't need it.

src/strings/ops.c
```diff
@@ -468,16 +471,36 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
+            consumed_a = 1; consumed_b = 1;
+            renormalized_section_graphs = MVM_string_graphs(tc, renormalized_section);
```
jnthn: This isn't used in the path that returns and could be moved afterwards.

src/strings/ops.c
jnthn: (Yes, it's probably needed later, but could be done later as a result, perhaps allowing the use of the easier-to-check `MVMROOT`.)

src/strings/ops.c
```diff
@@ -468,16 +471,36 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
+    if (total_graphs > MAX_GRAPHEMES) {
+        if (renormalized_section) MVM_gc_root_temp_pop(tc);
         MVM_exception_throw_adhoc(tc,
             "Can't concatenate strings, required number of graphemes %"PRIu64" > max allowed of %lld",
              total_graphs, MAX_GRAPHEMES);
+    }
 
     /* Otherwise, we'll assemble a result string. */
     MVMROOT(tc, a, {
     MVMROOT(tc, b, {
```
jnthn: I'd suggest `MVMROOT` of `renormalized_section` here; just be sure to set it to NULL at the point of declaration, then we needn't worry about doing it conditionally.

src/strings/ops.c
samcv: Changed it to a comment instead of the dead code.

src/strings/ops.c
samcv: Fixed

src/strings/ops.c
samcv: Fixed

src/strings/ops.c
```diff
@@ -522,36 +545,93 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
+                index_ss_re = strands_a - lost_strands;
+                ss_re = &(result->body.storage.strands[index_ss_re]);
+
+                /* Add the renormalized section in as a strand */
+                ss_re->blob_string = renormalized_section;
+                ss_re->start       = 0;
+                ss_re->end         = renormalized_section->body.num_graphs;
+                ss_re->repetitions = 0;
+                if (ss_re->start == ss_re->end) {
+                    MVM_gc_root_temp_pop(tc);
```
jnthn: Throwing an exception pops all roots automatically; this is not needed.

src/strings/ops.c
```diff
@@ -468,16 +471,36 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
+        if (renormalized_section) MVM_gc_root_temp_pop(tc);
         MVM_exception_throw_adhoc(tc,
             "Can't concatenate strings, required number of graphemes %"PRIu64" > max allowed of %lld",
              total_graphs, MAX_GRAPHEMES);
+    }
 
     /* Otherwise, we'll assemble a result string. */
     MVMROOT(tc, a, {
     MVMROOT(tc, b, {
+
```
zhuomingliang: I think `MVMROOT(tc, renormalized_section  {` can be put into here, so you could remove all `MVM_gc_root_temp_push` and `MVM_gc_root_temp_pop` on `renormalized_section `

src/strings/ops.c
```diff
@@ -522,36 +545,93 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
+                ss_b->end         = effective_b->body.num_graphs;
+                ss_b->repetitions = 0;
+            }
+            if (renormalized_section_graphs) {
+                /* Tweak the beginning of the first strand of string b */
+                MVMStringStrand *ss_b = &(result->body.storage.strands[index_ss_b]);
+                ss_b->start += consumed_b;
+                if (ss_b->start == ss_b->end) {
+                    lost_strands++;
+                    copy_strands(tc, result, index_ss_b + 1, result, index_ss_b, strands_b - 1);
```
jnthn: At the moment `copy_strands` uses `memcpy` rather than `memmove`, and that can go badly wrong when [the source and destination overlap](http://man7.org/linux/man-pages/man3/memcpy.3.html).

src/strings/ops.c
samcv: Even if it is NULL in some conditions? renormalized_section is only a string in certain conditions, otherwise it is NULL.

jnthn: Yes, it's fine, as it is pushing the address of the variable.

src/strings/ops.c
samcv: This is not ok but calling MVM_nfg_is_concat_stable is fine?

src/strings/ops.c
```diff
@@ -522,36 +556,94 @@ MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString
...
+                result->body.num_graphs += renormalized_section_graphs - consumed_b - consumed_a;
             }
         }
     });
     });
-
+    });
     STRAND_CHECK(tc, result);
-    return MVM_nfg_is_concat_stable(tc, a, b) ? result : re_nfg(tc, result);
+    /* Make sure to pop after STRAND_CHECK but before returning */
```
jnthn: I think this comment is a leftover from how the code was previously?

PR #616, opened 2017-07-19
Start off collapsing strands as 8bit instead of 32bit

samcv:
Here instead of starting as 32-bit and then at the end converting
the already iterated 32-bit blob to 8bit, we instead start off
converting as 8-bit and as soon as we see a character that won't
fit, we do the following:

1. Store the old blob pointer and previous value of i
2. Set the string type to 32bit and allocate space for the 32-bit blob
3. Copy the data so far copied from the 8bit blob into the new 32-bit blob
   since it's faster than setting up the grapheme iterator again and starting
   over
4. Free the 8-bit blob
5. Once that data is copied we store the grapheme which interupted our
   previous iteration. Once this is done we are able to continue where
   we left off with the original iterator

This greatly speeds up when we only have data that fits in 8-bits without
impacting the 32-bit strings too much.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -76,24 +76,37 @@ static MVMString * collapse_strands(MVMThreadContext *tc, MVMString *orig) {
     });
     ographs                      = MVM_string_graphs(tc, orig);
     result->body.num_graphs      = ographs;
-    result->body.storage_type    = MVM_STRING_GRAPHEME_32;
-    result->body.storage.blob_32 = MVM_malloc(ographs * sizeof(MVMGrapheme32));
+    result->body.storage_type    = MVM_STRING_GRAPHEME_8;
+    result->body.storage.blob_8  = MVM_malloc(ographs * sizeof(MVMGrapheme8));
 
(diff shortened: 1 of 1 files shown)


samcv:
### Here are some code and times or the new vs the old code

#### Code:
```perl6
# 8bit test:
use nqp; nqp::indexingoptimized(("abcde" x 100_000_000) ~ "fo")
32bit test:
# use nqp; nqp::indexingoptimized(("abcde" x 100_000_000) ~ "fo»")
```
#### Times:
```
32bit times:
now: 1.0827963
before: 0.81091399

8bit times: 
now: 0.7526170
before: 1.18165355
```

jnthn:
I wonder to what degree we can take the inputs being 8-bit as an indicator that this is likely to work out? Or maybe we can use a flag in the header to indicate "known as needing 32-bit". Probably quite a lot of tweaks we could do, but this seems fairly safe to merge now.

samcv:
Yeah maybe it would be a good idea for us to have a flag on strings so we can choose easier and faster. That would be pretty nice. Going to merge this now because of the benefit.

PR #617, opened 2017-07-19
attempt at a better error message for RT#126341

dwarring:
I got hit by the same vexing message this morning. Didn't discover the real cause until I searched the RT queue and found this ticket.

Is this wording better?


jnthn:
Probably can be tweaked further, but if this would have helped you figure the problem out sooner then it's worth including.

PR #626, opened 2017-08-05
Add nqp::coveragecontrol op

MasterDuke17:
If `MVM_COVERAGE_CONTROL` is set (along with `MVM_COVERAGE_LOG`), line
converage has to be started with a call to `nqp::coveragecontrol(1)` and
can be stopped with a call to `nqp::coveragecontrol(0)`. Additionally,
when `MVM_COVERAGE_CONTROL` is set the coverage report is no longer
de-duplicated.


timo:
:+1:

src/core/ops.h
```diff
@@ -768,82 +768,84 @@
...
-#define MVM_OP_sp_boolify_iter_hash 834
-#define MVM_OP_prof_enter 835
-#define MVM_OP_prof_enterspesh 836
-#define MVM_OP_prof_enterinline 837
-#define MVM_OP_prof_enternative 838
-#define MVM_OP_prof_exit 839
-#define MVM_OP_prof_allocated 840
-#define MVM_OP_ctw_check 841
-#define MVM_OP_coverage_log 842
+#define MVM_OP_<<<<<<< 767
```
samcv: Looks like this should be an op name instead of these `<<<<<<<`

src/moar.c
```diff
@@ -282,6 +282,13 @@ MVMInstance * MVM_vm_create_instance(void) {
             instance->coverage_log_fh = fopen_perhaps_with_pid(coverage_log, "a");
         else
             instance->coverage_log_fh = stderr;
+
+        instance->coverage_control = 0;
+        if (getenv("MVM_COVERAGE_CONTROL")) {
+            char *coverage_control = getenv("MVM_COVERAGE_CONTROL");
+            if (coverage_control && strlen(coverage_control))
```
samcv: I just updated the rest of the file to remove unneeded strlen's https://github.com/MoarVM/MoarVM/commit/911b6323de2cc3f55b92268a88141669c8857d89

Change this line to
```c
if (coverage_control && coverage_control[0])
```

MasterDuke17:
Should be good now.

PR #630, opened 2017-08-11
Add unix socket support

the-eater:
This merge add support for unix sockets, since we don't want Moar ops to change, we use the upper range from `port` as it's not used (port is 64 bits int, while in reality max port is 65535)

### DIFF
--- a/src/io/syncsocket.c
+++ b/src/io/syncsocket.c
@@ -6,6 +6,8 @@
 #else
     #include "unistd.h"
     #include <sys/socket.h>
+    #include <sys/un.h>
+
     typedef int Socket;
     #define closesocket close
 #endif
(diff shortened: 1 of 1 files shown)


src/io/syncsocket.c
```diff
@@ -228,27 +230,63 @@ static void gc_free(MVMThreadContext *tc, MVMObject *h, void *d) {
...
  * but it's a thing to remember if someone feels like peeking inside the returned struct. */
 struct sockaddr * MVM_io_resolve_host_name(MVMThreadContext *tc, MVMString *host, MVMint64 port) {
     char *host_cstr = MVM_string_utf8_encode_C_string(tc, host);
     struct sockaddr *dest;
-    struct addrinfo *result;
     int error;
+    int flags = 0;
+    struct addrinfo *result;
+#ifndef _WIN32
+    struct sockaddr_un *result_un = MVM_malloc(sizeof(struct sockaddr_un));
```
BenGoldberg1: `result_un` is malloced and never freed.
Data is copied into it, then copied into a (newly malloc-ed) `dest`, and `dest` is returned, and `result_un` is thrown away.
This seems silly.
It seems to me that `result_un` should be declared (and malloced) _inside_ of the `if (flags & AF_UNIX) {` block, and returned _instead_ of `dest`.

the-eater: I guess, like this?

src/io/syncsocket.c
```diff
@@ -228,27 +230,62 @@ static void gc_free(MVMThreadContext *tc, MVMObject *h, void *d) {
...
+            return sizeof(struct sockaddr_un);
+#endif
+        default:
+            return sizeof(struct sockaddr);
+    }
+}
+
 /* Actually, it may return sockaddr_in6 as well; it's not a problem for us, because we just
  * pass is straight to uv, and the first thing it does is it looks at the address family,
  * but it's a thing to remember if someone feels like peeking inside the returned struct. */
```
samcv: Can you add a comment to explain the newly added flags? Thanks.

src/io/syncsocket.c
```diff
@@ -228,27 +230,62 @@ static void gc_free(MVMThreadContext *tc, MVMObject *h, void *d) {
...
     struct sockaddr *dest;
-    struct addrinfo *result;
     int error;
+    int flags = 0;
+    struct addrinfo *result;
     char port_cstr[8];
+
+    if (port > 65535) {
+        flags = (port >> 16);
+    }
```
jnthn: We don't need to do the conditional and stuff here; can just always have `int flags = port >> 16`.

jnthn:
Looks good overall.

the-eater:
Thanks :) but still I still need to make the IO::Socket::Async also support this, with which I was kinda fighting with, and haven't found time to work on it. may have this weekend. will try to resolve all comments also :)

jnthn:
Cool, thanks. A separate PR for the async sockets after this one would be better anyway; easier to review one at a time. :+1:

src/io/syncsocket.c
the-eater: Is this okay?

src/io/syncsocket.c
the-eater: Fixed :)

the-eater:
@jnthn should be fixed now, only 1 check fails because apt fails .-.

jnthn:
Thanks, looks good to me. :+1:

PR #632, opened 2017-08-13
Alloc asyncsocket read buffer based on amount last read

MasterDuke17:
Instead of using the suggested_size (which is almost always 65536
according to http://docs.libuv.org/en/v1.x/handle.html#data-types), use
the next power of two greater than the amount last read (which has been
added to the ReadInfo struct).

Passes NQP's `make m-test` and Rakudo's `make m-test m-spectest`.

### DIFF
--- a/src/io/asyncsocket.c
+++ b/src/io/asyncsocket.c
@@ -13,11 +13,27 @@ typedef struct {
     int               seq_number;
     MVMThreadContext *tc;
     int               work_idx;
+    size_t            last_read;
 } ReadInfo;
 
 /* Allocates a buffer of the suggested size. */
 static void on_alloc(uv_handle_t *handle, size_t suggested_size, uv_buf_t *buf) {
-    size_t size = suggested_size > 0 ? suggested_size : 4;
+    ReadInfo *ri = (ReadInfo *)handle->data;
+    size_t size  = ri->last_read ? ri->last_read : 64;
+
+    if (size < 128) {
+        size = 128;
+    }
+    else {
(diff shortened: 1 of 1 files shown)


src/io/asyncsocket.c
```diff
@@ -13,11 +13,27 @@ typedef struct {
...
 static void on_alloc(uv_handle_t *handle, size_t suggested_size, uv_buf_t *buf) {
-    size_t size = suggested_size > 0 ? suggested_size : 4;
+    ReadInfo *ri = (ReadInfo *)handle->data;
+    size_t size  = ri->last_read ? ri->last_read : 64;
+
+    if (size < 128) {
+        size = 128;
+    }
+    else {
+        /* Equivalent to `size = pow(2, ceil(log2(size + 1)))`, but C89 compatible. */
```
MasterDuke17: Thanks, nice catch.

jnthn:
I'm not quite sure if this has so much value with sockets as it has with processes. With processes, line buffered output is pretty common and so we want to tune it down. But incoming packet sizes will typically be far more noisy, and the last one is likely not a particularly good indicator for the next one in plenty of common traffic patterns I've seen. So I'm a bit reluctant on this one.

jnthn:
Oops, didn't mean to close it; happy for further input/discussion.

MasterDuke17:
@jnthn thanks for looking. Do you have any suggestions for what to benchmark it with?

jnthn:
@MasterDuke17 Consider a persistent HTTP session. Responses will have highly variable sizes.

MasterDuke17:
I never could find a compelling enough benchmark for this, so closing.

PR #633, opened 2017-08-13
Make sure libffi callbacks can be reused.

markmont:
Rakudo test t/04-nativecall/21-callback-other-thread.t tests 2..9
fail with a segmentation fault or glibc error (double free or
malloc() problem) when MoarVM is built with --has-libffi.

The problem is due to the callback handler freeing the memory for
its first argument (the libffi call interface, cif).  Doing this
makes the callback not reusable (the callback still exists in
tc->native_callback_cache).  Fix the problem by not freeing
the memory.


markmont:
It looks like the native callback cache is never destroyed, is this correct?

For each entry in the cache, *types and target are garbage collected, but
not cb.  If cb is not GC'd (dcbFreeCallback() does not get called on it
in nativecall_dyncall.c), then cif should not be either
(in nativecall_libffi.c).

PR #635, opened 2017-08-14
Fix typo in MVM_6model_container_atomic_store

vendethiel:
In containers.c
The function checked for cs->atomic_load being present, but then went on using cs->atomic_store.


jnthn:
Nice catch!

PR #643, opened 2017-08-15
Add W->!X JIT support for NetBSD 8, SELinux with deny_execmem, PaX

markmont:
When MoarVM is compiled with --has-libffi, make the code produced by
the JIT compiler writable at an address but then never executable
through the same address (W->!X).

This allows MoarVM to be installed and used on systems that enforce
W->!X without any special filesystem permissions (NetBSD 8) or
writing special security policy (SELinux with deny_execmem). This
does not offer additional security beyond what is provided by
default when MoarVM uses mprotect(); it just makes MoarVM easier to
install and use on systems that require W->!X.

### DIFF
--- a/src/jit/compile.c
+++ b/src/jit/compile.c
@@ -10,7 +10,8 @@ static const MVMuint16 MAGIC_BYTECODE[] = { MVM_OP_sp_jit_enter, 0 };
 
 MVMJitCode * MVM_jit_compile_graph(MVMThreadContext *tc, MVMJitGraph *jg) {
     dasm_State *state;
-    char * memory;
+    char * memory;       /* writable */
+    char * exec_memory;  /* executable but not writable */
     size_t codesize;
     /* Space for globals */
     MVMint32  num_globals = MVM_jit_num_globals();
@@ -65,8 +66,20 @@ MVMJitCode * MVM_jit_compile_graph(MVMThreadContext *tc, MVMJitGraph *jg) {
 
     /* compile the function */
     dasm_link(&state, &codesize);
+#ifdef HAVE_LIBFFI
+    /* Map one area of memory at two addresses, one writable and one
+     * executable */
+    memory = ffi_closure_alloc(codesize, (void **) &exec_memory);
+    if (!memory)
+        MVM_panic(1, "Unable to allocate memory for jit");
+#else
+    /* Map memory that starts writable and is later made executable */
     memory = MVM_platform_alloc_pages(codesize, MVM_PAGE_READ|MVM_PAGE_WRITE);
+    exec_memory = memory;
+#endif
     dasm_encode(&state, memory);
+
+#ifndef HAVE_LIBFFI
     /* set memory readable + executable */
     if (!MVM_platform_set_page_mode(memory, codesize, MVM_PAGE_READ|MVM_PAGE_EXEC)) {
         MVM_jit_log(tc, "Setting jit page executable failed or was denied. deactivating jit.\n");
@@ -76,15 +89,18 @@ MVMJitCode * MVM_jit_compile_graph(MVMThreadContext *tc, MVMJitGraph *jg) {
         tc->instance->jit_enabled = 0;
         return NULL;
     }
-
+#endif
 
     MVM_jit_log(tc, "Bytecode size: %"MVM_PRSz"\n", codesize);
     /* Create code segment */
     code = MVM_malloc(sizeof(MVMJitCode));
-    code->func_ptr   = (MVMJitFunc)memory;
-    code->size       = codesize;
-    code->bytecode   = (MVMuint8*)MAGIC_BYTECODE;
-    code->sf         = jg->sg->sf;
+    code->func_ptr        = (MVMJitFunc)exec_memory;
+#ifdef HAVE_LIBFFI
+    code->writable_memory = memory;
+#endif
+    code->size            = codesize;
+    code->bytecode        = (MVMuint8*)MAGIC_BYTECODE;
+    code->sf              = jg->sg->sf;
 
     /* Get the basic block labels */
     code->num_labels = jg->num_labels;
@@ -93,7 +109,7 @@ MVMJitCode * MVM_jit_compile_graph(MVMThreadContext *tc, MVMJitGraph *jg) {
         MVMint32 offset = dasm_getpclabel(&state, i);
         if (offset < 0)
             MVM_jit_log(tc, "Got negative offset for dynamic label %d\n", i);
-        code->labels[i] = memory + offset;
+        code->labels[i] = exec_memory + offset;
     }
 
     /* Copy the deopts, inlines, and handlers. Because these use the label index
@@ -123,7 +139,11 @@ MVMJitCode * MVM_jit_compile_graph(MVMThreadContext *tc, MVMJitGraph *jg) {
 }
 
 void MVM_jit_destroy_code(MVMThreadContext *tc, MVMJitCode *code) {
+#ifdef HAVE_LIBFFI
+    ffi_closure_free(code->writable_memory);
+#else
     MVM_platform_free_pages(code->func_ptr, code->size);
+#endif
     MVM_free(code->labels);
     MVM_free(code->bb_labels);
     MVM_free(code->deopts);
--- a/src/jit/compile.h
+++ b/src/jit/compile.h
@@ -2,6 +2,9 @@ typedef MVMint32 (*MVMJitFunc)(MVMThreadContext *tc, MVMCompUnit *cu, void * lab
 
 struct MVMJitCode {
     MVMJitFunc func_ptr;
+#ifdef HAVE_LIBFFI
+    char      *writable_memory;
+#endif
     size_t     size;
     MVMuint8  *bytecode;
 


markmont:
To implement W->!X on systems where it is necessary, libffi will
create and unlink a temporary file which gets mapped to two separate
addresses, one address writable and one executable.  Only a single
file and file descriptor is used regardless of how many pieces of
JIT code are compiled.

On systems where W->!X is not necessary, libffi will use mprotect()
or just regular malloc().

The libffi function that implements W->!X, ffi_closure_alloc(), is
already used in MoarVM to implement nativecall callbacks when MoarVM
is compiled with --has-libffi.  Using ffi_closure_alloc() again for
for JIT compilation does not increase the risk as far as the temp
file creation is concerned, especially since libffi will use
O_TMPFILE on systems and filesystems which support it.

src/jit/compile.c
```diff
@@ -123,7 +139,11 @@ MVMJitCode * MVM_jit_compile_graph(MVMThreadContext *tc, MVMJitGraph *jg) {
 }
 
 void MVM_jit_destroy_code(MVMThreadContext *tc, MVMJitCode *code) {
+#ifdef HAVE_LIBFFI
+    ffi_closure_free(code->writable_memory);
+#else
     MVM_platform_free_pages(code->func_ptr, code->size);
+#endif
```
jnthn: Does the exec memory not have to be freed also, or does `ffi_closure_free` take care of that?

markmont: Correct, only the writable memory needs to be freed, not the exec memory, as they are actually the same memory just accessed via different mappings.

The exec memory and the writable memory are actually each just a base address and an offset into a mapping created via `mmap`.  These mappings are persistent across the life of the program.

`ffi_closure_alloc` uses a special allocator, `dlmalloc`, to manage memory via the writable mapping.  Memory is only allocated once but is visible through the exec mapping as well as through the writable mapping.  `ffi_closure_free` frees the memory via the writable mapping by calling `dlfree` at which point the exec memory address should no longer be used since it points (via the exec mapping) to memory which has been freed.

bdw:
Hi.
First of all, let me say that I very much appreciate the effort you put into working this out, and that I fully support the notion that user experience for MoarVM users on all platforms should be as frictionless as possible.

Unfortunately I feel like I must object to this patch as it is, on the following grounds:

+ I'm very worried about introducing separate behaviour to the JIT compiler based not on any compilation flag which we might set on the JIT itself but rather on a third-party library the user has chosen to install. This feels strongly like dangerous and opaque action at a distance to me, not to mention tight coupling between this and an unrelated library, that we might have to support now for a long time.
   + I'd much prefer this behaviour to be defined with it's own compilation flag and documentation, e..g --jit-use-dualmap or something. And perhaps refactor this into an API that doesn't expose switching permissions on the same memory, but rather allows an 'exchange' between writeable and executable memory, which would amount to a mprotect on normal systems, and a dualmap-dance on 'hardened' systems.
+ Strictly speaking this introduces a potential security issue for us, since we now have a writeable and executable pointer to the same memory, and updates between the two are shared. That is not currently the case, and I especially don't feel well about introducing that for users that don't see any benefit.
+ We rely on the function ffi_closure_alloc and ffi_closure_free, both of which have *three* separate implementations in libffi ([1](https://github.com/libffi/libffi/blob/master/src/closures.c#L197), [2](https://github.com/libffi/libffi/blob/master/src/closures.c#L831), [3](https://github.com/libffi/libffi/blob/master/src/closures.c#L874)), the first of which is not compatible with the usage here (the trampoline section is a fixed-size structure within another structure, and which appears to be compiled on `__mach__` exporting platforms, i.e. mac os x. Or at least, that is how it looks like to me.
+ For the second implementation (which I *think* is what we will actually run on most systems) I can't readily figure out how it works based on reading  the dlmalloc source code. This worries me, dualmapping should not be hard, see [this example](https://stackoverflow.com/questions/25400693/c-linux-dual-map-memory-with-different-permissions) on stack overflow.
+ If a user is working on a 'hardened' system that should prevent the execution of program-written memory, then I can't see how writing to a (temporary/unlinked) file or shared memory, and then mapping that as executable, is going to keep that security expectation. Indeed the fact that a user runs a W!->X system would seem to be inconsistent with  them caring much about JIT-compiler optimized performance, and hacking around it by dualmapping or any other method seems like the antisocial thing to do.
   + More to the point, the speed advantage of using the JIT is currently not so great that perl6 becomes unusable without it.
+ Both SELinux and NetBSD provide escape mechanisms to users that do care about JIT performance after all. It seems that the path of least surprise for these users is to document the issue and list possible solutions, potentially adding them to the build system.
   + Although arguably that may also be 'antisocial'.

So unfortunately I cannot support this PR as is.

I will not 'veto' it - it is not in my power to do so, and I don't want that power - but I hope you can see my points, and maybe we can find another way to help these users.

Thanks again for your effort.

markmont:
Hi, Bart -- thanks for your long and thoughtful response.  I'm willing to re-implement libffi's closures.c and dlmalloc.c in MoarVM and craft the API to do the 'exchange' between writable and executable memory as you describe above, if there is general consensus that this is the best thing to do and the code will be accepted into MoarVM.  I'll wait for the consensus before beginning since this would be significant work that would have to be done outside of my $dayjob.

Here are a few comments on your comments:

* Regarding introducing a potential security issue: There is [significant disagreement](https://bugzilla.mozilla.org/show_bug.cgi?id=506693) on this, but, according to the NetBSD and SELinux security experts, having two different addresses (one writable, one executable) is preferable to having a single address that starts writable and is later changed to executable, due to [race conditions](https://www.internetsociety.org/sites/default/files/09_2_2.pdf) and other concerns ([1](https://bugzilla.mozilla.org/show_bug.cgi?id=506693#c11), [2](https://bugzilla.mozilla.org/show_bug.cgi?id=506693#c54), [3](https://bugzilla.mozilla.org/show_bug.cgi?id=506693#c109)).
* The following two points are important since if I reimplement the libffi functionality in MoarVM, they would likely pertain to MoarVM's implementation, too:
    * The most appropriate of the three implementations of ffi_closure_alloc and ffi_closure_free is chosen at the time libffi is compiled, based on the security requirements of the target system.  The choices include just using malloc/free, using mprotect, and using dual mapping of a temp file.  This has the advantage that libffi won't do things which are more complicated than necessary.
    * The mapped regions are set up by [dlmmap_locked](https://github.com/libffi/libffi/blob/master/src/closures.c#L688) which is only a little more complicated than the StackOverflow example you provided.  dlmmap_locked is called by [dlmmap](https://github.com/libffi/libffi/blob/master/src/closures.c#L744) only if the system security requires it; dlmmap in turn is called from [dlmalloc](https://github.com/libffi/libffi/blob/master/src/dlmalloc.c#L3223).

bdw:
Hi Mark,

Thanks for the links, they are very interesting. I'll be honest, the stance of *some* developers and/or security experts make my blood boil.... but that is a discussion best left to another day!

There is one minor but important detail that we have glossed over so far. The papers and/or solutions all seem to presume that we require run-time updates of the generated code. The spidermonkey JIT does this. This is not true for the MoarVM JIT. Once executable, it is *never* made writeable again, not by *any* thread. 

The argument that a 'bad guy' can use a call to mprotect to made memory read-writeable again is also suspicious.... For that to work, the attacker has to:

+ Overwrite the stack (presumably by a stack-buffer-overflow) to insert an mprotect, and overwrite registers so that the arguments are *just so* that mprotect does what you mean.
+ And overwrite the heap (presumably by a heap-buffer overflow, but there are other ways...) to overwrite the now-writeable JIT memory. *Before* this code is ever entered again, because then the program won't work.
+ And overwrite the stack further so that the same memory is now executable again.

This is becoming a _really_ difficult attack, I think. 

On the other hand, if I have a (shared) mapping between a RW and RX piece of memory, all I need to do is find one heap-buffer-overflow - and keep on writing until I happen to find a writeable memory region that happens to be shared-mapped to an executable memory. And then I'm golden, because I now just have to wait until the interpreter enters that code. And both are reasonably probable, because a significant part of my memory region are going to represent compiled code. We could protect against that by surrounding the writeable memory with a region PROT_NONE so we get a segmentation violation, but even that is not going to be safe.

Furthermore, if I generate a private file with mkstemp as given in [this example](https://www.akkadia.org/drepper/selinux-mem.html) then I can not prevent another process from opening the file before I unlink it, and once the other process has an open file descriptor, I'm not sure how I'm going to prevent them from writing there, or mapping it themselves. I guess I could try and protect against *that* with file locking, but that would then set me up for a denial-of-service attack......

So I think you'll see why I think having RW/RX toggling is by far the preferred option on all platforms that support it.

Anyway, on to the more productive part of the discussion:
+ As mentioned before, I'd be fine with this patch being contained in its own API, e.g. a MVMJitExecMem structure, potentially embedded into the MVMJitCode structure, and such methods:
   - void MVM_jit_execmem_init(MVMThreadContext *tc, MVMJitExecMem *exec_mem, size_t size);
   - void * MVM_jit_execmem_get_writeable(MVMThreadContext *tc, MVMJitExecMem *exec_mem);
  - void * MVM_jit_execmem_get_executable(MVMThreadContext *tc, MVMJitExecMem *exec_mem);
   - void MVM_jit_execmem_deinit(MVMThreadContext *tc, MVMJItExecMem *exec_mem);
+ If we did that, I'd like to have a separate configuration flag and definition flag to switch to it. I'm partial to `--jit-security-theater` and `#ifdef MVM_JIT_SECURITY_THEATER` but that might be needlessly provocative.
+ If we have the code 'contained' like this, I'm okay with using the libffi `ffi_closure_alloc` function, but I'd prefer a handrolled solution that took into account the fact that we don't in fact need writeable memory, so we can do:
```
    fd = mkstemp(pattern);
    /* ... */
    void * writeable = malloc(code_size);
    dasm_encode(&dasm_state, writeable);
    for (left = code_size; left >= 0; left -= written) {
        written = write(fd, writeable, code_size);
        if (written < 0) { die(errno); } /* etc */
    }
    /* and only then */
    executable = mmap(NULL, code_size, PROT_READ|PROT_EXEC, MAP_PRIVATE, fd, 0);
    /* because there is really no reason to have it writeable */
    fclose(fd);
```

Or indeed any other such scheme that avoids having a read-writeable mapping active at the same time. Alternatively - as a 'performance improvement', we could map multiple executable segments on the same file descriptor, using the offset argument to mmap.

Again, thanks for bringing this to our attention, and for all the effort you've put in so far. I really appreciate it.

markmont:
Per [discussion in IRC](https://irclog.perlgeek.de/moarvm/2017-08-17#i_15031625), I'm closing this PR and will submit a new one in a week or so that abstracts out the W^X functionality for both the libffi dual-mapping case and the no-libffi mprotect case, behind a clean API.  This will keep ifdef's out of the JIT code and allow us to more easily in-source the functionality someday if we ever want to get away from libffi.

The new PR will also provide an explicit way to enable/disable W->!X dual-mapping, either via a Configure.pl flag and/or environment variable.

PR #644, opened 2017-08-15
Fix typo

MasterDuke17:


jnthn:
Conflicted after other changes, but done the equivalent fix.

PR #648, opened 2017-08-16
Implement the Knuth-Morris-Pratt string search algorithm

samcv:
We use this in MVM_string_index for:
* Strings with non-matching types
 ** That have more than one codepoint in the needle
 ** That don't have a needle more than 100 graphemes long


src/strings/ops.c
```diff
@@ -877,6 +882,42 @@ static MVMint64 string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *Hay
...
+                : j;
+        }
+        else j = next[j];
+    }
+}
+static MVMint64 knuth_morris_pratt_index (MVMThreadContext *tc, MVMString *pat, MVMString *MVM_text, MVMint64 H_offset) {
+    MVMint64 pat_offset = 0, text_offset = H_offset;
+    MVMStringIndex MVM_text_graphs = MVM_string_graphs(tc, MVM_text);
+    MVMStringIndex pat_graphs      = MVM_string_graphs(tc, pat);
+    MVMGrapheme32 *next = alloca(pat_graphs * sizeof(MVMGrapheme32));
```
bdw: You're doing the allocation before throwing an exception

src/strings/ops.c
```diff
@@ -877,6 +882,42 @@ static MVMint64 string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *Hay
...
+
+    if (knuth_morris_pratt_max_pattern_length < pat_graphs) {
+        MVM_exception_throw_adhoc(tc, "Tried to search a pattern %i long but can only do %i\n", pat_graphs, knuth_morris_pratt_max_pattern_length);
+    }
```
bdw: This seems like it'd be an internal error, and not something the user should be able to cause (and therefore catch). So an exception seems like it's not the right thing, and maybe a form of assert() instead.

src/strings/ops.c
```diff
@@ -877,6 +882,42 @@ static MVMint64 string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *Hay
         return H_graphs + H_expansion - H_offset >= n_fc_graphs  ? 1 : 0;
     return 0;
 }
+static void knuth_morris_pratt_process_pattern (MVMThreadContext *tc, MVMString *pat, MVMGrapheme32 next[], MVMStringIndex pat_graphs) {
+    MVMint64 i = 0;
+    MVMint64 j = next[0] = -1;
+    while (i < pat_graphs) {
+        if (j == -1 || MVM_string_get_grapheme_at_nocheck(tc, pat, i) == MVM_string_get_grapheme_at_nocheck(tc, pat, j)) {
```
bdw: Seems like with so many calls to `MVM_string_get_grapheme_at_nocheck` it might be beneficial to flatten the pattern into a MVMGrapheme32[] and access that instead.

samcv:
@bdw the reason for the max pattern length is that the longer the pattern, the more space we have to allocate on the stack for the jump table (which is the same length as the needle itself). So it will take extra time to process the pattern before starting the search. I'm not sure what limits we should have on the length of the pattern, and eventually we may be able to only process the start of the needle and only process the rest of the needle in case we find a match for that length of the needle (so we can avoid having to process a very long needle in full, when that needle may not end up matching).

bdw:
Cool, that makes sense.
On the other hand, I'd think the win for this would be most with a relatively long needle and a much longer haystack. Because the bigger the needle, the greater the possible skip will be. So maybe the ratio could be a good heuristic?

samcv:
@bdw hmm we could maybe move to ratio eventually. I did some tests and above 10,000 length needle, and a 100,000 length haystack, it takes more time to do the KMP algorithm. Though taking the time to flatten the needle could be contributing to that, so what I would recommend is for us to figure out when flattening the needle is not worth it time wise, and then once we have a cutoff regarding that, to then figure out some ratio for deciding if processing the whole needle is worth it or not.

For now i'm setting the max pattern size at 4069 as it's a little less than half the 10,000 mark which the KMP algorithm started to not be faster.

PR #652, opened 2017-08-19
Add error string to 'failed to resolve' exception

MasterDuke17:


src/io/syncsocket.c
```diff
@@ -252,7 +252,8 @@ struct sockaddr * MVM_io_resolve_host_name(MVMThreadContext *tc, MVMString *host
     }
     else {
         char *waste[] = { host_cstr, NULL };
-        MVM_exception_throw_adhoc_free(tc, waste, "Failed to resolve host name '%s'", host_cstr);
+        MVM_exception_throw_adhoc_free(tc, waste, "Failed to resolve host name '%s'. Error was: '%s'",
```
samcv: Can you change this from "Error was:" to just "Error:"?

MasterDuke17:
Updated.

PR #660, opened 2017-08-24
JIT param_op_(i|o)

MasterDuke17:
These were causing `INTERPOLATE` to bail. Follow the example of the recent
JITting of `param_rp_(i|o)` (e4e9c04).

NQP passes `make m-test` and Rakudo passes `make m-test m-spectest`.

`INTERPOLATE` now doesn't bail in the jit log, but still isn't being JITted according
to a profile, I don't know why.

### DIFF
--- a/src/core/interp.c
+++ b/src/core/interp.c
@@ -1013,10 +1013,10 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
             }
             OP(param_op_i):
             {
-                MVMArgInfo param = MVM_args_get_optional_pos_int(tc, &tc->cur_frame->params,
+                MVMint64 param = MVM_args_get_optional_pos_int(tc, &tc->cur_frame->params,
                     GET_UI16(cur_op, 2));
-                if (param.exists) {
-                    GET_REG(cur_op, 0).i64 = param.arg.i64;
+                if (param) {
+                    GET_REG(cur_op, 0).i64 = param;
                     cur_op = bytecode_start + GET_UI32(cur_op, 4);
                 }
                 else {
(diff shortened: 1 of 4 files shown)


MasterDuke17:
For anyone who hasn't been following #moarvm recently, this PR incorrectly JITs these ops. I'm just leaving it open for now for reference to a possible correct solution.

MasterDuke17:
@bdw @niner I think I first understood this to be incorrect because it doesn't adjust the runloop in the JITted case, but now I think that's not the right reason, though I'm not sure what is. Mind explaining?

src/core/interp.c
```diff
@@ -1013,10 +1013,10 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
             }
             OP(param_op_i):
             {
-                MVMArgInfo param = MVM_args_get_optional_pos_int(tc, &tc->cur_frame->params,
+                MVMint64 param = MVM_args_get_optional_pos_int(tc, &tc->cur_frame->params,
                     GET_UI16(cur_op, 2));
-                if (param.exists) {
-                    GET_REG(cur_op, 0).i64 = param.arg.i64;
+                if (param) {
```
niner: if (param) means if (param != 0) making it impossible to pass a perfectly good 0 as positional int parameter.

src/core/interp.c
```diff
@@ -1053,11 +1053,11 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
             OP(param_op_o):
             {
                 MVMuint16 arg_idx = GET_UI16(cur_op, 2);
-                MVMArgInfo param = MVM_args_get_optional_pos_obj(tc, &tc->cur_frame->params, arg_idx);
-                if (param.exists) {
-                    GET_REG(cur_op, 0).o = param.arg.o;
+                MVMObject *param = MVM_args_get_optional_pos_obj(tc, &tc->cur_frame->params, arg_idx);
+                if (param) {
```
niner: I'm not sure if object registers may be NULLed. Could be the difference between "a parameter was passed" and "a defined parameter was passed". But I think in the undefined case, we'd still pass a type object (!IS_CONCRETE), so for objects this may actually fly.

src/jit/graph.c
```diff
@@ -1664,6 +1664,26 @@ static MVMint32 jgb_consume_ins(MVMThreadContext *tc, JitGraphBuilder *jgb,
...
         break;
         /* Unspecialized parameter access */
+    case MVM_OP_param_op_i: {
+        MVMint16  dst     = ins->operands[0].reg.orig;
+        MVMuint16 arg_idx = ins->operands[1].lit_ui16;
+
+        MVMJitCallArg args[] = { { MVM_JIT_INTERP_VAR, { MVM_JIT_INTERP_TC } },
+                                 { MVM_JIT_INTERP_VAR, { MVM_JIT_INTERP_PARAMS } },
+                                 { MVM_JIT_LITERAL, { arg_idx } } };
+        jgb_append_call_c(tc, jgb, MVM_args_get_optional_pos_int, 3, args, MVM_JIT_RV_INT, dst);
```
niner: Now this unconditionally writes to the target register, even though the parameter may not have been passed at all.

MasterDuke17:
"jnthn: I'm still unconvinced by trying to JIT the param_* ops.
They mostly get specialized away, and I'd rather we fix the causes of them not being."

There have been some other questions about whether this is even needed/desired/may become outdated by a more complete re-working of JITting optionals anyway. It's also more complicated than its current incorrect state, so I'm just going to close this. The code that saw the bails originally now doesn't because of spesh.

PR #662, opened 2017-08-26
Generate stacktraces on Travis CI when getting a segfault

samcv:
On Linux we enable coredumps, and then set the coredumps folder to our
build directory. It is then set so that gdb will print out a stacktrace
from the coredumps if the build fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,6 +36,7 @@ addons:
       - llvm-5.0
       - expect
       - libffi-dev
+      - gdb
 
 before_script:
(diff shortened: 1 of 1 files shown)


samcv:
I am running Travis on my own forks. I made a PR because for some reason it was not segfaulting when I built it on Travis, but it would here on certain builds in the matrix.

It's still not triggering on the PR either, but I ended up getting it so Travis builds can now generate a core dump and then use use gdb to print a stacktrace at the very end of the build in case we get any segfaults.

PR #669, opened 2017-08-29
Cleanup legacy JIT

bdw:
Branch `even-moar-jit` contained numerous improvements and cleanups that affected the legacy JIT. Having these separate would mean that even-moar-jit could not not merge cleanly without developer involvement because changes to the legacy JIT would mean it diverted.
This branch separates out the legacy JIT changes from even-moar-jit, so that even-moar-jit can be rebased cleanly on top of them, and we have a simpler patch set to review.


src/jit/compile.c
```diff
@@ -1,83 +1,120 @@
...
 #include "platform/mmap.h"
-#include "emit.h"
 
-#define COPY_ARRAY(a, n, t) memcpy(MVM_malloc(n * sizeof(t)), a, n * sizeof(t))
+
+void MVM_jit_compiler_init(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJitGraph *jg);
+void MVM_jit_compiler_deinit(MVMThreadContext *tc, MVMJitCompiler *compiler);
+MVMJitCode * MVM_jit_compiler_assemble(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJitGraph *jg);
+
+#define COPY_ARRAY(a, n) memcpy(MVM_malloc(n * sizeof(a[0])), a, n * sizeof(a[0]))
```
jnthn: This would be preferable as a `MVM_STATIC_INLINE` rather than a macro.

src/jit/compile.c
```diff
@@ -88,61 +125,48 @@ MVMJitCode * MVM_jit_compile_graph(MVMThreadContext *tc, MVMJitGraph *jg) {
...
-    code->inlines      = code->num_inlines ? COPY_ARRAY(jg->inlines, jg->num_inlines, MVMJitInline) : NULL;
-
-    /* clear up the assembler */
-    dasm_free(&state);
-    MVM_free(dasm_globals);
+    /* Copy the deopts, inlines, and handlers. Because these use the
+     * label index rather than the direct pointer, no fixup is
+     * necessary */
+    code->num_deopts   = jg->deopts_num;
+    code->deopts       = code->num_deopts ? COPY_ARRAY(jg->deopts, jg->deopts_num) : NULL;
```
jnthn: Wonder if we should fold the zero check into `COPY_ARRAY` rather than repeating it here?

src/jit/compile.c
```diff
@@ -88,61 +125,48 @@ MVMJitCode * MVM_jit_compile_graph(MVMThreadContext *tc, MVMJitGraph *jg) {
...
+    code->num_handlers = jg->handlers_num;
+    code->handlers     = code->num_handlers ? COPY_ARRAY(jg->handlers, jg->handlers_alloc) : NULL;
+    code->num_inlines  = jg->inlines_num;
+    code->inlines      = code->num_inlines ? COPY_ARRAY(jg->inlines, jg->inlines_alloc) : NULL;
 
-    code->seq_nr = MVM_incr(&tc->instance->jit_seq_nr);
+    /* add sequence number */
+    code->seq_nr       = MVM_incr(&tc->instance->jit_seq_nr);
```
jnthn: Since the JIT now runs on the spesh thread only, an atomic increment ain't needed here.

src/jit/compile.h
```diff
@@ -1,20 +1,18 @@
...
+typedef void (*MVMJitFunc)(MVMThreadContext *tc, MVMCompUnit *cu, void * label);
 
 struct MVMJitCode {
     MVMJitFunc func_ptr;
     size_t     size;
     MVMuint8  *bytecode;
 
     MVMStaticFrame *sf;
-    /* The basic idea here is that /all/ label names are indexes into
+    /* The basic igdea here is that /all/ label names are indexes into
```
jnthn: I'm pretty sure it was an idea...

src/jit/pseudo.c
```diff
@@ -0,0 +1,490 @@
+#include "moar.h"
```
jnthn: Was this file meant to be included in this PR? It looks expr-JIT related.

jnthn:
Looks very good overall; left a few comments.

src/jit/pseudo.c
bdw: Nope, not sure how it got in, will remove it :-)

src/jit/compile.c
bdw: Agreed, and i was actually thinking of moving the assignment of the sequence number to spesh graph time. Will do that in some other commit though.

src/jit/compile.c
bdw: In that case it couldn't 'automatically' take over the types, which the macro can; it's really just a convenience though, so I don't care that much.

jnthn: Ah, good point. Can stay as a macro then.

PR #674, opened 2017-09-04
Merge 'expression' JIT backend

bdw:
Now that the changes to the legacy JIT have been applied and merged, this branch now contains (mostly) the addition of the expression JIT compiler.
Added functionality includes:

+ Expression tree building
+ Code generation by means of tree tiling
+ Full register selection via DynASM
+ A general register allocator
   + Although it can only handle forward-only graphs for now
+ Allocating values in local memory (for spilling)
+ Automated 'bad copmpilation' bisect process
+ A minor 'bitmap' library (used in the register allocator)

Not yet implemented, but planned:
+ An optimizer
+ REPR-specialized templates or tree generation
+ Multiple basic blocks per expression tree


src/core/vector.h
```diff
@@ -18,6 +18,13 @@
         x ## _alloc = 0; \
     } while (0)
 
+#define MVM_VECTOR_SIZE(x) \
+    sizeof(*x) * (x ## _alloc)
+
+#define MVM_VECTOR_TOP(x) \
+    (x) + (x ## _num)
```
vendethiel: those should use outer parentheses, so `MVM_VECTOR_TOP(x) * 5` works as intended

bdw: Done, thanks! although `MVM_VECTOR_TOP(x) * 5` makes little sense as that is a pointer :smile:

docs/jit/ir.md
```diff
@@ -0,0 +1,257 @@
...
+left-to-right order, meaning that textually later definitions can
+refer to earlier definition ames. An important side effect of 'let:'
+is that *let: guarantees the evaluation order of defintions*. That is,
+the tree node represented by '$val' is computed before its reference
+in the 'if' expression following it. (This is not the case for operand
+nodes). Finally, note that 'let:' statements declare definitions in a
+single expression-wide scope and that redefinitions are not allowed.
+
+
+**Statement Macro's** are lists of which the node name start with an
```
jnthn: s/'//

docs/jit/ir.md
```diff
@@ -0,0 +1,257 @@
...
+* substitutions are *linked* into the resulting tree, thus two
+  references of the same substitution refer to the same IR tree
+  nodes. Tree macro's are *spliced* into the tree, so two
+  invocations refer to two different trees. (These trees may be
+  resolved to be equal when using common subexpression evaluation, but
+  that's a separate issue).
+
+A tree macro is defined using the 'macro:' keyword, followed by the
+macro name (including the '^' prefix), a list of macro arguments, and
+the actual macro list. For example, the '^p6obody' macro is
```
jnthn: The '^spesh_slot` macro appears to be shown below?

docs/jit/ir.md
```diff
@@ -0,0 +1,257 @@
...
+implemented as:
+
+    (macro: ^spesh_slot (,a)
+       (idx
+          (load (addr (frame) (&offsetof MVMFrame effective_spesh_slots)))
```
jnthn: Also, `effective_spesh_slots` went away iirc, so this example is a tad out-dated.

docs/jit/tiles.md
```diff
@@ -0,0 +1,321 @@
...
+
+Let's focus on the first three, since it is my suspicion that these
+will be the most common (the latter two imply a STORE and a LOAD from
+the same operand, which is not directly supported in the memory model
+of the compiler, although it may be in the future).
+
+A tile description is a list that starts with the keyword
+<code>tile:</code>, followed by the tile *name*, the pattern proper,
+the *symbol* that it generates, and the (estimated) tile *cost*. There
+are two supported symbols, namely __reg__, __flag__ and __void__. The
```
jnthn: three? :)

docs/jit/tiles.md
```diff
@@ -0,0 +1,321 @@
...
+looks much like this:
+
+    /* in src/jit/x64/tiles.dasc, supposing we only care about 8 byte xor */
+    MVM_JIT_TILE_DECL(xor) {
+        int dst   = values[0]->reg_num;
+        int src_a = values[1]->reg_num;
+        int src_b = values[2]->reg_num;
+        if (values[0]->size < 8)
+            MVM_oops(tc, "oops!");
+        if (src_a != dsst) {
```
jnthn: s/dsst/dst/

docs/jit/tiles.md
```diff
@@ -0,0 +1,321 @@
...
+* <code>MVMJitExprValue **values</code> - an array of 'value' objects,
+  which holds (among other things) the register number and register
+  type allocated to the value. Significantly, <strong>there is a value
+  object for every symbol node in the pattern, starting from index 1;
+  and index 0 refers to the value generated by this node</strong>.
+* <code>MMVJitExprNode *args</code> - the arguments of this specific
+  node, if any.
+
+These values are thus available to any tile that is compiled. Often, a
+till will require data that is 'deeper' in the tree, for instance the
```
jnthn: s/till/tile/

src/core/bitmap.h
```diff
@@ -0,0 +1,62 @@
...
+    }
+    return i;
+}
+#endif
+
+
+/* NB - make this a separate 'library', use it for register bitmap */
+/* Witness the elegance of the bitmap for our purposes. */
+MVM_STATIC_INLINE void MVM_bitmap_set(MVMBitmap *bits, MVMint32 idx) {
+    bits[idx >> 6] |= (UINT64_C(1) << (idx & 0x3f));
```
jnthn: Why 6, out of curiosity? Are we keeping the upper bits of the byte for something?

src/jit/core_templates.expr
```diff
@@ -0,0 +1,229 @@
...
+                      (arglist (carg (tc) ptr)
+                               (carg $0 ptr))))
+
+(template: decont!
+    (ifv
+         (any (zr $1)
+              (^is_type_obj $1)
+              (zr (^getf (^stable $1) MVMSTable container_spec)))
+         (store $0 $1 ptr_sz)
+         (callv (^stable_cont_func $1 fetch)
```
jnthn: I guess this'd be a bit more efficient with a let that dereferenced the container_spec off the STable once?

src/jit/expr.c
```diff
@@ -0,0 +1,1006 @@
...
+                 * always is).  */
+                before_label = MVM_jit_label_before_ins(tc, jg, iter->bb, ins);
+                jg->handlers[ann->data.frame_handler_index].goto_label = before_label;
+                active_values_flush(tc, tree, values, sg->num_locals);
+                break;
+            case MVM_SPESH_ANN_DEOPT_OSR:
+                /* A label the OSR can jump into to 'start running', so to
+                 * speak. As it breaks the basic-block assumption, arguably,
+                 * this should only ever be at the start of a basic block. But
+                 * it's not. So we have to insert the label and compute it. */
```
jnthn: As of a while ago, this actually is the case: https://github.com/MoarVM/MoarVM/blob/master/src/spesh/graph.c#L329

docs/jit/ir.md
bdw: effective_spesh_slots still exists in `src/core/frame.h`, should it not?

src/core/bitmap.h
bdw: The idx indicates the bit we want, the first 6 bits represetn 0-63, so as we're dereferencing an array of 64 bit integers, we need to shift down by 6 bits to get the correct array index

src/jit/linear_scan.c
```diff
@@ -0,0 +1,1126 @@
...
+static void active_set_add(MVMThreadContext *tc, RegisterAllocator *alc, MVMint32 a) {
+    /* the original linear-scan heuristic for spilling is to take the last value
+     * in the set to expire, freeeing up the largest extent of code... that is a
+     * reasonably good heuristic, albeit not essential to the concept of linear
+     * scan. It makes sense to keep the stack ordered at all times (simplest by
+     * use of insertion sort). Although insertion sort is O(n^2), n is never
+     * large in this case (32 for RISC architectures, maybe, if we ever support
+     * them; 7 for x86-64. So the time spent on insertion sort is always small
+     * and bounded by a constant, hence O(1). Yes, algorithmics works this way
+     * :-) */
```
jnthn: :-P

src/jit/macro.expr
```diff
@@ -0,0 +1,62 @@
+# -*-whitespace-*-
+
+(macro: ^getf (,object ,type ,field)
+   (load (addr ,object (&offsetof ,type ,field)) (&SIZEOF_MEMBER ,type ,field)))
+(macro: ^setf (,object ,type ,field ,value)
+    (store (addr ,object (&offsetof ,type ,field)) ,value (&SIZEOF_MEMBER ,type ,field)))
+(macro: ^cu_string (,a) (idx  (^getf (cu) MVMCompUnit body.strings) ,a ptr_sz))
```
jnthn: Hm, but strings may be lazily loaded, or do we guard against that case somewhere else?

src/jit/macro.expr
```diff
@@ -0,0 +1,62 @@
...
+
+(macro: ^frame () (^getf (tc) MVMThreadContext cur_frame))
+(macro: ^parg (,a) (idx (^getf (^frame) MVMFrame params.args) ,a reg_sz))
+
+(macro: ^params () (addr (^frame) (&offsetof MVMFrame params)))
+(macro: ^caller () (addr (^frame) (&offsetof MVMFrame caller)))
+
+# get spesh slot address
+(macro: ^spesh_slot (,a)
+   (idx (^getf (^frame) MVMFrame effective_spesh_slots) ,a ptr_sz))
```
jnthn: Nothing to resolve right now, but I'm curious how we do the "if it's gen2, then we don't need to resolve it again because it won't move" trick with the expr JIT.

src/jit/x64/arch.c
```diff
@@ -0,0 +1,103 @@
+#include "moar.h"
+#include "jit/internal.h"
+
+#define MVM_JIT_PLATFORM_POSIX 1
+#define MVM_JIT_PLATFORM_WIN32 2
+
+#ifndef MIN
+#define MIN(a,b) ((a) < (b) ? (a) : (b))
+#endif
```
jnthn: I'm sure I saw this defined in `internal.h`

jnthn:
I've left a few small cleanup comments, but overall am very happy with this PR, and happy for it to be merged. Very nice work! :+1:

docs/jit/ir.md
jnthn: Oh, right, that one had to survive my MVMFrame shrinkage. :-)

AlexDaniel:
FWIW If it looks good, I'd prefer to see this merged this week. Any later than that raises the possibility of unnecessary headache with the upcoming release (this month we have rakudo * release too).

I don't know, maybe this won't be disruptive at all, but better safe than sorry.

docs/jit/ir.md
```diff
@@ -0,0 +1,257 @@
+# Expression 'Tree' Intermediate Representation
+
+The 'expression tree' IR has been developed developed to support
```
AlexDaniel: developed developed

jnthn:
@AlexDaniel Yes, getting this into the upcoming release in good time is why I spent a good chunk of my evening reviewing this. :-)

src/jit/core_templates.expr
bdw: Yes, but we'd have to guard against `$1` being non-null, and in general complicate the logic quite considerably... It'd be something like:

    (ifv (all (nz $1) (^is_concrete_obj $1)) 
          (let (($cont_spec (^getf (^stable $1) MVMSTable container_spec)))
                 (ifv (nz $const_spec)
                       (callv (^getf $const_spec MVMContainerSpec) 
                                 ...)
                       (store $0 $1 ptr_sz))
        (store $0 $1 ptr_sz))

And considering that due to tiling, the loads in `(^getf)` may be done doubly anyway (I'm planning an optimization to prevent that), I'm not sure we're actually doing better, and more importantly, the (future) optimizer should be able to resolve these things automatically.

But thanks for the comment, well found.

src/jit/expr.c
bdw: If that's the case then this can be cleaned up. Either way no label will actually be inserted since `MVM_jit_is_label_for_ins` will be false (it is a basic block label in that case) 😄

src/jit/macro.expr
bdw: Nope, good point, this is an old legacy and should be fixed

src/jit/macro.expr
bdw: In general, with a special node `objref` or `const_obj` or somesuch, that should be added by the expr tree builder. And then - to top it off - we can make that a GC root that is automatically updated.

src/jit/x64/arch.c
bdw: Correct, and then also undefined, because I didn't want them to leak out, i.e. to have nonsymbolic values... I'm not sure about the legitimacy of that concern

PR #676, opened 2017-09-06
JIT compile native calls

niner:
This is accompanied by jit_nativecall branches in nqp and rakudo.


src/spesh/facts.c
```diff
@@ -606,6 +606,11 @@ static void add_bb_facts(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpeshBB *bb,
             g->facts[result.reg.orig][result.reg.i].flags |= MVM_SPESH_FACT_DECONTED;
             break;
         }
+        case MVM_OP_nativecallinvokejit: {
+            for (i = 0; i < g->num_locals; i++)
+                g->facts[i][g->fact_counts[i] - 1].usages++;
```
jnthn: This looks weird and fragile. It seems to be giving every local a bump of 1...why? Usage counts are computed up at line 376 for every op.

src/6model/reprs/NativeCall.c
```diff
@@ -38,6 +38,7 @@ static void copy_to(MVMThreadContext *tc, MVMSTable *st, void *src, MVMObject *d
         memcpy(dest_body->arg_types, src_body->arg_types, src_body->num_args * sizeof(MVMint16));
     }
     dest_body->ret_type = src_body->ret_type;
+    dest_body->jitcode = src_body->jitcode;
```
jnthn: I think this file also wants a line for `jitcode` added to `gc_free`

src/spesh/facts.c
niner: The nativecallinvokejit OP is a bit special as it expects the arguments to be in the first locals of the call frame. Spesh didn't know this and removed them as dead weight. Upping the usage count is done to reflect the actual usage.

jnthn: Ouch, I think what I originally suggested (though it probably wasn't clear why, and so probably sounded odd...) was that `nativecallinvokejit` used the standard args buffer and then was just a kind of `invoke` instruction (so marked `-a` in `oplist` to terminate an args sequence). Then args would be placed into the args buffer like a normal call, and later we'd pull the values out of the `frame->args` buffer instead of `frame-work` (so I suspect it's a relatively small change on that side of things). This way, there'd be no issues with usage counts and so forth, since args in the arg buffer are obviously used by the callee.

Having spent so long of late removing cheats like this in spesh (many driven by the fact that each of them bit back somewhere and cost a bunch of debugging time), I'm quite reluctant to see another one like this go in. Especially as I suspect it's actually not sufficient in the case that a frame with `nativecallinvokejit` in it were to be inlined. If that happens, then we don't re-run the whole facts process on the inlinee, just do some much cheaper usage count setting, and so on that code path we wouldn't do extra usage bumps like is being done here, and then dead code elimination would eat them. We could of course do similar compensation as we do here, but that only makes me fear where we get caught out next...

src/core/oplist
```diff
@@ -830,6 +830,7 @@ atomicstore_o       r(obj) r(obj) :invokish
 atomicstore_i       r(obj) r(int64)
 barrierfull
 coveragecontrol     r(int64)
+nativecallinvokejit w(obj) r(obj) r(obj)
```
zhuomingliang: seems that some args is not used

src/jit/x64/emit.dasc
```diff
@@ -1917,9 +1917,38 @@ static void load_call_arg(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJit
         | mov TMP6, qword WORK[arg.v.reg];
         | lea TMP6, STOOGE:TMP6->data;
         break;
+    case MVM_JIT_REG_DYNIDX:
+        | get_cur_op TMP5;
+        | xor TMP6, TMP6;
+        | mov TMP6w, word [TMP5 + arg.v.reg*2]
```
bdw: Fwiw, the two lines above can, I think, become:
`| movzx TMP6, word [TMP5 + arg.v.reg*2];`
And, the nicer thing to do (I think) would probably be to use a two-word sized type:
`| movzs TMP6, word U16:TMP5[arg.v.reg*2];`

I *think* this is only necessary because you might call this nativecall object from separate positions in the bytecode (with the argument stored in a different location each time). In which case this makes sense, but it would probably be a good idea to add some documentation to explain why this can't be 'hardcoded'.

src/jit/x64/emit.dasc
```diff
@@ -1917,9 +1917,38 @@ static void load_call_arg(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJit
...
+        | mov TMP6, qword [WORK + TMP6*8];
+        break;
     case MVM_JIT_DATA_LABEL:
         | lea TMP6, [=>(arg.v.lit_i64)];
         break;
+    case MVM_JIT_SAVED_RV:
+        | mov TMP6, qword [rbp-0x28];
+        break;
```
bdw: There are other ways to do this; specifically, I've been using the args array as a temporary buffer in the compilation of unless_o and if_o. Not sure if that is possible in this case, but it would save you the implementation of SAVED_RV.
Alternatively, you can give  `[rbp-0x28]` a symbolic name with 
`|.define SAVED_RV, [rbp-0x28]`

src/jit/x64/emit.dasc
```diff
@@ -1917,9 +1917,38 @@ static void load_call_arg(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJit
...
+    case MVM_JIT_ARG_I64:
+        | mov TMP6, TC->cur_frame;
+        | mov TMP6, FRAME:TMP6->args;
+        | mov TMP6, qword REGISTER:TMP6[arg.v.lit_i64];
+        break;
```
bdw: The frame args and frame local variables are always contiguous (or used to be, maybe @jnthn knows), so *if* you know the size of the whole frame you can compute the offset directly from  args (and save yourself having to read FRAME->args, because you can just use rbx instead).

src/jit/x64/emit.dasc
```diff
@@ -1917,9 +1917,38 @@ static void load_call_arg(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJit
...
+    case MVM_JIT_ARG_PTR:
+        | mov TMP6, TC->cur_frame;
+        | mov TMP6, FRAME:TMP6->args;
+        | mov TMP6, qword REGISTER:TMP6[arg.v.lit_i64];
+        | mov TMP6, aword STOOGE:TMP6->data;
+        break;
```
bdw: Same as above, but more importantly, I think this should have a more descriptive name because it *dereferences* the data pointer of a MVMObjectStooge, which I wouldn't expect from the name.

src/jit/x64/emit.dasc
jnthn: Yes, it would be safe to do that.

src/core/nativecall.c
```diff
@@ -372,8 +372,162 @@ static const char *dlerror(void)
 }
 #endif
 
+void init_c_call_node(MVMThreadContext *tc, MVMSpeshGraph *sg, MVMJitNode *node, void *func_ptr, MVMint16 num_args, MVMJitCallArg *args) {
```
jnthn: Does this bunch of code need to be conditionally compiled depending on whether or not we build the JIT?

src/jit/x64/emit.dasc
niner: At the time when this code is compiled, the code calling ncinvoke_o does not even exist yet. So while args is always part of WORK, I don't know yet where exactly in WORK it will end up in. Hence the need to read FRAME->args.

Later when the frame is speshed and compiled again, this information is available and I emit MVM_JIT_PARAM_I64 instead of MVM_JIT_ARG_I64. The former can access WORK directly, as the offsets are known.

src/jit/x64/emit.dasc
niner: Naming is hard :) The enums are named after what is passed. Pointers are passed as CPointer objects which contain the actual pointer in their data. So it's indeed one level of dereferencing, but the result is still a pointer which is passed to the native function. So MVM_JIT_ARG_PTR does indeed result in a Pointer being passed. It just does the unboxing, too.
Maybe a comment like this ^^^ will suffice?

src/core/nativecall.c
niner: I can't find any conditional compilation of JIT code? Only the emit part but not the graph itself.

src/jit/x64/emit.dasc
niner: I don't see an advantage of using the args buffer over just putting it on the stack. I'd have to locate the args buffer first, at least for ahead of time compilation causing unnecessary memory churn. I have it the symbolic name. Works like a charm :)

niner:
I've addressed all the excellent suggestions (some by adding explanatory comments).

src/core/nativecall.c
bdw: What we do is 'stub' the JIT compiled code. The JIT graph declarations are always compiled, though, so this should be safe. We do not, in fact, have declared a `MVM_JIT_IS_COMPILED` parameter or something of that sort, so at worst, this is wasteful.

src/jit/x64/emit.dasc
bdw: Yep, I'll be perfectly happy with a comment.

PR #682, opened 2017-09-08
Bump libuv to ver. 1.14.1

dogbert17:
The version we're currently using, i.e. 1.8 is quite old.


jnthn:
I presume this has been tested and found to build NQP/Rakudo and pass test/stresstest?

dogbert17:
On my 32 bit Linux system I've tested the following

\# build MoarVM with the new libuv present in 3rdparty/libuv
make realclean
perl Configure.pl --debug --prefix=../../install/
make -j6 install

\# build rakudo
git pull
perl Configure.pl --gen-moar --gen-nqp --backends=moar
make
make test
make install

\# build nqp using the MoarVM built with libuv 1.14.1
perl Configure.pl --prefix=../install --backends=moar
make test

\# in the rakudo dir 
make stresstest TEST_JOBS=7

Everything works for me, whether that's enough is another question :)

dogbert17:
On my 32 bit Linux system I've tested the following. Note that i don't believe that this is the correct way to do this so it would be nice if someone else tested this as well.

\# build MoarVM with the new libuv present in 3rdparty/libuv
\# in order to do this I had to disable the syncing of submodules
\# which was done automatically on my system when running 'perl Configure.pl'
\# if not the version of libuv I had checked out would be silently replaced with
\# v 1.8 thus rendering all further testing useless
make realclean
perl Configure.pl --debug --prefix=../../install/
make -j6 install

\# build rakudo
git pull
perl Configure.pl --gen-moar --gen-nqp --backends=moar
make
make test
make install

\# build nqp using the MoarVM built with libuv 1.14.1
perl Configure.pl --prefix=../install --backends=moar
make test

\# in the rakudo dir 
make stresstest TEST_JOBS=7

Everything works for me, whether that's enough is another question :)

jnthn:
Let's merge this right after the next release; it's only 4 days away now, and that's a tad short for testing.

PR #687, opened 2017-09-11
Nativecall code : fix invalid pointer creation

Skarsnik:
This fix what is probably an extra indirection. If I understand this correctly cptr is the address of a field inside a CStruct, the previous code was assigning to cptr the value of the field.
This make efence/valgrind/asan happy (fix an invalid read size of 8 errors message  in this function)
The uintptr _t cast are to respect https://www.securecoding.cert.org/confluence/display/c/INT36-C.+Converting+a+pointer+to+integer+or+integer+to+pointer

All nativecall tests in rakudo still run fine with this fix

### DIFF
--- a/src/core/nativecall.c
+++ b/src/core/nativecall.c
@@ -752,13 +752,13 @@ void MVM_nativecall_refresh(MVMThreadContext *tc, MVMObject *cthingy) {
         for (i = 0; i < repr_data->num_attributes; i++) {
             MVMint32 kind = repr_data->attribute_locations[i] & MVM_CSTRUCT_ATTR_MASK;
             MVMint32 slot = repr_data->attribute_locations[i] >> MVM_CSTRUCT_ATTR_SHIFT;
-            void *cptr;   /* The pointer in the C storage. */
-            void *objptr; /* The pointer in the object representing the C object. */
+            void *cptr = NULL;   /* The pointer in the C storage. */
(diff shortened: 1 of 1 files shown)


src/core/nativecall.c
```diff
@@ -752,13 +752,13 @@ void MVM_nativecall_refresh(MVMThreadContext *tc, MVMObject *cthingy) {
         for (i = 0; i < repr_data->num_attributes; i++) {
             MVMint32 kind = repr_data->attribute_locations[i] & MVM_CSTRUCT_ATTR_MASK;
             MVMint32 slot = repr_data->attribute_locations[i] >> MVM_CSTRUCT_ATTR_SHIFT;
-            void *cptr;   /* The pointer in the C storage. */
-            void *objptr; /* The pointer in the object representing the C object. */
+            void *cptr = NULL;   /* The pointer in the C storage. */
+            void *objptr = NULL; /* The pointer in the object representing the C object. */
```
jnthn: These NULL initializations look like dead code (they are initialized on every code path).

jnthn:
Seems like a reasonable fix.

src/core/nativecall.c
Skarsnik: I know. It's just a habit I have with pointers. It's kinda safer if the code change bellow.

jnthn: Kinda, though one'd hope the compiler would catch that, and if it doesn't Valgrind or similar will.

PR #692, opened 2017-09-15
getsockname / end of non-void function

duke-m:
`getsockname` expects a socklen_t* as third argument (which is an unsigned int*) and not a signed int.
pragma to turn off the "end of non-void function" warning (hope you don't find pragmas ugly)

### DIFF
--- a/src/io/syncsocket.c
+++ b/src/io/syncsocket.c
@@ -15,7 +15,7 @@
 #define snprintf _snprintf
 #endif
 
-/* Assuemd maximum packet size. If ever changing this to something beyond a
+/* Assumed maximum packet size. If ever changing this to something beyond a
  * 16-bit number, then make sure to change the receive offsets in the data
  * structure below. */
 #define PACKET_SIZE 65535
--- a/src/io/syncsocket.c
+++ b/src/io/syncsocket.c
@@ -2,6 +2,7 @@
 
 #ifdef _WIN32
     #include <winsock2.h>
+    #include <ws2tcpip.h>
     typedef SOCKET Socket;
 #else
     #include "unistd.h"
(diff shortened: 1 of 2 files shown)


src/io/syncsocket.c
```diff
@@ -393,6 +394,10 @@ static const MVMIOOps op_table = {
     gc_free
 };
 
+/* since we'll throw an error in case of an error we can turn off -Wreturn-type here,
+works with GCC and clang, others will ignore this */
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wreturn-type"
```
jnthn: We already have a standard way to do this kind of thing, which is to annotate the `throw_error` function with `MVM_NO_RETURN`. I think that would solve it too, and we already know that works across plenty of different compilers.

duke-m:
thanks for the information, I've changed throw_error to have the no-return-attribute

jnthn:
Looks better. Just checked and I suspect that socklen_t may need one more include on Windows; see https://stackoverflow.com/questions/4453992/error-c2065-socklen-t-undeclared-identifier

duke-m:
I'll get back to this next Tuesday, thanks and have a nice weekend

duke-m:
Hi Jonathan, I did not see the appveyor build complain, what did I miss? thanks

src/io/syncsocket.h
```diff
@@ -1,3 +1,4 @@
+MVM_NO_RETURN static void throw_error(MVMThreadContext *tc, int r, char *operation) MVM_NO_RETURN_GCC;
```
zhuomingliang: It should be in `syncsocket.c` because it's not a global function.

jnthn:
This looks OK to me now; thanks.

PR #696, opened 2017-09-19
Probe for gcc -Werror=* support

### LINKED ISSUE #277: build broken on RHEL5 because of -Werror=declaration-after-statement
...
compiling src/main.o
cc1: error: unrecognized command line option "-Werror=declaration-after-statement"
make: **\* [src/main.o] Error 1
...

[jdv@jdv rakudo]$ cat /etc/redhat-release 
CentOS release 5.11 (Final)
[jdv@jdv rakudo]$

[jdv@jdv rakudo]$ cc -v
Using built-in specs.
Target: x86_64-redhat-linux
Configured with: ../configure --prefix=/usr --mandir=/usr/share/man
--infodir=/usr/share/info --enable-shared --enable-threads=posix
--enable-checking=release --with-system-zlib --enable-__cxa_atexit
--disable-libunwind-exceptions --enable-libgcj-multifile
--enable-languages=c,c++,objc,obj-c++,java,fortran,ada --enable-java-awt=gtk
--disable-dssi --disable-plugin
--with-java-home=/usr/lib/jvm/java-1.4.2-gcj-1.4.2.0/jre --with-cpu=generic
--host=x86_64-redhat-linux
Thread model: posix
gcc version 4.1.2 20080704 (Red Hat 4.1.2-55)
[jdv@jdv rakudo]$

...
15:11 < FROGGS> yes, we added these -Werror options just recently
15:11 < jdv79> that would be my guess
15:12 < FROGGS> it is meant to force devs to care about MSVC
15:12 < FROGGS> and actually not break things :/
15:14 < FROGGS> jdv79: we have a build/probe.pm that tries to figure out what works and what not... seems that this -Werror option needs to go in there
15:14 < FROGGS> :o)
15:14 < jdv79> ok
15:14 < jdv79> neato
15:14 < FROGGS> jdv79: so, if you don't want to tackle it yourself, please open an issue in the moarvm repo
15:14 < jdv79> looks like RHEL 5 is "not EOL'd til 2017 or 2020
15:14 < jdv79> so maybe worth supporting
15:15 < FROGGS> aye
15:15 < jdv79> i'll ticket it
...

patzim:
This fixes #277.
I had to shuffle the probes around a little to get them working.
Tested on fedora 25 and RHEL5. I do appreciate testing on other systems.

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -143,10 +143,24 @@ $config{translate_newline_output} = 0  unless defined $config{translate_newline_
 $config{ld}           = $config{cc} unless defined $config{ld};
 $config{ldout}        = $config{ccout} unless defined $config{ldout};
 $config{ldsys}        = $config{ldusr} unless defined $config{ldsys};
-$config{ldmiscflags}  = $config{ccmiscflags} unless defined $config{ldmiscflags};
 $config{ldoptiflags}  = $config{ccoptiflags} unless defined $config{ldoptiflags};
 $config{lddebugflags} = $config{ccdebugflags} unless defined $config{lddebugflags};
 $config{ldinstflags}  = $config{ccinstflags} unless defined $config{ldinstflags};
+
+# Probe the compiler.
+build::probe::compiler_usability(\%config, \%defaults);
+
+# Remove unsupported -Werror=* gcc flags if gcc doesn't support them.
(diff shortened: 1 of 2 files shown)


jnthn:
Set off another Travis run, since all the OSX runs didn't work out. So far as I understand the configure/build system, I think this looks OK.

PR #702, opened 2017-09-22
Fix hint

the-eater:


the-eater:
Yeah you're correct. Now I don't really know how to solve the BSD situation, I'm assuming Mac also uses 28, being a BSD-born system. will look into it! thanks for noticing!

PR #705, opened 2017-09-24
Optimize MVM_string_join in both case of all strands and case of a single element to join

MasterDuke17:
When everything to be joined are strands of the same type, just copy
them directly instead of creating a single flat string. For small
strings the overhead of being strands counteracts the saving in copying,
but for large strings it's much faster.

For example:
```
my $b = "bbbbbbbb" x 200;
my $c = "cccccccccccccc" x 300;
for ^100_000 {
    $ = join(",", $b, $c, "dddd")
}
```
took ~3.7s before this change and ~1.5s after.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -1630,6 +1630,20 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
     is_str_array = REPR(input)->pos_funcs.get_elem_storage_spec(tc,
         STABLE(input)).boxed_primitive == MVM_STORAGE_SPEC_BP_STR;
 
+    /* If there's only one element to join, just return it. */
+    if (elems == 1) {
+        if (is_str_array) {
+            MVMString *piece = MVM_repr_at_pos_s(tc, input, 0);
+            if (piece)
+                return piece;
+        }
+        else {
+            MVMObject *item = MVM_repr_at_pos_o(tc, input, 0);
+            if (item && IS_CONCRETE(item))
+                return MVM_repr_get_str(tc, item);
+        }
+    }
+
     /* Allocate result. */
     MVMROOT(tc, separator, {
     MVMROOT(tc, input, {
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -88,7 +88,7 @@ static char * NFG_checker (MVMThreadContext *tc, MVMString *orig, char *varname)
                 char *out = MVM_malloc(sizeof(char) * out_size);
                 char *waste[] = {orig_render, renorm_render, NULL};
                 char **w = waste;
-                snprintf(out, out_length,
+                snprintf(out, out_size,
                     format,
                     varname,
                         orig_graphs, renorm_graphs,
@@ -1614,7 +1614,7 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
     MVMString **pieces = NULL;
     MVMint64    elems, num_pieces, sgraphs, i, is_str_array, total_graphs;
     MVMuint16   sstrands, total_strands;
-    MVMint32    concats_stable = 1;
+    MVMint32    concats_stable = 1, all_strands;
     size_t      bytes;
 
     MVM_string_check_arg(tc, separator, "join separator");
@@ -1639,7 +1639,7 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
 
     /* Take a first pass through the string, counting up length and the total
      * number of strands we encounter as well as building a flat array of the
-     * strings (to we only have to do the indirect calls once). */
+     * strings (so we only have to do the indirect calls once). */
     sgraphs  = MVM_string_graphs_nocheck(tc, separator);
     if (sgraphs)
         sstrands = separator->body.storage_type == MVM_STRING_STRAND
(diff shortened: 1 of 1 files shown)


MasterDuke17:
40da3e6 removes the need for https://github.com/rakudo/rakudo/pull/1171 (and still passes all tests and spectests).

src/strings/ops.c
```diff
@@ -1630,16 +1631,35 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
...
+            if (piece)
+                return piece;
+        }
+        else {
+            MVMObject *item = MVM_repr_at_pos_o(tc, input, 0);
+            if (item && IS_CONCRETE(item))
+                return MVM_repr_get_str(tc, item);
+        }
+    }
+
```
jnthn: This part of the PR looks just fine.

src/strings/ops.c
```diff
@@ -1630,16 +1631,35 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
...
     MVMROOT(tc, separator, {
     MVMROOT(tc, input, {
         result = (MVMString *)MVM_repr_alloc_init(tc, tc->instance->VMString);
     });
     });
 
+    /* If the separator is a strand, find out its storage_type. */
+    all_strands_of_same_type = separator->body.storage_type == MVM_STRING_STRAND;
+    if (all_strands_of_same_type)
+        strand_type = separator->body.storage.strands->blob_string->body.storage_type;
```
jnthn: Why does it matter that all strands have the same storage type? A strand string can reference one 8-bit and another 32-bit wide strand just fine.

src/strings/ops.c
```diff
@@ -1693,13 +1718,27 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
...
+    if (total_strands <  MVM_STRING_MAX_STRANDS &&
+            16        <= total_graphs / total_strands &&
+            all_strands_of_same_type) {
+        MVMuint16 offset = 0;
+        result->body.storage_type    = MVM_STRING_STRAND;
+        result->body.storage.strands = allocate_strands(tc, total_strands);
+        result->body.num_strands     = total_strands;
+        for (i = 0; i < num_pieces; i++) {
+            MVMString *piece = pieces[i];
+            if (0 < i) {
```
jnthn: This is one of the places where I really don't find the "always use `<`" thing helps at all...

src/strings/ops.c
```diff
@@ -1693,13 +1718,27 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
...
+                copy_strands(tc, separator, 0, result, offset, separator->body.num_strands);
+                offset += separator->body.num_strands;
+            }
+            copy_strands(tc, piece, 0, result, offset, piece->body.num_strands);
+            offset += piece->body.num_strands;
```
jnthn: This code doesn't check if the result is NFG stable or not, which means it'll produce wrong results if two of the things being joined would form a single grapheme.

src/strings/ops.c
MasterDuke17: Personally I agree, but copied the code from what's now the "else" branch and wanted it to be consistent. Should I change it?

src/strings/ops.c
MasterDuke17: I wondered about this, but couldn't craft an example where the behavior after my change did anything different than the code before. I tried various combinations of joining elements that had strands of just combining characters, just combining characters as the separator, etc. Also, it does pass a spectest. Either the collapsing is correctly happening elsewhere, my ad-hoc test cases weren't actually testing the right things, or we should add some tests for this to roast. Do you have an example I can test?

src/strings/ops.c
MasterDuke17: Huh, I think I misinterpreted something @timo said (https://irclog.perlgeek.de/perl6-dev/2017-09-23#i_15206788) about doing that as an optimization. I thought he meant make support for mixed strings, but now I see he meant actually make strings with mixed strands, since that is actually supported. Cool, I'll add a commit removing that restriction.

MasterDuke17:
@jnthn removed the bit about all strands having to have the same storage_type and added in the concat stability checking code from the not-all-strands else branch.

With @samcv++'s help, added in some debugging code that did get tripped up with the initial changes, but now is clean.

src/strings/ops.c
```diff
@@ -1693,13 +1714,51 @@ MVMString * MVM_string_join(MVMThreadContext *tc, MVMString *separator, MVMObjec
     }
     result->body.num_graphs = total_graphs;
 
-    /* If we just collect all the things as strands, are we within bounds, and
-     * will be come out ahead? */
-    if (total_strands < MVM_STRING_MAX_STRANDS && 16 <= total_graphs / total_strands) {
-        /* XXX TODO: Implement this, conditionalize branch thing below. */
+    /* If the separator and pieces are all strands, and there are
+     * enough strands of a large enough size, just copy them directly. */
```
samcv: Can you change it to this please:
```
/* If the separator and pieces are all strands, and
 * there are on average at least 16 graphemes in each of the strands */
if (all_strands && total_strands < MVM_STRIND_MAX_STRANDS
           && total_strands * 16 <= total_graphs)
```
The conditional becomes much clearer and clearer comment.

MasterDuke17:
@samcv and done.

samcv:
Thank you :-)

PR #709, opened 2017-09-29
Used FSA allocator for jump table in KMP index algorithm

samcv:
We can save cpu cycles by using FSA, since we just end up freeing it as
soon as the function returns.
Also, since the loop is duplicated, turn it into a #define so we can
use the same code with different functions for accessing the Haystack.


samcv:
Commited this directly

PR #710, opened 2017-09-30
Change realloc condition in MVM_string_utf8_decode

MasterDuke17:
The conditional being `>=` instead of `>` meant that the buffer was
frequently being realloced to the same size.

Callgrind reports ~13k fewer instructions and heaptrack reports ~90
fewer realloc calls for `perl6 -e ''`.

NQP builds ok and passes `make m-test` and Rakudo builds ok and
passes `make m-test m-spectest`.


samcv:
Looks good.

PR #716, opened 2017-10-04
join: simplify some conditionals and factor out some code

samcv:
Reduce the complexity of several conditionals, from multiple to
just one. Reduce the depth of one set of conditionals.

Factor out the by grapheme copying code to make the somewhat long
function shorter and easier to navigate.


MasterDuke17:
I'm definitely on board with pulling out the copy_to_32bit function. However, (and this may just be personal preference) I don't find the re-worked conditionals easier to read/understand.

samcv:
@MasterDuke17 I could change the conditionals to use `if (!sgraphs)` instead of including with the others. So instead of reducing 3 to 1 conditional. It goes from 3 to 2. 1 for the !concats_stable and 1 for the concat checks

PR #717, opened 2017-10-05
Disable tests on Appveyor until we can fix the build

samcv:
If it stalls during testing, we get no feedback on when the build itself
has broken.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -31,6 +31,6 @@ branches:
     - master
 
 
-test_script:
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Hold on, but https://github.com/MoarVM/MoarVM/issues/718 ? It does not seem to stall there.

AlexDaniel:
> samcv (IRC): because of that we didn't notice the build actually breaking with the recent JIT merge
> AlexDaniel: samcv: oooh, I understand your point now

So the status will be showing if it's possible to compile Moar on not. OK then.

PR #731, opened 2017-10-24
Fix seekability detection on MacOS

zoffixznet:
This fixes [the bug](https://rt.perl.org/Ticket/Display.html?id=132349) it was meant to fix, but needs a spectest/stresstest on MacOS done and seeing if it introduces any significant performance degradation on MacOS. Also, I'm out of my depth here and poorly know what I'm doing.

-----

POSIX does not define which devices must support lseek() and on
MacOS it returns some non-zero value that isn't -1 indicating failure
when lseek()ing TTY STDIN. This results in such handles marked
as seekable and in turn causes issues, such as mvm_eof() trying
to figure out whether we've reached EOF and falsely returning True
even when we didn't because the statbuf.st_size is zero.

After discussion[^1] and some research, the solution in this patch
is what I came up with and that actually works. We try to seek
a handle with lseek and on MacOS, if it doesn't error out, then
try to see if the file is a regular file, which is guaranteed to
be seekable.

Does this work? It passes the rakudo test[^1] to cover the bug,
but so far I'm having trouble completing spectest on the macos VM
I got (maybe something with VM I dunno).

Is this sane? I've no idea. I'm a n00b at this.

[1] (Notes section) http://man7.org/linux/man-pages/man2/lseek.2.html
[2] https://irclog.perlgeek.de/perl6-dev/2017-10-23#i_15341511
[3] https://github.com/perl6/roast/commit/27833272b4

### DIFF
--- a/build/Makefile.in
+++ b/build/Makefile.in
@@ -55,7 +55,8 @@ PLATFORM_WIN32 = src/platform/win32/mmap@obj@ \
                  src/platform/win32/io@obj@
 
 PLATFORM_POSIX = src/platform/posix/mmap@obj@ \
-                 src/platform/posix/time@obj@
+                 src/platform/posix/time@obj@ \
+                 src/platform/posix/io@obj@
 
 DASM_FLAGS   = @dasm_flags@
 JIT_ARCH_X64 = src/jit/x64/emit@obj@ src/jit/x64/arch@obj@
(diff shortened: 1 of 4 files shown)


geekosaur:
fwiw a more robust check is probably to `isatty()` it on open and mark it non-seekable if true.

zoffixznet:
> fwiw a more robust check is probably to isatty() it on open and mark it non-seekable if true.

Thanks. Made that change now. If I remove the MacOS detection and stresstest on Linux with the patch, I see no noticable perf impact. Don't have access to a MacOS ATM so if someone could test this out that'd be great (just see if `perl6 -e '$*IN.getc.say'` waits to read a char and prints it instead of printing `Nil` right away)

PR #733, opened 2017-10-25
Handle unsigned types to switch in MVM_iter()

samcv:
Fixes "Unknown lexical type encountered while building context iterator"
error when MVM_iter() tries to handle unsigned integers.


niner:
The only nit I have to pick with this is the commit message. It tells me the same as the patch (some cases added to a switch statement) but not [b]why[/b] we'd want that.

samcv:
@niner updated to include the reasoning. Thanks.

PR #734, opened 2017-10-26
If dividing by 1, just return the numerator

MasterDuke17:
This avoids allocating a new mp_int and actually performing the
division.

NQP builds ok and passes `make m-test`. Rakudo builds ok and passes `make m-test`. However, there are two spectest failures, one in S02-types/set.t and one in S32-num/int.t. These are both because `Int.new(int   \value)` and `Int.new(Int:D \value = 0)` are explicitly using `nqp::div_I` to create new Ints.

Some benchmarking results for this code: `perl6 -e 'use nqp; my $s; for ^1_000_000 { $s = nqp::div_I($_, 1, Int) }; say now - INIT now; say $s'`

Callgrind reports ~2m more instructions for dividing by 17, but 300m fewer instructions for dividing by 1.


zoffixznet:
This would break the code like this, won't it:

```
m: use nqp; class Meow is Int {}; dd nqp::div_I(<42>, 1, Meow).^name
rakudo-moar 9554a97c5: OUTPUT: «"Meow"␤»
```

In other words, the optimized path doesn't bother setting the correct requested type. 

Also, any idea what to do in Int.new to get the wanted behaviour back?

jnthn:
We should only do the optimization in this operation if the required type and the numerator type match exactly. Rakudo shouldn't have to change.

MasterDuke17:
Something like adding ' && REPR(a) == REPR(b)' to the conditional?

zoffixznet:
> Rakudo shouldn't have to change.

It'll likely need to, mostly because I'm [abusing `div_I` op in Int.new](https://github.com/rakudo/rakudo/commit/e4a5bb17c9b16a6e2b2e7ec973c09bc8918a41e0) to get a fresh object. I wasn't sure what's a proper to do that was… Basically fixes stuff like this:

```
m: my $orig = 2; my $new = Int.new: my int $ = 2; $new does role Meows {}; say $orig.^name
rakudo-moar 10e7af009: OUTPUT: «Int␤»
```

Where `Int.new` is a way to get a fresh object you could mess with without affecting all the other cached constants.

jnthn:
@MasterDuke17 STABLE, not REPR

MasterDuke17:
If https://github.com/MoarVM/MoarVM/pull/737, https://github.com/perl6/nqp/pull/378, and https://github.com/rakudo/rakudo/pull/1217 are merged, an NQP and Rakudo built with this PR pass all tests.

jnthn:
I'm happy for this to go in, though it'll need deconflicting first.

MasterDuke17:
Don't know what happened there, but resolved now.

PR #735, opened 2017-10-26
Don't throw exception for EINTR when writing to fd

ugexe:
Fixes `Failed to write bytes to filehandle: Interrupted system call` when running:
`strace perl6 -e 'while 1 { use nqp; my $s := $*SCHEDULER; my $c := nqp::list("echo", "-n", q||); my $e := CLONE-HASH-DECONTAINERIZED %*ENV; await ^500 .map: { start nqp::spawnprocasync($s.queue, $c, "/", $e, nqp::hash()) }; print "." }'`


ugexe:
Note that I have been unable to reproduce without invoking `strace`

jnthn:
Good catch, though I suspect this is not the only place we need such a patch. Will create a ticket to audit the code for other places.

PR #737, opened 2017-10-27
Create and JIT coerce_II op

MasterDuke17:
Implement it with the new `MVM_bigint_from_bigint` function. This will be
used in Rakudo's `Int.new` to create one Int from another, but without any
mix-ins the source might have.


src/math/bigintops.c
```diff
@@ -829,6 +829,26 @@ void MVM_bigint_from_str(MVMThreadContext *tc, MVMObject *a, const char *buf) {
     }
 }
 
+MVMObject * MVM_bigint_from_bigint(MVMThreadContext *tc, MVMObject *result_type, MVMObject *a) {
+    MVMObject * const result = MVM_repr_alloc_init(tc, result_type);
```
jnthn: Need to `MVMROOT` `a` while doing this allocation.

MasterDuke17:
Added the MVMROOT.

src/math/bigintops.c
```diff
@@ -817,6 +817,32 @@ void MVM_bigint_from_str(MVMThreadContext *tc, MVMObject *a, const char *buf) {
     }
 }
 
+MVMObject * MVM_bigint_from_bigint(MVMThreadContext *tc, MVMObject *result_type, MVMObject *a) {
+    MVMP6bigintBody *a_body  = get_bigint_body(tc, a);
```
jnthn: This returns an interior pointer into `a`, so this assignment should be moved until after the `MVMROOT` also.

MasterDuke17: Fixed

Issue #736, opened 2017-10-26
Audit for missing EINTR handling


jnthn:
In #735 one case was patched, but there are surely some others, especially in the sync I/O code (files and sockets). So, let's audit the calls we make and, for those that can return EINTR, make sure we're doing the retry.

PR #738, opened 2017-10-27
Retry in certain cases of EINTR

ugexe:
These are based on test cases from libuv, although there is no
evidence this fixes anything.

See: #736

### DIFF
--- a/src/io/syncfile.c
+++ b/src/io/syncfile.c
@@ -70,15 +70,15 @@ static MVMint64 mvm_fileno(MVMThreadContext *tc, MVMOSHandle *h) {
  * buffering output. */
 static void perform_write(MVMThreadContext *tc, MVMIOFileData *data, char *buf, MVMint64 bytes) {
     MVMint64 bytes_written = 0;
+    MVM_gc_mark_thread_blocked(tc);
     while (bytes > 0) {
         int r;
         do {
-            MVM_gc_mark_thread_blocked(tc);
             r = write(data->fd, buf, (int)bytes);
-            MVM_gc_mark_thread_unblocked(tc);
         } while (r == -1 && errno == EINTR);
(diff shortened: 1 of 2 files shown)


src/io/syncfile.c
```diff
@@ -70,26 +70,22 @@ static MVMint64 mvm_fileno(MVMThreadContext *tc, MVMOSHandle *h) {
  * buffering output. */
 static void perform_write(MVMThreadContext *tc, MVMIOFileData *data, char *buf, MVMint64 bytes) {
     MVMint64 bytes_written = 0;
-    MVM_gc_mark_thread_blocked(tc);
     while (bytes > 0) {
         int r;
-
         do {
+            MVM_gc_mark_thread_blocked(tc);
```
zhuomingliang: Move `MVM_gc_mark_thread_blocked(tc)` into `while (bytes > 0) { ` block may cause a lot of needless marking blocked and marking unblocked

src/io/syncsocket.c
```diff
@@ -198,19 +200,21 @@ MVMint64 socket_write_bytes(MVMThreadContext *tc, MVMOSHandle *h, char *buf, MVM
     unsigned int interval_id;
 
     interval_id = MVM_telemetry_interval_start(tc, "syncsocket.write_bytes");
-    MVM_gc_mark_thread_blocked(tc);
     while (bytes > 0) {
-        int r = send(data->handle, buf, (int)bytes, 0);
-        if (MVM_IS_SOCKET_ERROR(r)) {
+        int r;
+        do {
+            MVM_gc_mark_thread_blocked(tc);
```
zhuomingliang: And same here.

jnthn:
Looks sensible to me. :+1:

PR #739, opened 2017-10-28
Add two more missing MVMROOTs around managed mutex acquire

dogbert17:
The mutex acquire may GC block and thus trigger GC, moving objects.


lizmat:
Leaving this to jnthn to vet

PR #740, opened 2017-10-28
Add another MVMROOT around managed mutex acquire. jnthn++

dogbert17:


lizmat:
The indenting appears to be a little weird ?

jnthn:
@lizmat We don't tend to add an extra level of indentation for every `MVMROOT` when there's a bunch in a row; in some places there's 4 and it'd get crazy. :-)

PR #744, opened 2017-11-02
subtract size of profiling ops from bytecode size

timo:
so that turning profiling on can't push some functions
over the inline size limit, which causes differences
between profiled and non-profiled runs.

Not inlining some candidate only during profiling can
have ripple effects if an inlined piece of code causes
a bigger frame to bail in the JIT.


niner:
I'm not a huge fan of assigning meaning to the opcode number, when opcodes can already have adverbs. Could you at least put the ins->info->opcode >= MVM_OP_prof_enter test into a macro, so a casual reader doesn't have to know about the convention and to make it easier to change it later?

PR #750, opened 2017-11-10
Optimize MVM_string_concat when the RHS is a...

MasterDuke17:
single strand that's equal to the LHS's final strand. In that case, just
increase the LHS's repetition count by the number of repetition of the
RHS.

This adapted code from https://rt.perl.org/Ticket/Display.html?id=131774:
`$ /usr/bin/time perl6-m -e 'for (1, 10, 100, 1_000, 10_000, 100_000) -> $limit {my $x; my $y = "x" x 100; $x ~= $y for (1..$limit); say "$limit: ", now - ENTER now}'` used to report:
```
1: 0.0045150 
10: 0.0004916 
100: 0.00096704 
1000: 0.0077635 
10000: 0.4149077 
100000: 40.1284791 
37.36user 3.66system 0:41.02elapsed 100%CPU (0avgtext+0avgdata 3776812maxresident)k
```
and now reports:
```
1: 0.0043542
10: 0.0005357
100: 0.0008045
1000: 0.00442383
10000: 0.0260266
100000: 0.1531845
0.58user 0.01system 0:00.47elapsed 125%CPU (0avgtext+0avgdata 91024maxresident)k
```

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -612,14 +612,16 @@ static MVMString * string_from_strand_at_index(MVMThreadContext *tc, MVMString *
     return MVM_string_substring(tc, ss->blob_string, ss->start, ss->end - ss->start);
 }
 
-/* Append one string to another. */
 static MVMuint16 final_strand_match_with_repetition_count(MVMThreadContext *tc, MVMString *a, MVMString *b) {
     if (a->body.storage_type == MVM_STRING_STRAND) {
         MVMStringStrand *sa = &(a->body.storage.strands[a->body.num_strands - 1]);
-        if (sa->end - sa->start == MVM_string_graphs(tc, b)) {
+        /* If the final strand of a eq b, we'll just increment the final strand of a's repetitions. */
+        if (sa->end - sa->start == MVM_string_graphs_nocheck(tc, b)) {
             if (MVM_string_equal_at(tc, sa->blob_string, b, sa->start))
                 return 1;
         }
+        /* If the final strand of a eq the first (and only) strand of b, we'll just add b's repetitions
+	 * (plus 1 for the strand itself) to the final strand of a's repetitions. */
         else if (b->body.storage_type == MVM_STRING_STRAND && b->body.num_strands == 1) {
             MVMStringStrand *sb = &(b->body.storage.strands[0]);
             if (sa->end - sa->start == sb->end - sb->start)
@@ -629,6 +631,8 @@ static MVMuint16 final_strand_match_with_repetition_count(MVMThreadContext *tc, 
     }
     return 0;
 }
+
+/* Append one string to another. */
 MVMString * MVM_string_concatenate(MVMThreadContext *tc, MVMString *a, MVMString *b) {
     MVMString *result = NULL, *renormalized_section = NULL;
     int renormalized_section_graphs = 0, consumed_a = 0, consumed_b = 0;
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -607,13 +607,23 @@ MVMString * MVM_string_replace(MVMThreadContext *tc, MVMString *original, MVMint
     return result;
 }
 
+static MVMString * string_from_strand_at_index(MVMThreadContext *tc, MVMString *a, MVMuint16 index) {
+    MVMStringStrand *ss = &(a->body.storage.strands[index]);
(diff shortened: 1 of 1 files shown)


lizmat:
This is a great fix for this special case.  But I can’t help but wonder what costs are involved for the other, more often ocurring cases of string concatenation.  Could you provide some benchmarks for that as well?

MasterDuke17:
I ran several different cases and didn't notice any slowdown because of this fix.
```
for (1, 10, 100, 1_000, 10_000, 100_000) -> $limit {

# these are the different bodies I tried for this loop
    my $x = "yyy"; my $y = "x" x 100; $x ~= $y for (1..$limit); 
    my $x = "yyy"; my $y = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"; $x ~= $y for (1..$limit);
    my $x = "y" x 5; my $y = "x" x 100; $x ~= $y for (1..$limit);
    my $x = "y" x 5; my $y = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"; $x ~= $y for (1..$limit);
    my $x; $x ~= ("a".."z").pick x 100 for (1..$limit);

    say "$limit: ", now - ENTER now
}
```
This should be the worst case, as it passes the first part of the conditional I added, but fails the second more expensive part.
```
my $a = "aaaaaaaaaa" x 100; my $b = "aaaaaaaaaab" x 100; my $c; $c ~= $a ~ $b for ^10_000; say now - INIT now
```
I didn't see a noticeable slowdown in any of the above cases. Anything I might have missed?

lizmat:
If you didn't notice a slowdown, then this is fine by me :-)

src/strings/ops.c
```diff
@@ -607,13 +607,25 @@ MVMString * MVM_string_replace(MVMThreadContext *tc, MVMString *original, MVMint
...
+
 /* Append one string to another. */
-static MVMint32 final_strand_matches(MVMThreadContext *tc, MVMString *a, MVMString *b) {
+static MVMuint16 final_strand_match_with_repetition_count(MVMThreadContext *tc, MVMString *a, MVMString *b) {
     if (a->body.storage_type == MVM_STRING_STRAND) {
-        MVMStringStrand *ss = &(a->body.storage.strands[a->body.num_strands - 1]);
-        if (ss->end - ss->start == MVM_string_graphs(tc, b))
-            if (MVM_string_equal_at(tc, ss->blob_string, b, ss->start))
+        MVMStringStrand *sa = &(a->body.storage.strands[a->body.num_strands - 1]);
+        if (sa->end - sa->start == MVM_string_graphs(tc, b)) {
```
samcv: Change this to MVM_string_graphs_nocheck

samcv: Add a comment above this conditional to describe what it is checking

src/strings/ops.c
```diff
@@ -607,13 +607,25 @@ MVMString * MVM_string_replace(MVMThreadContext *tc, MVMString *original, MVMint
...
+            if (MVM_string_equal_at(tc, sa->blob_string, b, sa->start))
                 return 1;
+        }
+        else if (b->body.storage_type == MVM_STRING_STRAND && b->body.num_strands == 1) {
```
samcv: And a comment above this conditional as well to describe what it is checking in the 2nd part

src/strings/ops.c
```diff
@@ -607,13 +607,25 @@ MVMString * MVM_string_replace(MVMThreadContext *tc, MVMString *original, MVMint
     return result;
 }
 
+static MVMString * string_from_strand_at_index(MVMThreadContext *tc, MVMString *a, MVMuint16 index) {
+    MVMStringStrand *ss = &(a->body.storage.strands[index]);
+    return MVM_string_substring(tc, ss->blob_string, ss->start, ss->end - ss->start);
+}
+
 /* Append one string to another. */
```
samcv: This comment is in the wrong location

MasterDuke17:
Updated.

samcv:
Looks good now. Thanks!

PR #719, opened 2017-10-08
Use uv_fs_copyfile API

ugexe:
`uv_fs_copyfile` uses `CopyFileW` on windows and `copyfile(3)` on OSX (when available), while otherwise falling back to using  a loop + `uv_fs_sendfile` (similar to the code this replaces).

PR #752, opened 2017-11-13
Revert "Use uv_fs_copyfile API"

ugexe:
Reverts MoarVM/MoarVM#719

See: https://github.com/libuv/libuv/issues/1625


ugexe:
May be worth waiting a day or two before reverting, but with an imminent release we should be ready.

PR #759, opened 2017-11-30
Update Unicode version

briandfoy:
The moarvm.org page touts Unicode 10.


samcv:
I changed the commit title and description to be a bit more clear. Merging now. Thanks

PR #767, opened 2017-12-24
Fix unwanted null strings in VMArray holes

zoffixznet:
Whenever we read past end or zero out str VMArrays, stick empty strings instead of a null pointer there.

**Before:**
```
$ perl6 -e 'my str @a; @a[3] = "x"; dd @a'
chars requires a concrete string, but got null
  in block <unit> at -e line 1
```

**After:**
```
$ perl6 -e 'my str @a; @a[3] = "x"; dd @a'
array[str].new("", "", "", "x")
```

This fixes the issue, but I just did brute-force-fixage and...

no-idea

### DIFF
--- a/src/6model/reprs/VMArray.c
+++ b/src/6model/reprs/VMArray.c
@@ -153,7 +153,7 @@ static void at_pos(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *d
             if (kind != MVM_reg_str)
                 MVM_exception_throw_adhoc(tc, "MVMArray: atpos expected string register");
             if (index >= body->elems)
-                value->s = NULL;
+                value->s = tc->instance->str_consts.empty;
             else
                 value->s = body->slots.s[body->start + index];
             break;
@@ -251,7 +251,7 @@ static MVMuint64 zero_slots(MVMThreadContext *tc, MVMArrayBody *body,
             break;
         case MVM_ARRAY_STR:
             while (elems < ssize)
-                body->slots.s[elems++] = NULL;
+                body->slots.s[elems++] = tc->instance->str_consts.empty;
             break;
         case MVM_ARRAY_I64:
             while (elems < ssize)


lizmat:
I think this is a problem of native array .perl, rather than this should be fixed at the VM level.  Also, wouldn’t this mess with :exists ???

zoffixznet:
> I think this is a problem of native array .perl

It's a problem with anything that touches it:

```
<Zoffix_> m: my str @a; @a[3] = "x"; put @a[1]
<camelia> rakudo-moar 5b147fc81: OUTPUT: «concatenate requires a concrete string, but got null␤  in block <unit> at <tmp> line 1␤␤»
<Zoffix_> m: my str @a; @a[3] = "x"; say @a[1]
<camelia> rakudo-moar 5b147fc81: OUTPUT: «concatenate requires a concrete string, but got null␤  in block <unit> at <tmp> line 1␤␤»
<Zoffix_> m: my str @a; @a[3] = "x"; say @a[1] ~ 42
<camelia> rakudo-moar 5b147fc81: OUTPUT: «concatenate requires a concrete string, but got null␤  in block <unit> at <tmp> line 1␤␤»
```

> rather than this should be fixed at the VM level

Hence the dog picture :) I notice `nqp::null` gets HLLified to a Mu, but `nqp::null_s` doesn't. Is that how to fix the problem maybe?

> Also, wouldn’t this mess with :exists ???

That thought occurred to me, but then I noticed that issue... exists... even without my PR, so at the time I assumed since you can't have `str` type objects that was why. But perhaps this is a bug?

```
<Zoffix_> m: my str @a; @a[3] = "x"; say @a[1]:exists
<camelia> rakudo-moar 5b147fc81: OUTPUT: «True␤»
```

lizmat:
> On 25 Dec 2017, at 21:29, Zoffix Znet <notifications@github.com> wrote:
> 
> I think this is a problem of native array .perl
> 
> It's a problem with anything that touches it:
> 
> <Zoffix_> m: my str @a; @a[3] = "x"; put @a[1]
> <camelia> rakudo-moar 5b147fc81: OUTPUT: «concatenate requires a concrete string, but got null␤  in block <unit> at <tmp> line 1␤␤»
> <Zoffix_> m: my str @a; @a[3] = "x"; say @a[1]
> <camelia> rakudo-moar 5b147fc81: OUTPUT: «concatenate requires a concrete string, but got null␤  in block <unit> at <tmp> line 1␤␤»
> <Zoffix_> m: my str @a; @a[3] = "x"; say @a[1] ~ 42
> <camelia> rakudo-moar 5b147fc81: OUTPUT: «concatenate requires a concrete string, but got null␤  in block <unit> at <tmp> line 1␤␤»
> 
> rather than this should be fixed at the VM level
> 
> Hence the dog picture :) I notice nqp::null gets HLLified to a Mu, but nqp::null_s doesn't. Is that how to fix the problem maybe?

That feels better.  Perhaps that should HLLize to Str ?


> Also, wouldn’t this mess with :exists ???
> 
> That thought occurred to me, but then I noticed that issue... exists... even without my PR, so at the time I assumed since you can't have str type objects that was why. But perhaps this is a bug?
> 
> <Zoffix_> m: my str @a; @a[3] = "x"; say @a[1]:exists
> <camelia> rakudo-moar 5b147fc81: OUTPUT: «True␤»

After looking at it earlier today, I think the rule is that anything below elems exists, and anything beyond that doesn’t

$ 6 'my int @a; @a[2] = 42; dd @a.EXISTS-POS(1); dd @a.EXISTS-POS(3)'
Bool::True
Bool::False

zoffixznet:
> That feels better.  Perhaps that should HLLize to Str ?

Wouldn't that be weird that we'd have a native type that HLLizes to Str, yet you can't put Str into it?

For the record, Scalar native types seem to behave like what this PR makes them as in arrays; empty strings:
```
<Zoffix_> m: my str $x; dd $x
<camelia> rakudo-moar 8aa17f273: OUTPUT: «""␤»
```

zoffixznet:
ping for review

Issue #769, opened 2017-12-27
[Windows] Memory allocation failures say "could not allocate zu bytes"


lefth:
I compiled rakudo on Windows 10 with the Strawberry Perl toolchain (5.24.1001), which includes gcc 4.9.2.

The error is: 'MoarVM panic: Memory allocation failed; could not allocate zu bytes'

See: src/core/exceptions.c

geekosaur pointed out that `%zu` is the format string for `size_t`. This doesn't work on Windows in gcc. (It works in g++, but that isn't useful.) `%Iu` works, so it should be easy to tweak src/platform/inttypes.h.

This is Rakudo version 2017.12 built on MoarVM version 2017.12

PR #770, opened 2017-12-27
On Windows (not using Visual Studio), print size_t with %Iu

lefth:
Don't use the %zu format string because it's not handled, so a literal
"zu" is printed.

See issue #769.


zoffixznet:
You mention `%zu` doesn't work on gcc, so presumably the fix is just for that, but the conditional being added in the PR also includes MVC version `>=1800`.  Does `Iu` with a different header work there or does the condition need to be changed?

lefth:
@zoffixznet I fixed the condition. Thanks for catching that.

zoffixznet:
👍 Thanks!

PR #773, opened 2018-01-01
Use FSA for string storage

MasterDuke17:
This covers a lot of `MVMString` usage, but not everything. E.g., it might be possible to do the same thing in `src/strings/normalize.c`.

### DIFF
--- a/src/core/fixedsizealloc.c
+++ b/src/core/fixedsizealloc.c
@@ -280,7 +280,12 @@ static void add_to_bin_freelist(MVMThreadContext *tc, MVMFixedSizeAlloc *al,
 }
 void MVM_fixed_size_free(MVMThreadContext *tc, MVMFixedSizeAlloc *al, size_t bytes, void *to_free) {
 #if FSA_SIZE_DEBUG
-    MVMFixedSizeAllocDebug *dbg = (MVMFixedSizeAllocDebug *)((char *)to_free - 8);
+    MVMFixedSizeAllocDebug *dbg;
+    if (!to_free) {
+        MVM_free(to_free);
+        return;
+    }
+    dbg = (MVMFixedSizeAllocDebug *)((char *)to_free - 8);
     if (dbg->alloc_size != bytes) {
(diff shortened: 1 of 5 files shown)


src/core/fixedsizealloc.c
```diff
@@ -281,7 +281,12 @@ static void add_to_bin_freelist(MVMThreadContext *tc, MVMFixedSizeAlloc *al,
 }
 void MVM_fixed_size_free(MVMThreadContext *tc, MVMFixedSizeAlloc *al, size_t bytes, void *to_free) {
 #if FSA_SIZE_DEBUG
-    MVMFixedSizeAllocDebug *dbg = (MVMFixedSizeAllocDebug *)((char *)to_free - 8);
+    MVMFixedSizeAllocDebug *dbg;
+    if (!to_free) {
```
vendethiel: If to_free is NULL, free NULL?

MasterDuke17: Hm. I've been working on this on-and-off for a couple months now, so I don't remember the exact thinking behind this. However, I believe the reason for the check is that valgrind complained about the `(char *)to_free - 8` if to_free was NULL. I also think there was some chatting in #moarvm about whether freeing NULL was ok (with the conclusion being yes).

zoffixznet: But you already know `to_free` is NULL in that branch before you call `free`

MasterDuke17: I looked back at the [#moarvm logs](https://irclog.perlgeek.de/moarvm/2017-11-23#i_15486027) from around when the [commit](https://github.com/MoarVM/MoarVM/commit/dcfcb41f2397f164e27fd8af2adedeb1fc14a57e) with this change went in (in my non-squashed branch). I was running into a problem with 0 length fixed_size_allocs, so maybe I thought the MVM_free was needed for bookkeeping? Anyway, I just tried removing it (so only the `return` is in the `if` body) and NQP built ok and passed `make m-test` and Rakudo built ok and passed `make m-test m-spectest`.

PR #780, opened 2018-01-12
Try to use egcc when compiling on OpenBSD

Kaiepi:
egcc's a package providing an up-to-date version of gcc in OpenBSD,
which still uses v4.2 at the moment

### DIFF
--- a/build/setup.pm
+++ b/build/setup.pm
@@ -281,6 +281,28 @@ our %COMPILERS = (
         formatattribute   => '__attribute__((format(X, Y, Z)))',
     },
 
+    egcc => {
+        -toolchain => 'gnu',
+
+        cc => 'egcc',
+        ld => undef,
+
+        ccmiscflags  => '-Werror=declaration-after-statement -Werror=pointer-arith',
+        ccwarnflags  => '',
+        ccoptiflags  => '-O%s -DNDEBUG',
+        ccdebugflags => '-g%s',
+        ccinstflags  => '-pg',
+
+        ldmiscflags  => '',
+        ldoptiflags  => undef,
+        lddebugflags => undef,
+        ldinstflags  => undef,
+
+        noreturnspecifier => '',
+        noreturnattribute => '__attribute__((noreturn))',
+        formatattribute   => '__attribute__((format(X, Y, Z)))',
+    },
+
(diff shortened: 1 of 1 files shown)


build/setup.pm
```diff
@@ -437,6 +459,7 @@ our %OS_LINUX = (
 our %OS_OPENBSD = (
     %OS_POSIX,
 
+    cc => (qx!egcc -v 2>&1 >$devnull! !~ 'egcc') ? 'gcc' : 'egcc',
     syslibs => [ @{$OS_POSIX{syslibs}}, qw( kvm ) ],
```
zoffixznet: This would get executed everywhere, even on non-BSDs, won't it? For example, here's what that `qx` gives on Windows:

```
C:\rakudo>perl -e "qx!egcc -v 2>&1 >$devnull! !~ 'egcc'"
The syntax of the command is incorrect.
```

Perhaps, we should make an anon sub `cc => sub { (qx!egcc -v 2>&1 >$devnull! !~ 'egcc') ? 'gcc' : 'egcc' },` and then on line 548 change `$OS_OPENBSD{cc}` to `$OS_OPENBSD{cc}()`

Kaiepi: I run into the same problem on my machine as well. I'm not sure if the fix you gave worked though:

```
kennel% perl -e "sub a() { (qx!cc -v 2>&1 >$devnull! 'gcc') ? 'gcc' : 'clang' }; print a()"
zsh: event not found: cc
```

Kaiepi: How well would this work?
```perl
cc => system('egcc -v > /dev/null 2>&1') ? 'gcc' : 'egcc',
```

zoffixznet: > I'm not sure if the fix you gave works either

Windows has reversed quotes on the command line, which is why I had `-e ""`  with single quotes for strings. If you're running that command on non-Windows, you need to flip the quotes (also, you don't have `$devnull` variable declared):

    $ perl -e 'sub a() { (qx!cc -v 2>&1 >/dev/null! !~ "gcc") ? "gcc" : "clang" }; print a()'
    clang

zoffixznet:
👍 Thanks!

PR #781, opened 2018-01-18
UCD Cleanup

samcv:


jnthn:
Glad to see that getting cleaned up. :+1:

samcv:
Merging now. Have a bit more but I think it's time to merge since there's already many commits.

PR #784, opened 2018-01-21
Un-bitrot tools/parse_jitgraph.p6

MasterDuke17:
Convert `first-index(...)` to `first(..., :k)`.

It now produces output instead of dying with `No such method 'first-index' for invocant of type 'Array'`.


zoffixznet:
👍 Thanks!

PR #787, opened 2018-01-23
Add JIT template for sp_findmeth

MasterDuke17:
Also mark it invokish in the oplist, since its slow path might invoke,
and the Rakudo build breaks without it. Some chat about this starting
here: https://irclog.perlgeek.de/moarvm/2018-01-23#i_15726145

`sp_findmeth` was the most common op (at 7,890 instances) that caused a `Cannot get template for:` in a JIT log of building Rakudo's CORE.setting.


bdw:
So, two thoughts:
- we currently have a special-cased invokish check for the slow case. (When I implemented that with an expicit `GUARD` node for the expr JIT, I found the `guard` node overwrote my `rdi` register, which is bad, and I have to debug it).
-  This patch makes us pay the cost of the invokish check in all cases, which is suboptimal; then again, making it 'internal' is not compatible with how the expr JIT handles the guard as a barrier, so that optimization will have to go.
- The lego JIT will now do the throwish check twice - in this patch, it should be removed

MasterDuke17:
Is [this](https://github.com/MoarVM/MoarVM/blob/master/src/jit/x64/emit.dasc#L1507-L1542) where the lego JIT does the second throwish check? If so, should it be reduced to just the first/if branch?

MasterDuke17:
Closing because this has already been done.

PR #796, opened 2018-02-03
Fix alignment of sizes for gc allocated objects.

wictory:
Hi! 

So I discovered that there are some alignment problems sometimes in moarvm. I found this using the OpenBSD of rakudo, because the atomic_ops library has alignment assertions enabled by default. The noticed the bug the first time when I ran `zef install OpenSSL`. The output I get from a triggering program is this:
```
$ perl6 -I . -e "use CrashTest;"                         
assertion "((size_t)addr & (sizeof(*addr) - 1)) == 0" failed: file "/usr/local/include/atomic_ops/sysdeps/gcc/../loadstore/atomic_load.h", line 31, function "AO_load"
assertion "((size_t)addr & (sizeof(*addr) - 1)) == 0" failed: file "/usr/local/include/atomic_ops/sysdeps/gcc/../loadstore/atomic_load.h", line 31, function "AO_load"
Abort trap (core dumped)
```

I started investigating this issue in the garbage collector, but @jnthn suggested on IRC that it would be better to fix the issue when the types are created.

Anyhow, I have started with adding assertions all over the place, and I will check if the CI will give same results as I get on my system. After this I will commit a fix.

### DIFF
--- a/src/6model/reprs/P6opaque.c
+++ b/src/6model/reprs/P6opaque.c
@@ -21,6 +21,14 @@ MVM_STATIC_INLINE void set_obj_at_offset(MVMThreadContext *tc, MVMObject *root, 
     MVM_ASSIGN_REF(tc, &(root->header), *((MVMObject **)location), value);
 }
 
+/* Helper for aligning sizes */
+MVM_STATIC_INLINE void align_to(MVMuint64 *size, MVMuint32 align) {
+    if (*size % align) {
+        printf("Aligning something to %u\n", align);
+        *size += align - *size % align;
+    }
+}
+
 /* Helper for finding a slot number. */
 static MVMint64 try_get_slot(MVMThreadContext *tc, MVMP6opaqueREPRData *repr_data, MVMObject *class_key, MVMString *name) {
     if (repr_data->name_to_index_mapping) {
@@ -823,9 +831,7 @@ static void compose(MVMThreadContext *tc, MVMSTable *st, MVMObject *info_hash) {
             /* C structure needs careful alignment. If cur_alloc_addr is not
              * aligned to align bytes (cur_alloc_addr % align), make sure it is
              * before we add the next element. */
-            if (cur_alloc_addr % align) {
-                cur_alloc_addr += align - cur_alloc_addr % align;
-            }
+            align_to(&cur_alloc_addr, align);
 
             /* Attribute will live at the current position in the object. */
             repr_data->attribute_offsets[cur_slot] = cur_alloc_addr;
@@ -861,8 +867,13 @@ static void compose(MVMThreadContext *tc, MVMSTable *st, MVMObject *info_hash) {
                         "While composing %s: Associative delegate attribute must be a reference type", MVM_6model_get_stable_debug_name(tc, st));
             }
 
-            /* Add the required space for this type. */
+            /* Add the required space for this type.
+             *
+             * Note: This is the only place in this function where cur_alloc_addr can
+             * become unaligned.
+             */
             cur_alloc_addr += bits / 8;
+            align_to(&cur_alloc_addr, ALIGNOF(void *));
 
             /* Increment slot count. */
             cur_slot++;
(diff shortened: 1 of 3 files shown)


src/6model/reprs/P6opaque.c
```diff
@@ -21,6 +21,14 @@ MVM_STATIC_INLINE void set_obj_at_offset(MVMThreadContext *tc, MVMObject *root,
     MVM_ASSIGN_REF(tc, &(root->header), *((MVMObject **)location), value);
 }
 
+/* Helper for aligning sizes */
+MVM_STATIC_INLINE void align_to(MVMuint64 *size, MVMuint32 align) {
+    if (*size % align) {
+        printf("Aligning something to %u\n", align);
```
niner: Leftover debug output?

wictory: yes, I will fix that.

src/6model/reprs/P6opaque.c
```diff
@@ -893,17 +906,15 @@ static void deserialize_stable_size(MVMThreadContext *tc, MVMSTable *st, MVMSeri
...
     for (i = 0; i < num_attributes; i++) {
         if (MVM_serialization_read_int(tc, reader)) {
             MVMSTable *st = MVM_serialization_read_stable_ref(tc, reader);
             const MVMStorageSpec *ss = st->REPR->get_storage_spec(tc, st);
             if (ss->inlineable) {
                 /* TODO: Review if/when we get sub-byte things. */
-                if (cur_offset % ss->align) {
-                    cur_offset += ss->align - cur_offset % ss->align;
-                }
+                align_to(&cur_offset, (MVMuint32)ss->align);
```
wictory: Here I have simply translated the `if (...) { ... }` idiom into the `align_to()`, but
shouldn't the alignment happen after `cur_offset +=`?

src/moar.h
```diff
@@ -63,6 +63,8 @@ typedef double   MVMnum64;
 #define ALIGNOF(t) ((char *)(&((struct { char c; t _h; } *)0)->_h) - (char *)0)
 #endif
 
+#define ASSERT_ALIGNED(var, align) assert(!((MVMuint64)(var) % (MVMuint64)(align)))
```
wictory: Is it justified to pollute `src/moar.h` with this macro?

wictory: Also, running on `%` on non-constants is could be costly computationally.

src/6model/reprs/P6opaque.c
wictory: It seems that this one it OK, there should however be an `align_to()` added before the write to `st->size`.

src/moar.h
jnthn: I'd prefer that any macros we add have the `MVM_` prefix, so we know they are ours and so we don't pollute things for any embedders.

jnthn:
There'll likely be more places that need this, so we could in the future lift the `align_to` out. But I guess aligning `P6Opaque` already goes a long way, given it's one of the most common representations.

wictory:
I grepped for `st->size`, to find the places where the size changed to something not coming from a `sizeof()` constant. I assumed that the `sizeof()` constants are aligned, but that could obviously be false under certain circumstances. Also, there might be other places where the `size` is assigned to.

It's a bit annoying that this kind of bug is so difficult to test properly. I was thinking about learning some nqp and and write a compose torture program which would compose all kinds of types and see which size is obtained.

AlexDaniel:
Thank you for the PR, @wictory! It was getting a bit old so I rebased it and then merged it merged into master!

* https://github.com/MoarVM/MoarVM/commit/ab089e947f4520d1789f2aee2958eda4e90cd8fb
* https://github.com/MoarVM/MoarVM/commit/e6e36fbfd778d3ab598b797c15fdf134454b23b1
* https://github.com/MoarVM/MoarVM/commit/bcf3969d8958eea47c82a8f80c1f837ce02629a9
* https://github.com/MoarVM/MoarVM/commit/5d65d9e64b089f3e2ff5bbaaac4290e42067e2d8
* https://github.com/MoarVM/MoarVM/commit/e932be9ac3f528f7b082bcbbac8c19def2d20df2
* https://github.com/MoarVM/MoarVM/commit/1eba4e8b3281b637a03ad7859cfee56947f0f94c
* https://github.com/MoarVM/MoarVM/commit/02486a31900cb0a3c682a26a3efb77c788da447f
* https://github.com/MoarVM/MoarVM/commit/02486a31900cb0a3c682a26a3efb77c788da447f

PR #798, opened 2018-02-07
RFC Allow optional strict decoding of windows-1251/1252 (new ops)

samcv:
This is an RFC for comment on these changes. The op names need to be 
changed, so name suggestions are appreciated. Also I would like comment
on whether I should just change the arguments for any function which
is not used in a MoarVM OP, instead of having redundant functions,
would be appreciated.

---------------------------

We have new versions of `encode`, `decode`, and `encoderep` that allow
the new config flag. `decoderep` never existed before, so this feature
was added to the decoder functions and a function to decode supporting
replacements and a flag was added (`decode2rep`)

Flags for these op currently have only one setting:
Config flag 1:
* Throws when it encounters a character which does not map to the
  other encoding.
* When used under replacement mode, this causes
  replacements to be done with codepoints that may fit into the target
  encoding but are invalid (i.e. 129 in windows-1252).

This adds new ops (names will likely be changed)
* `decode2rep`: we never had a `decoderep` op. This op replaces decoded
   characters that don't have official mappings with a supplied
   replacement string. Currently it is limited to replacement of strings
   with only one grapheme.
* `encode2`: like `encode` except you can set an additional config flag
* `decode2`: like `decode` except you can set an additional config flag
* `encoderep2`: like `encoderep` except you can set an additional config
  flag


jnthn:
Not entirely keen on the `2` in the op name. Maybe `conf` could be better (so, `encodeconf`, `encoderepconf`, `decoderepconf`, etc.)

I think I'd prefer it slightly if the default were to be strict and we had a flag that made it loose.

One other thought: it's tiresome to implement every decoder twice, once streaming and once not. I wonder if we can do everything in terms of the streaming (not at the op level, but at the function level).

samcv:
@jnthn I've updated it so that all the ops end with `config` and the default is to be strict. Full details in the commit message of the updated commit. I will probably merge it after the next release.

samcv:
This was added in to master. Closing.

PR #799, opened 2018-02-08
Add libatomic_ops as a submodule instead of in-tree

samcv:
The old in-tree 3rdparty/libatomic_ops files have been removed and they
are replaced by a submodule. This submodule was created by merging only
the commits and files from 3rdparty/libatomic_ops into the upstream
libatomic_ops repository.


jnthn:
:+1: from me

PR #814, opened 2018-03-08
Rakudo Issue 1588: Create getsignals op to query for valid sigs 

jstuder-gh:
Create the getsignals op in MoarVM which will return a mapping of the
signals supported by MVM and map them to the proper values on the host
system.

See <https://github.com/rakudo/rakudo/issues/1588>


jstuder-gh:
Corresponding Rakudo PR: <https://github.com/rakudo/rakudo/pull/1601>

src/core/interp.c
```diff
@@ -3759,6 +3759,10 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                 GET_REG(cur_op, 0).o = MVM_proc_clargs(tc);
                 cur_op += 2;
                 goto NEXT;
+            OP(getsignals):
+                GET_REG(cur_op, 0).o = MVM_io_get_signals(tc);
+                cur_op += 2;
+                goto NEXT;
```
jnthn: We try to keep the order of implementations in this file following the order of ops in `oplist`; please could you move it?

src/io/signals.c
```diff
@@ -64,118 +64,233 @@ static const MVMAsyncTaskOps op_table = {
...
+        }
+
+        instance->sig_hash = sig_hash;
+        populate_instance_valid_sigs(tc, sig_wanted_vals);
+    });
+
+    return sig_hash;
+}
+
+/* Creates a new timer. */
```
jnthn: This comment seems wrong.

src/io/signals.c
```diff
@@ -64,118 +64,233 @@ static const MVMAsyncTaskOps op_table = {
...
+
+                key = MVM_string_substring(tc, full_key, 4, -1);
+                val = MVM_repr_box_int(tc, hll->int_box_type, sig_wanted_vals[i]);
+
+                MVM_repr_bind_key_o(tc, sig_hash, key, val);
+            });
+        }
+
+        instance->sig_hash = sig_hash;
+        populate_instance_valid_sigs(tc, sig_wanted_vals);
```
jnthn: Do we need to populate `valid_sigs` here?

There's a potential race condition if two threads are in this initialization at once; the first sets `instance->sig_hash`, then the second sees it is set in the condition above above and returns. The caller of the second then assumes `valid_sigs` is ready, but the first may not have finished setting it up yet. This might be avoided if we remove any expectation that this function sets up `valid_sigs`. Alternatively, I think swapping the order of the two lines above this review comment would also suffice.

src/core/interp.c
jstuder-gh: Absolutely. I will move this when I revise later tonight.

src/io/signals.c
jstuder-gh: I will change this to "Creates a new signal handler".

src/io/signals.c
jstuder-gh: I added a check within the MVM_io_signal_handle function that checks that the valid_sigs value is populated prior to attempting to create the signal handler (at [line 288](https://github.com/MoarVM/MoarVM/pull/814/files#diff-a9d671f9f13e3d12535265ab82bde088R288)). Right now MVM_io_signal_handle is the only place where valid_sigs is being used.

The populate_instance_valid_sigs function is idempotent so it should not make a difference if multiple threads are running it simultaneously. That should be enough to guard against unpopulated valid_sigs, right?

src/core/interp.c
jstuder-gh: Moved this to after "getppid" op in the last commit. Not sure why GH doesn't indicate that this has changed like the other items.

src/core/instance.h
```diff
@@ -349,6 +349,12 @@ struct MVMInstance {
     /* Cache the environment hash */
     MVMObject      *env_hash;
 
+    /* Cache the signal hash */
+    MVMObject      *sig_hash;
```
jnthn: I missed this in my last review, but this one needs to be marked in the instance roots function in `src/gc/roots.c`.

jstuder-gh: Thanks. This has been addressed in the latest commits.

jstuder-gh:
I have updated the branch both to address the instance root issue and to update the ops so that it merges cleanly into the current HEAD of master.

jstuder-gh:
In the Rakudo Issue 1588 thread, I proposed an additional change to the signal code to deal with the difference between what signals are available for the host platform and what signals are supported by the backend running Rakudo. You can see the post at <https://github.com/rakudo/rakudo/issues/1588#issuecomment-373210166>.

This would involve registering a redundant op in MoarVM, but would be helpful for backends such as the JVM where not many sigs are supported.

Does the proposal sound good? Should I implement the proposed change into this branch and PR or save it for another time?

jstuder-gh:
Regarding the proposal from the last comment, niner made great suggestion in the other thread to use negative values to indicate lack of VM support, and I implemented just that in the NQP and Rakudo branches, requiring no change in MoarVM. All of my concerns regarding signal usage in P6 have been addressed, so assuming there is approval and that this addresses the community's concerns as well, I believe this is finished.

jstuder-gh:
@jnthn, I believe that this is completed, assuming that it has addressed your concerns.

jnthn:
Indeed, I don't see any further problems. Thanks!

PR #815, opened 2018-03-11
Add proper support for replacements in decodestream for windows1252/1

samcv:
We add the setting for strict/permissive and the replacement string to the
MVMDecodeStream object. Fixes bugs in the previous version which caused it
not to properly throw if strict was set and there was no replacement. Also
adds support for multiple character replacements.

Change arguments for windows1252/1 decodestream ops

We no longer need replacement and config to be arguments since they are
included in the MVMDecodeStream struct. Remove the variations which contain
this as the argument completely.

### DIFF
--- a/src/6model/reprs/Decoder.c
+++ b/src/6model/reprs/Decoder.c
@@ -125,11 +125,9 @@ static int should_translate_newlines(MVMThreadContext *tc, MVMObject *config) {
     return 0;
 }
 static MVMString * has_replacement(MVMThreadContext *tc, MVMObject *config) {
-    char *replacement_const = "replacement";
-    int replacement_const_length = strlen(replacement_const);
     if (IS_CONCRETE(config) && REPR(config)->ID == MVM_REPR_ID_MVMHash) {
         MVMObject *value = MVM_repr_at_key_o(tc, config,
-            MVM_string_ascii_decode(tc, tc->instance->VMString, replacement_const, replacement_const_length));
+            tc->instance->str_consts.replacement);
         return IS_CONCRETE(value)
             ? MVM_repr_get_str(tc, value)
             : NULL;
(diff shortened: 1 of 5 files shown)


samcv:
@jnthn asking for your input on this. With modifications to rakudo this is working but I wanted to ask you how you felt about adding the replacement and strict/permissive config to MVMDecodeStream struct. Looking through the code, it seemed to be the only sane way, and makes sure we don't have to change more than a dozen functions to add additional arguments, because the MVMDecodeStream object gets passed around quite a bit. This also causes us to not have to add a new op.

src/6model/reprs/Decoder.c
```diff
@@ -124,6 +124,29 @@ static int should_translate_newlines(MVMThreadContext *tc, MVMObject *config) {
     }
     return 0;
 }
+static MVMString * has_replacement(MVMThreadContext *tc, MVMObject *config) {
+    char *replacement_const = "replacement";
+    int replacement_const_length = strlen(replacement_const);
+    if (IS_CONCRETE(config) && REPR(config)->ID == MVM_REPR_ID_MVMHash) {
+        MVMObject *value = MVM_repr_at_key_o(tc, config,
+            MVM_string_ascii_decode(tc, tc->instance->VMString, replacement_const, replacement_const_length));
```
jnthn: Please add this to the `str_consts` in instance. Otherwise `MVM_string_ascii_decode` may trigger GC. We can `MVMROOT` instead, but faster to not do this work again and again.

src/strings/decode_stream.h
```diff
@@ -37,6 +37,9 @@ struct MVMDecodeStream {
     /* Optional place for the decoder to keep any extra state it needs between
      * decode calls. Will be freed when the decode stream is destroyed. */
     void *decoder_state;
+
+    MVMString *replacement;
```
jnthn: This will need to be GC marked.

src/6model/reprs/Decoder.c
samcv: Yes that is something I was going to do before finalizing this.

src/strings/windows1252.c
```diff
@@ -340,25 +340,18 @@ static MVMuint8 windows1251_cp_to_char(MVMint32 codepoint) {
...
     MVMuint32 reached_stopper;
-    MVMStringIndex repl_length;
-    if (replacement) {
-        repl_length = MVM_string_graphs(tc, replacement);
-        if (1 < repl_length) {
-            MVM_exception_throw_adhoc(tc,
-            "Windows-1252 and Windows-1251 decoding with a replacement string with more than one grapheme is not yet implemented.\n"
-            "Provided string had %"PRIu32" graphemes\n", repl_length);
-        }
-    }
```
jnthn: I see this is now supported - but there's a problem still: what if the grapheme in the replacement isn't a valid one under this encoding also?

samcv: This is just the decoder, so I wouldn't think that would be an issue. Unless you can think of some instances when it could be?

jnthn: Oh, goodness, you're right. Sorry, should probably not do review this close to sleep time. :-)

samcv:
@jnthn can you take a look at it again. I think I corrected your suggestions correctly. Thanks!

PR #816, opened 2018-03-12
More templates to the eleventeenth power

MasterDuke17:


MasterDuke17:
@bdw should be good to go.

PR #824, opened 2018-03-25
Allow NativeCall support for wchar_t

Kaiepi:
A library I'm writing bindings for uses `wchar_t` in some places, so I thought it'd be useful to be able to use it as a type. NQP and Rakudo pullreqs will come later

### DIFF
--- a/src/6model/reprs/P6int.c
+++ b/src/6model/reprs/P6int.c
@@ -145,6 +145,7 @@ static void compose(MVMThreadContext *tc, MVMSTable *st, MVMObject *info_hash) {
                 case MVM_P6INT_C_TYPE_BOOL:     repr_data->bits = 8 * sizeof(char);      break;
 #endif
                 case MVM_P6INT_C_TYPE_ATOMIC:   repr_data->bits = 8 * sizeof(AO_t);      break;
+                case MVM_P6INT_C_TYPE_WCHAR_T:  repr_data->bits = 8 * sizeof(wchar_t);   break;
             }
 
             if (repr_data->bits !=  1 && repr_data->bits !=  2 && repr_data->bits !=  4 && repr_data->bits != 8
--- a/src/6model/reprs/P6int.h
+++ b/src/6model/reprs/P6int.h
@@ -6,6 +6,7 @@
 #define MVM_P6INT_C_TYPE_SIZE_T    -6
 #define MVM_P6INT_C_TYPE_BOOL      -7
 #define MVM_P6INT_C_TYPE_ATOMIC    -8
+#define MVM_P6INT_C_TYPE_WCHAR_T   -9
 
 /* Representation used by P6 native ints. */
 struct MVMP6intBody {


niner:
For full NativeCall support, you're gonna need a couple for changes in src/core/nativecall.c, src/core/nativecall_dyncall.c and src/core/nativecall_libffi.c basically wherever you see  MVM_NATIVECALL_ARG_LONG mentioned. I guess with your patch support is restricted to wchar_t arrays.

Kaiepi:
Would that be for making `Str` work instead of `CArray[wchar_t]`? What do I need to change?

Kaiepi:
I tested this with the following code:
```c
#include <stddef.h>

extern const wchar_t *test(const wchar_t* foo) {
     return foo;
}
```

When I test in perl6, this happens with `Str`:
```perl-6
> sub test(Str --> Str) is native('./libtest.so') { * }
sub test (Str $ --> Str) { #`(Sub+{Callable[Str]}+{NativeCall::Native[Sub+{Callable[Str]},Str]}|18948345937280) ... }
> test('test')
test
> test('test')
test
Malformed UTF-8 at line 1 col 1
  in block <unit> at <unknown file> line 1
```

With `CArray[wchar_t]`, the function returns an empty instance of `CArray[wchar_t]`. So you're right, the implementation is not quite complete yet.

Kaiepi:
OK, you're right about the support being incomplete:
```c
#include <stdio.h>
#include <wchar.h>

extern wchar_t *test(const wchar_t *foo) {
	wchar_t *bar;
	wint_t len = wcslen(foo);
	wmemcpy(bar, foo, len);
	return bar;
}
```
```perl-6
> use NativeCall
Nil
> sub test(CArray[wchar_t] --> CArray[wchar_t]) is native('./libtest.so') { * }
sub test (NativeCall::Types::CArray[NativeCall::Types::wchar_t] $ --> NativeCall::Types::CArray[NativeCall::Types::wchar_t]) { #`(Sub+{Callable[NativeCall::Types::CArray[NativeCall::Types::wchar_t]]}+{NativeCall::Native[Sub+{Callable[NativeCall::Types::CArray[NativeCall::Types::wchar_t]]},Str]}|34912317730176) ... }
> test(CArray[wchar_t].new: 'test'.ords)
NativeCall::Types::CArray[NativeCall::Types::wchar_t].new
> my $a = test(CArray[wchar_t].new: 'test'.ords)
Internal error: unhandled dyncall argument type
  in block <unit> at <unknown file> line 1
```
I'll read through the nativecall files and see what I'll need to do to get full support implemented.

Kaiepi:
Something I noticed trying to use `wchar_t` with strings is that it doesn't get encoded properly, with "012" becoming "48 49 50" it becomes "48 0 0 0 49 0 0 0 50 0 0 0". I tested casting a `wchar_t *` to a `char *` and got the same results. I think for full support I'll need to do a lot more work.

I also forgot to add `wint_t` lol

Kaiepi:
Changed my mind about `wint_t`. There wasn't a good way that I know of to get the right values of it for the JVM

Kaiepi:
I'll redo this properly with support for `wchar_t`, `wint_t`, `char16_t` and `char32_t` at some point.

PR #826, opened 2018-03-27
Refactor optimize_iffy

bdw:
Over time optimize_iffy has accumulated a number of special-case optimizations like lowering-the-conditional from istrue to known values or type-specific optimizations, and even an unbox-remover.
Also, it was clear that having a 'normal' optimize_iffy case and a special case for `if_o` and `unless_o` splitting was far too complex.

I felt like it'd be cleaner if these optimizations were separated out of optimize_iffy and into their own optimizations. So, that is what happens now:

- rather than split 'remaining' `if_o` after `optimize_iffy`, split them directly, then lower the resulting `istrue`
   - push the split `if_i` forward rather than `istrue` backwards, which makes sure that the 
- teach `optimize_istrue_isfalse` all the tricks that `optimize_iffy` knew
  - make the `isfalse` split into boolifier + not_i happen *after* the boolification-lowering (also, simpler)
- take the `unbox`-removal out into its own optimization, although that doesn't really work yet
  - this needs some more thinking - or maybe just allocating a unique temporary register for it

This also fixed the bug that was fixed by nine++ and timotimo++ independently, which is that decompose_object_conditional would remove a branch twice in some cases

### DIFF
--- a/src/spesh/optimize.c
+++ b/src/spesh/optimize.c
@@ -918,55 +918,93 @@ static void optimize_istrue_isfalse(MVMThreadContext *tc, MVMSpeshGraph *g, MVMS
     /* Unknown value, known type. We may be able to lower this to a
      * nonpolymorphic operation */
     if (input_facts->flags & MVM_SPESH_FACT_KNOWN_TYPE) {
-        MVMBoolificationSpec *bs = STABLE(input_facts->type)->boolification_spec;
-        MVMSpeshOperand  orig    = ins->operands[0];
-        MVMSpeshOperand  temp;
+        /* Go by boolification mode to pick a new instruction, if any. */
+        MVMObject *type            = input_facts->type;
+        MVMBoolificationSpec *bs   = type->st->boolification_spec;
+        MVMuint8 guaranteed_concrete = input_facts->flags & MVM_SPESH_FACT_CONCRETE;
+        MVMuint8 mode = bs == NULL ? MVM_BOOL_MODE_NOT_TYPE_OBJECT : bs->mode;
+        MVMSpeshOperand orig = ins->operands[0];
 
-        if (negated_op)
-           temp = MVM_spesh_manipulate_get_temp_reg(tc, g, MVM_reg_int64);
-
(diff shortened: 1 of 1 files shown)


src/spesh/optimize.c
```diff
@@ -1065,60 +951,182 @@ static void optimize_istrue_isfalse(MVMThreadContext *tc, MVMSpeshGraph *g, MVMS
...
+             op_info = MVM_op_get_op(MVM_OP_unbox_i);
+             break;
+            */
+        default:
+            return;
         }
-        /* Now we can take care of the negation. */
+//        fprintf(stderr, "Known type optimized to %s\n", ins->info->name);
+        /* Now we can take care of the negation. - NB I'm not entirely sure why
+         * this would need it's own register though! */
```
jnthn: Because otherwise we end up with multiple assigns to the same version of the register and violate SSA form.

src/spesh/optimize.c
```diff
@@ -1065,60 +951,182 @@ static void optimize_istrue_isfalse(MVMThreadContext *tc, MVMSpeshGraph *g, MVMS
...
             new_ins->info = MVM_op_get_op(MVM_OP_not_i);
             new_ins->operands = operands;
             operands[0] = orig;
             operands[1] = temp;
+            ins->operands[0] = temp;
             MVM_spesh_manipulate_insert_ins(tc, bb, ins, new_ins);
-            MVM_spesh_get_facts(tc, g, temp)->usages++;
             MVM_spesh_manipulate_release_temp_reg(tc, g, temp);
+
+            /* Set facts on temporary - NB this is *very* confusing */
```
jnthn: Why?

src/spesh/optimize.c
bdw: Yeah, I figured that out, I'll add a comment to that effect.
We could've had a 'new version' rather than an entirely new register, though, that would have had the same effect.

src/spesh/optimize.c
bdw: It did something different before, that was also wrong :smile:

bdw:
I had missed a patch on the isconcrete optimization that removed the comment (and fixed the bug there), pushed now, should be good to merge

PR #830, opened 2018-03-30
Speed up collapse_strands 4x by full rewrite targeting SIMD optims

samcv:
collapse_strands and certain conditions of substr rely on
iterate_gi_into_string() for flattening a specified section of a string.
This has been a performance issue since forever, and though I made
changes to make it faster, it still had issues. These new changes are a
full rewrite of that function and a new function which checks if a
string for the specified length can fit into 8 or 32 bits.

Both these functions have been rewritten to take advantage of compiler
vectorization which will utilize instructions such as SSE* or AVX*.

In string_can_be_8bit() we use multiple bitwise AND operations to check
the 32 bit integers while ORing the result with `val`.
Care must be taken so that any variables altered by the loop are not used
outside of the loop. Even more critical, there cannot be  any branching or
function calls. For example: `i` is not used outside the loop, while `val`
is allowed as biwise OR works with vectorization well.

To support both of these changes a new grapheme iterator function has
been added called MVM_string_gi_next_strand() which moves to the next
strand. This allows us to process each strand at a time without having
to process each grapheme at a time. This was needed to either memcpy()
or use vectorizable loops to copy the graphemes based on strand bit
type.

### DIFF
--- a/src/strings/iter.h
+++ b/src/strings/iter.h
@@ -53,6 +53,24 @@ MVM_STATIC_INLINE void MVM_string_gi_init(MVMThreadContext *tc, MVMGraphemeIter 
     (gi->end - gi->pos + gi->repetitions * (gi->end - gi->start))
 /* graphs left in strand + graphs left in repetitions of current strand */
 
+static void MVM_string_gi_next_strand (MVMThreadContext *tc, MVMGraphemeIter *gi) {
+    MVMStringStrand *next = NULL;
+    if (gi->repetitions) {
+        gi->pos = gi->start;
+        gi->repetitions--;
+        return;
+    }
+    if (gi->strands_remaining <= 0)
+        MVM_exception_throw_adhoc(tc, "Iteration past end of grapheme iterator\n");
+
+    next = (gi->next_strand)++;
+    gi->pos = gi->start = next->start;
+    gi->end             = next->end;
+    gi->repetitions     = next->repetitions;
+    gi->blob_type       = next->blob_string->body.storage_type;
+    gi->active_blob.any = next->blob_string->body.storage.any;
+    gi->strands_remaining--;
+}
 /* Sets the position of the iterator. (Can be optimized in many ways in the
  * repetitions and strands branches.) */
 MVM_STATIC_INLINE void MVM_string_gi_move_to(MVMThreadContext *tc, MVMGraphemeIter *gi, MVMuint32 pos) {
--- a/src/strings/iter.h
+++ b/src/strings/iter.h
@@ -48,12 +48,13 @@ MVM_STATIC_INLINE void MVM_string_gi_init(MVMThreadContext *tc, MVMGraphemeIter 
     }
 };
 /* Gets the number of graphemes remaining in the current strand of the grapheme
- * iterator */
-#define MVM_string_gi_graphs_left_in_strand(gi) \
+ * iterator, including repetitions */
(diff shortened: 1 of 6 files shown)


src/strings/iter.h
```diff
@@ -53,6 +53,24 @@ MVM_STATIC_INLINE void MVM_string_gi_init(MVMThreadContext *tc, MVMGraphemeIter
...
 
+static void MVM_string_gi_next_strand (MVMThreadContext *tc, MVMGraphemeIter *gi) {
+    MVMStringStrand *next = NULL;
+    if (gi->repetitions) {
+        gi->pos = gi->start;
+        gi->repetitions--;
+        return;
+    }
+    if (gi->strands_remaining <= 0) {
+        MVM_exception_throw_adhoc(tc, "ff\n");
```
jnthn: ff? Is that the noise a cat makes when throwing an exception? :-)

src/strings/ops.c
```diff
@@ -206,38 +205,130 @@ static void turn_32bit_into_8bit_unchecked(MVMThreadContext *tc, MVMString *str)
...
+ * to not use any variables altered by the loop outside of the loop and to not
+ * have any branching or funcion calls. `i` is not used outside the loop
+ * `val` is allowed as biwise OR works with the vectorization well.
+ * NOTE: Final state of the grapheme iterator passed into i is undefined. It should
+ * not be reused. */
+static int string_can_be_8bit(MVMThreadContext *tc, MVMGraphemeIter *gi, MVMStringIndex num_graphs) {
+    int val = 0;
+    MVMStringIndex pos = 0;
+    while (1) {
+        size_t strand_len = gi->end - gi->pos;
```
jnthn: I think I'd maybe prefer that we introduce static inline functions for dealing with the grapheme iterator data structure, rather than poking in to it. The design so far was - at least hopefully - such that we didn't poke into the struct directly, but worked with it in terms of static inline functions. Since they're effectively just a nicer way to do macros, they shouldn't have any performance impact, while keeping it easier to change things in the future.

src/strings/ops.c
```diff
@@ -206,38 +205,130 @@ static void turn_32bit_into_8bit_unchecked(MVMThreadContext *tc, MVMString *str)
...
+            size_t i = 0;
+            size_t togo     = strand_len > num_graphs - pos ? num_graphs - pos : strand_len;
+            size_t togofull = togo + gi->pos;
+            for (i = gi->pos; i  < togofull; i++) {
+                val |= ((gi->active_blob.blob_32[i] & 0xffffff80) + 0x80) & (0xffffff80-1);
+            }
+            if (val) return 0;
+            pos += togo;
+        }
+        else {
```
MasterDuke17: Both statements in this else branch are also in the if branch. Couldn't both be brought out of the if/else entirely, with `size_t togo = strand_len > num_graphs - pos ? num_graphs - pos : strand_len;` going before the if and `pos += togo;` going after the if (and the else going away). I.e.,
```
size_t togo = strand_len > num_graphs - pos ? num_graphs - pos : strand_len;
if (gi->blob_type == MVM_STRING_GRAPHEME_32) {
    size_t i = 0;
    size_t togofull = togo + gi->pos;
    for (i = gi->pos; i  < togofull; i++) {
        val |= ((gi->active_blob.blob_32[i] & 0xffffff80) + 0x80) & (0xffffff80-1);
    }
    if (val) return 0;
}
pos += togo;
```

PR #831, opened 2018-03-30
Add missing write barrier enforcement in CPPStruct

dogbert17:
This was done earlier for CStruct, by jnthn++, in commit 85fc758ce976.
Stress testing of t/04-nativecall/11-cpp.t showed that the same fix
had to be applied to CPPStruct as well.


jnthn:
Only Travis failure looks like an apt repo fail. So, merging. Thanks!

Issue #839, opened 2018-04-09
Move 3rdparty/libatomic_ops to a different directory?


zoffixznet:
This is in reference to a recent change that [requires deleting `3rdparty/libatomic_ops` to unbreak the build](https://alerts.perl6.org/alert/7).

Is there any way for us to assuage the damage by, say, moving that stuff to a different directory?

The issue affects everyone [using build-from-source instructions](https://github.com/zoffixznet/r) as well as [rakudobrew users](https://irclog.perlgeek.de/perl6/2018-04-08#i_16021418). That's quite a lot of users, not all of whom would know to talk to someone on IRC.

If we move it to a different dir, then anyone checking out latest commit won't be affected, would they?

samcv:
@zoffixznet I made a commit that should help a lot with this https://github.com/MoarVM/MoarVM/commit/ef4e6fe9455f71afc81413a5590435bb955d3806

It shows the user a pretty visible message telling them exactly what went wrong and what folder to delete.

The message shows at the very bottom of the ./Configure.pl run for best
visibility.

`ERROR: Cannot update submodule because directory exists and is not empty.`
`>>> Please delete the following folder and try again:`
`/home/perl/.rakudobrew/moar-blead-master/nqp/MoarVM/3rdparty/libatomic_ops`

zoffixznet:
So does that mean using a different directory is not an option?

For the record, that fix doesn't solve the problem that motivated my filing this issue at all: my program had erroneous behaviour it didn't use to have and I was trying to find a rakudo build that didn't have the issue. I crossed the boundary where `3rdparty/libatomic_ops` problem was introduced a few times and that required me deleting that directory each time. Worse yet, I'd fire off the build and leave to do Other Things™ only to come back and realize it still haven't built anything due to that problem.

The more visible error is an improvement, but I rather the users didn't have to do anything at all.

samcv:
@zoffixznet I could have Configure.pl delete the directory, but I don't know if that is the right option. The directory *could* be renamed I suppose. I wanted to explore other options as well though.

jnthn:
Please just rename the directory. I [gently suggested that before](https://irclog.perlgeek.de/moarvm/2018-02-20#i_15840035), and in hindsight should have just said it must be done that way.

samcv:
@jnthn okay I will do that. Thanks.

jnthn:
Seems it was done; closing.

PR #840, opened 2018-04-11
Move libatomic_ops module to a different directory

samcv:
This should make sure there aren't issues when git fast forwards.

Fixes issue #839


timo:
maybe we should also rename the old folder during Configure or something so that people don't get confused, because they'll end up with two folders for libatomicops? And i say rename because perhaps someone has some legit local changes to the lib code that we might not want to just nuke.

samcv:
@timo we can't rename because then the git reference will be wrong since the folder is a git repository itself and it stores where it thinks it should be located.

PR #841, opened 2018-04-12
Randomly seed a hash secret to mitigate ease of DOS attack

samcv:
Without a random hash seed it is easy for an attacker to generate
strings which will result in the same hash. This devolves to O(n**2)
time for the hash insertion and is increased even more by the fact that
a string must be compared with all strings in the bucket until the right
string is found.

This attack is done by creating a function that essentially is our
hashing function backward. We hash our target string, `t`. We then
use random 3 character sequences (in our case graphemes) and plug them
into our backward hashing function along with the hash for our target
`t`. The backward hash and the random character sequence are stored in
the dictionary and the process is repeated until we have a very large
number of backward hash's and random 3 grapheme prefixes.

We can then use this dictionary to construct successively longer strings
(or short if we so desire) which are the same hash as our target string
`t`.

This has been fixed in most programming languages (Python, Ruby, Perl),
and several CVE's have been issues over the years for this exploit.

It may also be a good idea to later implement a stronger hashing
function. Many languages are now using SipHash which is meant to protect
against an attacker discovering a hash secret remotely. This change decreases
the ease of this attack and makes the hash secret
unpredictable.

Randomness source:
We prefer function calls rather than reading from /dev/urandom. Reasons
include: not having to open a file descriptor and that /dev/urandom may
not exist if we are in a chroot. Since we don't want to stop startup of
MoarVM, if we have to fallback to /dev/urandom since the OS doesn't
support the function call we continue starting MoarVM normally.
Linux, FreeBSD, OpenBSD and MacOS all use system provided random calls
to get the data rather than having to open /dev/urandom. All these OS's
guarantee these to be non-blocking, though MacOS's documentation does
not comment on it.
Whether the calls block is primarily a concern during very early boot
which is why Python 3 makes sure to use non-blocking calls when their
hash secret is seeded an interpreter start up.

If not available we fall back to using /dev/urandom on Unix like OS's.
This change was tested on Linux both pre-addition of `getrandom()` and
after, as well as Solaris, FreeBSD and OpenBSD. CI testing for Windows
and MacOS look good. All Unix are supported due to /dev/urandom fallback
and Windows is supported to version 95 with the API we use.

### DIFF
--- a/3rdparty/uthash.h
+++ b/3rdparty/uthash.h
@@ -295,7 +295,7 @@ do {                                                                            
 do {                                                                             \
   unsigned _hj_i,_hj_j,_hj_k;                                                    \
   unsigned char *_hj_key=(unsigned char*)(key);                                  \
-  hashv = 0xfeedbeef;                                                            \
+  hashv = tc->instance->hashSecret;                                                            \
   _hj_i = _hj_j = 0x9e3779b9;                                                    \
   _hj_k = (unsigned)(keylen);                                                    \
   while (_hj_k >= 12) {                                                          \
--- a/3rdparty/uthash.h
+++ b/3rdparty/uthash.h
@@ -295,7 +295,7 @@ do {                                                                            
 do {                                                                             \
   unsigned _hj_i,_hj_j,_hj_k;                                                    \
   unsigned char *_hj_key=(unsigned char*)(key);                                  \
-  hashv = tc->instance->hashSecret;                                                            \
+  hashv = tc->instance->hashSecret;                                              \
   _hj_i = _hj_j = 0x9e3779b9;                                                    \
   _hj_k = (unsigned)(keylen);                                                    \
   while (_hj_k >= 12) {                                                          \
--- a/src/moar.c
+++ b/src/moar.c
@@ -88,7 +88,6 @@ MVMInstance * MVM_vm_create_instance(void) {
     MVMuint32 hashSecret;
     MVMuint64 now = MVM_platform_now();
 
-
     /* Set up instance data structure. */
     instance = MVM_calloc(1, sizeof(MVMInstance));
 
--- a/src/moar.c
+++ b/src/moar.c
@@ -1,6 +1,7 @@
 #include "moar.h"
 #include <platform/threads.h>
-
+#include "platform/random.h"
+#include "platform/time.h"
 #if defined(_MSC_VER)
 #define snprintf _snprintf
 #endif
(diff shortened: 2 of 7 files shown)


3rdparty/uthash.h
```diff
@@ -295,7 +295,7 @@ do {
 do {                                                                             \
   unsigned _hj_i,_hj_j,_hj_k;                                                    \
   unsigned char *_hj_key=(unsigned char*)(key);                                  \
-  hashv = 0xfeedbeef;                                                            \
+  hashv = tc->instance->hashSecret;                                                            \
```
Tyil: This backslash is out of line compared to the rest of the backslashes.

src/moar.c
```diff
@@ -84,12 +85,17 @@ MVMInstance * MVM_vm_create_instance(void) {
     char *jit_log, *jit_expr_disable, *jit_disable, *jit_bytecode_dir, *jit_last_frame, *jit_last_bb;
     char *dynvar_log;
     int init_stat;
+    MVMuint32 hashSecret;
+    MVMuint64 now = MVM_platform_now();
+
```
Tyil: Only a single line of whitespace should be enough here.

src/moar.c
```diff
@@ -84,12 +85,17 @@ MVMInstance * MVM_vm_create_instance(void) {
...
 
     /* Set up instance data structure. */
     instance = MVM_calloc(1, sizeof(MVMInstance));
 
     /* Create the main thread's ThreadContext and stash it. */
     instance->main_thread = MVM_tc_create(NULL, instance);
+    instance->hashSecret ^= now;
+    instance->hashSecret ^= MVM_proc_getpid(instance->main_thread) * now;
```
niner: Err....shouldn't this use MVM_getrandom()?

samcv: Yes, I removed some debugging code and that got removed too!

zhuomingliang:
@samcv https://github.com/google/highwayhash, FYI .

src/platform/random.c
```diff
@@ -61,35 +60,64 @@
  * All BSD's should support arc4random
  * AIX is a unix but has no arc4random, does have /dev/urandom */
 #include "moar.h"
+/* On on Unix like platforms that don't support getrandom() or getentropy()
```
AlexDaniel: On on

src/platform/random.c
```diff
@@ -61,35 +60,64 @@
...
+ * we use /dev/urandom. On platforms that do support them, we fall back to
+ * /dev/urandom. This is also important on Linux, since if MoarVM was compiled
```
AlexDaniel: Confusing… so we fall back to /dev/urandom either way?

samcv:
@jnthn when you get a chance, can you review this pull request?

PR #843, opened 2018-04-20
Add 'slice' op

jstuder-gh:
This PR is a new op 'slice', inspired by @lizmat's [RFC issue](https://github.com/rakudo/rakudo/issues/1639) and implemented for VMArray.

Repurposes the element copying functionality from VMArray's 'splice' implementation to create a new VMArray array from slice of existing one.

[Profile NQP slicing (~3 secs total)](https://rawgit.com/jstuder-gh/97ab0f1baaf24e022fc2571eabec016e/raw/a9f25c8390a29436eec1e9aa4da8cb3e4b122ec1/profile_slice_nqp.html)
[Profile P6 slicing (~42 secs total)](https://rawgit.com/jstuder-gh/9a4432d867905c5b435cea3430ec5961/raw/63158459bbaa608950aead6c52ec328d6f980ad7/profile_slice_p6.html)

The above profiles are measuring runs of [this script](https://gist.github.com/jstuder-gh/b59ffcae9b6a0590d4e64f6bf4fec824) (once with --nqp opt, once without)


src/6model/reprs/VMArray.c
```diff
@@ -792,15 +792,99 @@ static void shift(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *da
...
+    MVMArrayREPRData *d_repr_data = REPR(dest)->ID == MVM_REPR_ID_VMArray
+                                      ? STABLE(dest)->REPR_data : NULL;
+
+    MVMint64 last_elem = REPR(src)->elems(tc, st, src, s_body) - 1;
+    MVMint64 elems;
+
+    if ( last_elem < start || last_elem < end
+         || (end < start && 0 <= start && 0 <= end) )
+    {
+        MVM_exception_throw_adhoc(tc, "MVMArray: Slice Index Out of Bounds");
```
MasterDuke17: I would lowercase "Index Out of Bounds" to be more consistent with the rest of the "MVMArray:" errors.

MasterDuke17:
I'm sure people are going to want this jitted if it gets added. Have you tried jitting it?

jstuder-gh:
I have not. I'll have to look more closely at the JIT documentation. Right now I don't really know how it all  works, to be honest.

jstuder-gh:
Do you know of any good resources not just on our JIT, but JITing in general? I could probably stand to read up on it a bit :P

src/6model/reprs/MultiDimArray.c
```diff
@@ -459,6 +459,9 @@ static void unshift(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *
 static void shift(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *data, MVMRegister *value, MVMuint16 kind) {
     MVM_exception_throw_adhoc(tc, "Cannot shift a fixed dimension array");
 }
+static void aslice(MVMThreadContext *tc, MVMSTable *st, MVMObject *src, void *data, MVMObject *dest, MVMint64 start, MVMint64 end) {
+    MVM_exception_throw_adhoc(tc, "Multidim slice not yet implemented");
```
timo: I don't think the given signature makes sense for multidim arrays, so maybe it should die like asplice just below?

timo:
There's multiple facets to jitting an op. You're adding a reprop, i.e. an op that just looks up a function pointer from a REPR (for example "this object uses the VMArray REPR") and calls that; we have a special optimization for those in our regular jit. But in general, we can make the distinction between two kinds of ops. One kind just takes some stuff out of registers, calls some MVM_ function, and optionally stashes the return value in some other register. Another kind has some more stuff to do, like check if the right REPR is present. The last kind doesn't involve calling C functions at all. You can tell the difference in the interp.c implementation of the given op.

For ops that really just take values from registers in some way and call a C function and optionally do something with the return value, we have an API that makes this easy: Some simple examples start here: https://github.com/MoarVM/MoarVM/blob/master/src/jit/graph.c#L2467 - your corresponding C function also has to be added to op_to_func here: https://github.com/MoarVM/MoarVM/blob/master/src/jit/graph.c#L96

For the repr-specific optimization, look at https://github.com/MoarVM/MoarVM/blob/master/src/jit/graph.c#L829 - the idea is that if the bytecode specializer has already made sure that the type is known, we can pull up the REPR for the given type and instead of calling functions from reprconv.c that forward to the real function by looking up what the REPR is, we just put the right function right in the assembly code.

I hope that kind of helps :)

jstuder-gh:
@timo, it more than kind of helps. Thanks!

I do have one other question. How do the [expression templates](https://github.com/MoarVM/MoarVM/blob/master/src/jit/core_templates.expr) factor in? Are they for the legacy JIT? And would there be any benefit in attempting to implement one for this op?

I added a case to jit the slice op in graph.c linked above. The interface provided seems pretty straight-forward, but if any more experienced members have a bit of time to review, it would be appreciated :)

jstuder-gh:
I think the slice jit is working but it doesn't seem to affect the results too much, even in longer running scripts (like when turning using ```(^10).permutations``` in [this script](https://gist.github.com/jstuder-gh/b59ffcae9b6a0590d4e64f6bf4fec824)).

One run might have a difference of 100-200 ms in favor of an interp-only execution and then the next run might be 100-200 ms in favor of the jitted execution.

timo:
Hey, sorry for taking ages to respond, I didn't notice this issue had some more comments.

The expression templates thing is "the new jit", and it's basically "plugged into" the legacy jit so that the legacy jit can take over whenever the exprjit doesn't have a template for some op, or faces some other situation it can't yet handle. It re-attempts the exprjit compilation after every basic block boundary, though, so one missing exprjit template doesn't doom the whole compilation (unlike a missing op for the legacy jit which causes the whole jitting process for the given routine to be aborted).

It could very well be that this is actually what's making the jitted vs non-jitted execution so similar in run time on your end; try either --profile or setting MVM_JIT_LOG to a filename and searching for BAIL in the resulting file to see if the subs are actually getting the jit treatment or not. I'll try to have a look myself later today, though!

timo:
I've looked into it some more: in the nqp path of the example code (the one based on permutations) the frame containing the slice op is not actually running jitted, and it's because of an op that I recently introduced that doesn't get jitted yet: getstringfrom. I've got an exprjit implementation, but not a template jit one, and the BB it's in is even the same one that contains slice, but it has an op that bails out the exprjit before it gets to that op.

In short, if I implement gestringfrom (and i should also implement the others i recently added) could start seeing a difference.

jstuder-gh:
So the slice jit template was never being jitted at all? That would explain the results.

I have a few more questions about the jit. I'm trying to get a sense of how it works, if only at a high level.

So if I understand correctly, the jit creates a series of jit graphs made up of basic blocks (which themselves are constructed from the expr jit templates and augmented with the legacy jit nodes). Tiles are defined that map the blocks of the expression graphs to machine code (for a given architecture) and the resulting machine code is substituted for hot code. Is this correct or am I off?

What do the basic blocks and jit graphs correspond to? Do the graphs directly map to frames/routines in the HLL?

So when creating a new op, it is a good idea to attempt a expr jit template and include a legacy jit template as a fallback? The legacy jit it seems is important to include as omitting that bails on the construction of that entire jit graph, but I presume that there would be some benefit to rewriting as many ops as possible for the expr jit (more effective retrieval of branching code due to it's graph structure)?

Don't worry about the time delay. I appreciate you taking the time to answer my questions.

jstuder-gh:
I've created and uploaded an exprjit template for the 'slice' op.

I've added it to the branch a bit tentatively, as I haven't really seen it in action yet. I haven't found a scenario yet where it does not either bail or fallback to legacy templates before utilizing the 'slice' exprjit. But I wanted to submit it so that it could be reviewed.

If it turns out that there is a problem with this implementation, I will of course remove it from the branch.

jstuder-gh:
@timo, if you have the time, could you review this again? I believe that this functionality is complete and, if you and @jnthn believe it would be helpful, is ready to be merged.

The accompanying NQP PR ([located here](https://github.com/perl6/nqp/pull/436)) also ports the op to the JVM P6 runtime and adds tests and a bit of documentation. Also worth noting that the Travis [JVM build passes](https://travis-ci.org/perl6/nqp/builds/374836710?utm_source=github_status&utm_medium=notification) the newly created tests (MoarVM ones fail for obvious reasons).

lizmat:
Re: lib/MAST/Ops.nqp : I thought it was a problem if existing ops have their number changed?  aka, shouldn't `slice` be behind `breakpoint` ?

jstuder-gh:
Looks like one of the [Appveyor builds failed](https://ci.appveyor.com/project/samcv/moarvm-35u74/build/1.0.653/job/hlnt03b2rf79o7sl#L37), but only because it could not find Strawberry Perl. One of those odd flukes?

ugexe:
I'd say that is because the appveyor file was written to only use `--version 5.20.1.1` for some reason ( although I can not think of a reason for this restriction )

PR #845, opened 2018-04-23
Use an array instead of a big switch for ShiftJIS decoding

samcv:
Since we can easily map index->codepoint, turn the big switch into an
array. There are only 11 gaps of the indexes, so we just check we aren't
in a gap and apply an offset. Makes the shiftjis files compile twice as
fast.

We still use a big switch for encoding because codepoints are not at all
in order, so for now let the compiler handle this.


samcv:
Closing since I commited directly.

PR #851, opened 2018-04-27
Fix installs having an outdated version

Kaiepi:
This was breaking configuring nqp and Rakudo when I was building all three manually


Kaiepi:
Never mind, I tested again and this didn't work

Kaiepi:
The problem was just me forgetting to `git fetch` before building. My bad

PR #852, opened 2018-04-29
Clean up src/jit/core_templates.expr a bit

MasterDuke17:
Put jit templates in same order as ops in interp.c. Also try to make formatting consistent across all the templates.

### DIFF
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -15,21 +15,19 @@
 
 (template: set (copy $1))
 
-(template: trunc_i8! (store $0 $1 1))
+(template: trunc_i8!  (store $0 $1 1))
 (template: trunc_i16! (store $0 $1 2))
 (template: trunc_i32! (store $0 $1 4))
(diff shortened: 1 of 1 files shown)


MasterDuke17:
First commit inspired by https://irclog.perlgeek.de/moarvm/2018-03-12#i_15911914, second was a bonus. @bdw @niner @timo you three are the only other people to write templates, I just merged together all the different formatting styles into one, let me know if anybody objects to how I did it.

bdw:
It looks good and clean; it is unfortunately too large to 'review' by hand, so if this passes tests, I trust that it is good. The only drawback is that it makes code archeology a bit harder - git blame cannot make sense of this. But if this makes collaborating on the expr jit even a bit easier then I'm all in favor.

MasterDuke17:
I forgot to mention, but yes, NQP builds ok and passes `make m-test` and Rakudo builds ok and passes `make m-test m-spectest`.

PR #854, opened 2018-05-05
We can haz moar JIT-compiled opcodes

bdw:
Per popular request, some of the more common reasons for bailing.


lizmat:
Thank you!

PR #858, opened 2018-05-13
Moar JIT ops

bdw:
Now with a fixed `ctxlexpad`.
And I removed the `SAVE_RV` node, folding that into the call node itself.

### DIFF
--- a/src/jit/x64/emit.dasc
+++ b/src/jit/x64/emit.dasc
@@ -1796,6 +1796,25 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
         | mov WORK[dest], TMP2;
         break;
     }
+    case MVM_OP_captureexistsnamed: {
+        MVMint16 dst = ins->operands[0].reg.orig;
+        MVMint16 obj = ins->operands[1].reg.orig;
+        MVMint16 name = ins->operands[2].reg.orig;
+        | mov TMP1, WORK[obj];
+        | is_type_object TMP1;
+        | jnz >1;
+        | cmp_repr_id TMP1, TMP2, MVM_REPR_ID_MVMCallCapture;
+        | je >2;
+        |1:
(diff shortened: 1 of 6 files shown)


lizmat:
Thanks!

src/jit/x64/emit.dasc
```diff
@@ -2403,6 +2404,25 @@ void MVM_jit_emit_block_branch(MVMThreadContext *tc, MVMJitCompiler *compiler, M
             }
             break;
         }
+        case MVM_OP_lexprimspec: {
+            MVMint16 dst = ins->operands[0].reg.orig;
+            MVMint16 ctx = ins->operands[1].reg.orig;
+            MVMint16 name = ins->operands[2].reg.orig;
+            | mov TMP1, WORK[ctx];
+            | cmp_repr_id, TMP1, TMP2, MVM_REPR_ID_MVMContext;
```
niner: I wonder if checks like this could be omitted in speshed code?

bdw: Yes. Maybe. Depends.
We may have a fact that tells us that we may assume that this or that object is a concrete instance of a specific type, which would indeed allow us to circumvent this check. (It is one of the planned optimizations for the expr jit optimizer, in fact).
But, we can only *rely* on that fact if there is a guard for it, either by spesh candidate selection, or inserted during the optimization phase. The spesh optimizer does so, but the JIT compiler runs in the code-generation phase, and that makes it much harder to correctly insert a guard. I think - but I'm not 100% sure - that we can actually read from the facts whether or not they are guarded.

There's two solutions possible, as far as I can see:
- be more aggressive about asserting types for ops that the JIT can optimize for
- move JIT graph building to a earlier step in the compilation / optimization process, where it is still possible to insert guards.

I think the first is simpler to do. I haven't made up my mind about the second one.
Of course it is also possible to build a PoC for a safe subset (e.g. when we know we don't have to insert a guard).

PR #863, opened 2018-05-17
Reorder a few ops in interp.c to match oplist

jstuder-gh:
For compiler optimization purposes.

SEE: 82eb5fc722d0e05bd86fdcac9a30b56dacc6aaca

### DIFF
--- a/src/core/interp.c
+++ b/src/core/interp.c
@@ -1522,13 +1522,6 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                     GET_REG(cur_op, 2).s, GET_REG(cur_op, 4).s);
                 cur_op += 6;
                 goto NEXT;
-            OP(unicmp_s):
-                GET_REG(cur_op, 0).i64 = MVM_unicode_string_compare(tc,
-                    GET_REG(cur_op,  2).s,   GET_REG(cur_op, 4).s,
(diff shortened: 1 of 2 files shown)


jstuder-gh:
I didn't move any of the DEPRECATED ops. Having them out of order is ok?

MasterDuke17:
I recently re-ordered the JIT templates to be in the same order as interp.c. If any of the ops this commit moves are templated, could you also re-order them there?

jstuder-gh:
@MasterDuke17, they look good. There was only one template that needed to be moved; it's included in the commit.

@bdw, good point about git blame. We don't want to make searching history more difficult if it doesn't get us anything. I was thinking consistency would be good, but not at the cost of history. I'll try to profile this a bit; if it isn't any faster I'll close the PR.

jstuder-gh:
I tried profiling both the reordered version and the previous, not-reordered version.

Both were compiled with gcc using the -03 flag. When running Rakudo, I set the MVM_SPESH_DISABLE=1 env var, so that it was running pure interp.

I used the following (rather sophisticated) test script...

    use nqp;

    my str $haystack = 'This is a rather large haystack';
    my str $needle   = 'the';

    say nqp::indexicim($haystack, $needle, 0) xx 10_000_000;

with the thought that repeatedly dispatching to one of the reordered ops might yield some difference.

There was no difference. Closing.

PR #864, opened 2018-05-20
Rotate hash value randomly for each hash and on hash table expansion

samcv:
This is another improvement to prevent DOS as well as to improve our
conflict resolution. Currently a conflict of the lower `x` bits of the
hash value will cause conflicts every time we expand the table. It is
much smarter to attempt to use the full bit width of our hash value.

We define an offset for each hash table which is randomized. The hash
values are then rotated that number of bits to determine which bucket
the hash key goes into. On hash table expansion the offset is generated
randomly again.


samcv:
Yeah I believe I've implemented this separately. Going to close this PR

PR #869, opened 2018-06-03
Set UV_TCP_IPV6ONLY flag when binding async IPv6 TCP connnections

Kaiepi:
Binding without it on OpenBSD causes uv_tcp_bind to return EINVAL when
calling MVM_io_socket_listen_async with an IPv6 address.

Fixes https://github.com/rakudo/rakudo/issues/1879


Kaiepi:
Never mind, this breaks hosting on `localhost`. I'll keep looking for a proper fix

PR #876, opened 2018-06-15
Add support for all native int types to getlexref_ni and new atkey_u op

samcv:
Unlike with atkey we don't have to add additional ops because getlexref_ni
returns an object and not a native int. Make getlexref_ni accept any native
int type so we don't have to create more opcodes without reason.

### Add new atkey_u op
This is needed in nqp for the lexpad.

### Add support for all native int's in lexprimspec
This was missing the non-64 bit signed ints as well as all unsigned types.

###  Allow MVMContext atkey to work on all sizes of native ints


lib/MAST/Ops.nqp
```diff
@@ -4784,81 +4819,88 @@ BEGIN {
...
-    'ctw_check', 876,
-    'coverage_log', 877,
-    'breakpoint', 878);
+    'atkey_i8', 804,
+    'atkey_i16', 805,
+    'atkey_i32', 806,
+    'atkey_u8', 807,
+    'atkey_u16', 808,
+    'atkey_u32', 809,
+    'atkey_u64', 810,
```
jnthn: Adding a `_u` is fine, but the others should not exist. Somewhere around [here](https://github.com/MoarVM/MoarVM/blob/master/src/6model/reprs/MVMContext.c#L51) it should be possible to do a `switch` over `frame->static_info->body.lexical_types[entry->value]` for all the sized types, and accept the integer ones if `kind` is `MVM_reg_in64` and so forth. Similar will be needed for handling sized num32 too, which this PR doesn't address.

Issue #878, opened 2018-06-22
NQP no longer builds on OpenBSD post Mar 24 (bisected on NQP)


stmuk:
NQP no longer builds on OpenBSD 

Bisected on NQP to MoarVM bump of

https://github.com/perl6/nqp/commit/aa605d0ef6da958fd87f1c7342e361cf21c12d01

Error 
````
[..snip..]

perl tools/build/gen-cat.pl moar src/core/NativeTypes.nqp  src/core/NQPRoutine.nqp  src/core/NQPMu.nqp  src/core/NQPCapture.nqp  src/cor
e/IO.nqp  src/core/Regex.nqp  src/core/Hash.nqp  src/core/NQPLock.nqp  src/core/testing.nqp  src/core/YOUAREHERE.nqp   > gen/moar/stage1
/NQPCORE.setting
/home/steve/sandbox/rakudo/install/bin/moar --libpath=src/vm/moar/stage0 src/vm/moar/stage0/nqp.moarvm --bootstrap --module-path=gen/moa
r/stage1 --setting=NULL --no-regex-lib --target=mbc --stable-sc=stage1  --output=gen/moar/stage1/NQPCORE.setting.moarvm gen/moar/stage1/
NQPCORE.setting
SC not yet resolved; lookup failed
   at src/vm/moar/ModuleLoader.nqp:61  (src/vm/moar/stage0/ModuleLoader.moarvm:load_module)
 from gen/moar/stage2/NQP.nqp:270  (src/vm/moar/stage0/nqp.moarvm:load_module)
 from gen/moar/stage2/NQP.nqp:2418  (src/vm/moar/stage0/nqp.moarvm:statement_control:sym<use>)
 from gen/moar/stage2/QRegex.nqp:1656  (src/vm/moar/stage0/QRegex.moarvm:!reduce)
 from gen/moar/stage2/QRegex.nqp:1599  (src/vm/moar/stage0/QRegex.moarvm:!cursor_pass)
 from <unknown>:1  (src/vm/moar/stage0/nqp.moarvm:statement_control:sym<use>)
 from gen/moar/stage2/QRegex.nqp:1696  (src/vm/moar/stage0/QRegex.moarvm:!protoregex)
 from <unknown>:1  (src/vm/moar/stage0/nqp.moarvm:statement_control)
 from gen/moar/stage2/NQP.nqp:954  (src/vm/moar/stage0/nqp.moarvm:statement)

[..snip..]
````

[...]

PR #879, opened 2018-06-22
Use clang with OpenBSD since their ancient gcc creates a moar which can't generate NQP anymore. See #878

stmuk:


samcv:
I think this is a great plan. Going to merge. Thanks for this!

PR #880, opened 2018-06-22
Expose file descriptors of IO::Socket::INET instances

Kaiepi:
Before, `.native-descriptor` was available from IO::Socket, but always
returned -1. 

Could be useful if people want to do stuff with sockets natively

### DIFF
--- a/src/io/syncsocket.c
+++ b/src/io/syncsocket.c
@@ -3,8 +3,11 @@
 #ifdef _WIN32
     #include <winsock2.h>
     #include <ws2tcpip.h>
+    #include <io.h>
+
     typedef SOCKET Socket;
     #define sa_family_t unsigned int
+    #define isatty _isatty
 #else
     #include "unistd.h"
     #include <sys/socket.h>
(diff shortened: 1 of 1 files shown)


src/io/syncsocket.c
```diff
@@ -434,19 +434,32 @@ MVMint64 socket_getport(MVMThreadContext *tc, MVMOSHandle *h) {
     return port;
 }
 
+static MVMint64 socket_is_tty(MVMThreadContext *tc, MVMOSHandle *h) {
+    MVMIOSyncSocketData *data = (MVMIOSyncSocketData *)h->body.data;
+    return isatty(data->handle);
```
jnthn: Would a socket ever be a TTY? I think this will break the Windows build; it'll need [something like this](https://github.com/MoarVM/MoarVM/blob/master/src/io/syncfile.c#L23) doing.

Kaiepi: I'm not sure if it's actually possible or not, but I added it just to be safe

PR #882, opened 2018-06-25
Add more exprjit templates (06/24/18)

jstuder-gh:
66 new templates.

### DIFF
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -439,6 +439,14 @@
       (carg $1 ptr)
       (carg $2 ptr)) ptr_sz))
 
+(template: nfarunproto
+  (call (^func &MVM_nfa_run_proto)
+    (arglist
+      (carg (tc) ptr)
+      (carg $1 ptr)
+      (carg $2 ptr)
+      (carg $3 int)) ptr_sz))
+
 (template: null (^vmnull))
 
 (template: isnull
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -458,6 +458,15 @@
       (carg $4 ptr)
       (carg $5 ptr))))
 
+(template: radix
+  (call (^func &MVM_radix)
+    (arglist
+      (carg (tc) ptr)
+      (carg $1 int)
+      (carg $2 ptr)
+      (carg $3 int)
+      (carg $4 int)) ptr_sz))
+
 (template: null (^vmnull))
 
 (template: isnull
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -963,6 +963,12 @@
       (carg (tc) ptr)
       (carg $1 ptr)) ptr_sz))
 
+(template: getcodeobj
+  (call (^func &MVM_frame_get_code_object)
+    (arglist
+      (carg (tc) ptr)
+      (carg $1 ptr)) ptr_sz))
+
 (template: takedispatcher
   (let: (($disp     (^getf (tc) MVMThreadContext cur_dispatcher))
          ($disp_for (^getf (tc) MVMThreadContext cur_dispatcher_for))
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -423,6 +423,15 @@
       (carg $3 int)
       (carg $4 int)) int_sz))
 
+(template: findnotcclass
+  (call (^func &MVM_string_find_not_cclass)
+    (arglist
+      (carg (tc) ptr)
+      (carg $1 int)
+      (carg $2 ptr)
+      (carg $3 int)
+      (carg $4 int)) int_sz))
+
 (template: null (^vmnull))
 
 (template: isnull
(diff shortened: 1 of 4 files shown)


lizmat:
Can't wait to see these merged.  But I'm in no position to judge their correctness  :-(

src/jit/core_templates.expr
```diff
@@ -35,6 +35,51 @@
...
+(template: getlex_no
+  (let: (($found (call (^func &MVM_frame_find_lexical_by_name)
+                   (arglist
+                     (carg (tc) ptr)
+                     (carg (^cu_string $1) ptr)
+                     (carg (const (&QUOTE MVM_reg_obj) int_sz) int)) ptr_sz)))
+    (if (nz $found)
+      (do (when (nz (call (^func &MVM_spesh_log_is_logging)
+                      (arglist
+                        (carg (tc) ptr)) int_sz))
```
bdw: I see now that I should not have called that int_sz (better int64_sz). MVM_spesh_is_logging nominally returns a 32-bit int, and you're telling the JIT that it is 64 bit, which will affect the `test` operation. That will not matter in this case - anytime we assign to the `rax` register as a 32 bit operand, we clear the upper operands anyway. But it is something to be aware of.

bdw: Also, JIT-tted code should not be logging at all, so this while check can probably be removed.

src/jit/core_templates.expr
```diff
@@ -137,13 +182,79 @@
...
 
+(template: exception
+  (let: (($ah (^getf (tc) MVMThreadContext active_handlers)))
+    (if
+      (nz $ah)
+      (^getf $ah MVMActiveHandler ex_obj)
+      (^vmnull))))
+
+(template: bindexpayload!
+  (callv (^func &MVM_bind_exception_payload)
```
bdw: In this case, the  'destructive' '!' is redundant, since this is a read-only operand anyway. (It doesn't change the $0, which is why you get that as a value and not as an address). It isn't actually harmful (it has no effect at all)

src/jit/core_templates.expr
```diff
@@ -381,18 +498,122 @@
...
+
+(template: isfalse_s
+  (if
+    (nz
+      (call (^func &MVM_coerce_istrue_s)
+        (arglist
+          (carg (tc) ptr)
+          (carg $1 ptr)) int_sz))
+    (const 0 int_sz)
+    (const 1 int_sz)))
```
bdw: Sidenote: I'm thinking of making macro ^one and ^zero for this

src/jit/core_templates.expr
```diff
@@ -381,18 +498,122 @@
...
+
 (template: null (^vmnull))
 
 (template: isnull
-  (or
-    (flagval (zr $1))
-    (flagval (eq $1 (^vmnull)))))
+  (if (^is_vmnull $1) (const 1 int_sz) (const 0 int_sz)))
```
bdw: This will usually compile into bigger, branchier code (that maybe an optimizer can eliminate). But for clarity it is better.

src/jit/core_templates.expr
```diff
@@ -913,6 +1165,22 @@
...
+
+(template: hllhash
+  (^getf (^hllconfig) MVMHLLConfig slurpy_hash_type))
+
+(template: settypehll!
+  (^setf (^stable $0) MVMSTable hll_owner
+    (call (^func &MVM_hll_get_config_for)
+      (arglist
+        (carg (tc) ptr)
+        (carg $1 ptr)) ptr_sz)))
```
bdw: I'm a little bit suspicious whether this would come through type specialization... but it looks good.

src/jit/core_templates.expr
```diff
@@ -1119,9 +1452,118 @@
...
+
+(template: asynclisten
+  (call (^func &MVM_io_socket_listen_async)
+    (arglist
+      (carg (tc) ptr)
+      (carg $1   ptr)
+      (carg $2   ptr)
+      (carg $3   ptr)
+      (carg $4   int)
+      (carg $5   (&sizeof MVMint32))
```
bdw: carg's parameter is not a size but a type. this should be int.

src/jit/core_templates.expr
```diff
@@ -1119,9 +1452,118 @@
...
+                     (carg $1   ptr)
+                     (carg (const (&QUOTE MVM_reg_obj) int_sz) int)) ptr_sz)))
+    (if (nz $found)
+      (do (when (nz (call (^func &MVM_spesh_log_is_logging)
+                      (arglist
+                        (carg (tc) ptr)) int_sz))
+                (callv (^func &MVM_spesh_log_static)
+                  (arglist
+                    (carg (tc) ptr)
+                    (carg (load $found ptr_sz) ptr))))
```
bdw: Also this call can probably go away.

jstuder-gh:
@bdw, Thanks for the review! I've incorporated your suggestions and pushed up the new commits.

PR #886, opened 2018-06-27
Implement JIT templates for 12 ops (27/06/18)

Kaiepi:
Adds JIT support for the following ops:
- isinvokable
- getlexreldyn
- objprimunsigned
- scgethandle
- scobjcount
- setobjsc
- getcodename
- captureposarg_i
- captureposarg_n
- captureposarg_s
- ctxcallerskipthunks
- setwho

### DIFF
--- a/src/jit/x64/emit.dasc
+++ b/src/jit/x64/emit.dasc
@@ -873,6 +873,37 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
         | mov WORK[dst], TMP1;
         break;
     }
+    case MVM_OP_ctxcallerskipthunks: {
+        MVMint16 dst = ins->operands[0].reg.orig;
+        MVMint16 ctx = ins->operands[1].reg.orig;
+        | mov TMP5, aword WORK[ctx];
+        | test_type_object TMP5;
+        | jz >1;
+        | cmp_repr_id TMP5, TMP6, MVM_REPR_ID_MVMContext;
+        | je >1;
+        | throw_adhoc "ctxcallerskipthunks needs an MVMContext";
+        |1:
+        | mov ARG1, TC;
+        | mov ARG2, CONTEXT:TMP5->body.context->caller;
+        |2:
+        | test ARG2, ARG2;
+        | jz >4;
+        | mov TMP6, FRAME:ARG2->static_info->body.is_thunk;
+        | test TMP6, TMP6;
+        | jz >3;
+        | mov ARG2, FRAME:ARG2->caller;
+        | jmp >2;
+        |3:
+        | callp &MVM_frame_context_wrapper;
+        | mov TMP6, RV;
+        | test TMP6, TMP6;
+        | jnz >5;
+        |4:
+        | get_vmnull TMP6;
+        |5:
+        | mov WORK[dst], TMP6;
+        break;
+    }
     case MVM_OP_curcode: {
         MVMint16 dst = ins->operands[0].reg.orig;
         | mov TMP1, TC->cur_frame;
--- a/src/jit/x64/emit.dasc
+++ b/src/jit/x64/emit.dasc
@@ -103,6 +103,7 @@
 |.type CONTAINERSPEC, MVMContainerSpec
 |.type STORAGESPEC, MVMStorageSpec
 |.type HLLCONFIG, MVMHLLConfig;
+|.type SCREF, MVMSerializationContext
 |.type SCREFBODY, MVMSerializationContextBody
 |.type NFGSYNTH, MVMNFGSynthetic
 |.type CODE, MVMCode
@@ -535,6 +536,25 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
         | mov WORK[dst], TMP5;
         break;
     }
+    case MVM_OP_getlexreldyn: {
+        MVMint16 dst  = ins->operands[0].reg.orig;
+        MVMint16 ctx  = ins->operands[1].reg.orig;
+        MVMint16 name = ins->operands[2].reg.orig;
+        | mov TMP5, aword WORK[ctx];
+        | cmp_repr_id TMP5, TMP6, MVM_REPR_ID_MVMContext;
+        | je >1;
+        | test_type_object TMP5;
+        | jz >1;
+        | throw_adhoc "getlexreldyn needs a context"
+        |1:
+        | mov TMP6, CONTEXT:TMP5->body.context;
+        | mov ARG1, TC
+        | mov ARG2, aword WORK[name];
+        | mov ARG3, TMP6;
+        | callp &MVM_frame_getdynlex;
+        | mov WORK[dst], RV;
+        break;
+    }
     case MVM_OP_getlex_no:
     case MVM_OP_sp_getlex_no: {
         MVMint16  dst = ins->operands[0].reg.orig;
@@ -595,7 +615,7 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
         | mov ARG2, CONTEXT:TMP1->body.context;
         | mov ARG3, WORK[name];
         | mov ARG1, TC;
-        | callp MVM_frame_lexical_primspec;
+        | callp &MVM_frame_lexical_primspec;
         | mov WORK[dst], RV;
         break;
     }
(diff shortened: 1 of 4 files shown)


src/jit/x64/emit.dasc
```diff
@@ -1520,6 +1600,28 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
...
+        | mov TMP5, aword WORK[type];
+        | test TMP5, TMP5;
+        | jz >1;
+        | mov ARG1, TC;
+        | mov ARG2, OBJECT:TMP5->st;
+        | mov FUNCTION, STABLE:ARG2->REPR;
+        | mov FUNCTION, REPR:FUNCTION->get_storage_spec;
+        | call FUNCTION;
+        | mov TMP6, STORAGESPEC:RV->boxed_primitive;
+        | cmp TMP6, 1;
```
bdw: `test TMP6, TMP6` is shorter and simpler assembly

bdw: Matter of fact, you might be able to do:

     mov TMP6, STORAGESPEC:RV->boxed_primitive;
     and TMP6, STORAGESPEC:RV->is_unsigned;
     mov WORK[dst], TMP6;

Provided both are actually 1 when you need them (which you rely on now anyway)

src/jit/x64/emit.dasc
```diff
@@ -1534,6 +1636,48 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
...
+        MVMint16 sc  = ins->operands[1].reg.orig;
+        | mov TMP5, WORK[sc];
+        | cmp_repr_id TMP5, TMP6, MVM_REPR_ID_SCRef;
+        | je >1;
+        | throw_adhoc "Must provide an SCRef operand to setobjsc"
+        |1:
+        | mov ARG1, TC;
+        | mov ARG2, aword WORK[dst];
+        | mov ARG3, SCREF:TMP5;
+        | callp &MVM_sc_set_obj_sc
```
bdw: The semicolon that is missing here is unnecessary for DynASM, but essential for proper C syntax highlighting and formatting :smile:

src/jit/x64/emit.dasc
```diff
@@ -1650,6 +1794,24 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
         |2:
         break;
     }
+    case MVM_OP_isinvokable: {
+        MVMint16 dst = ins->operands[0].reg.orig;
+        MVMint16 src = ins->operands[1].reg.orig;
+        | mov TMP1, aword WORK[src];
+        | mov TMP1, OBJECT:TMP1->st;
+        | mov TMP2d, dword STABLE:TMP1->invoke;
+        | cmp TMP2d, dword MVM_6model_invoke_default;
```
bdw: `invoke` is a C function pointer and should be assumed to be 64 bits wide. This comparison of the lower 32 bits is a bug (that you'll probably get away with most of the time).

There's multiple correct ways of doing this; the simpler is to use `movabs` (or `mov64` in DynASM):

      mov64 TMP3, MVM_6model_invoke_default;
      cmp TMP3, aword STABLE:TMP1->invoke;

src/jit/x64/emit.dasc
```diff
@@ -1650,6 +1794,24 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
...
+        | jne >1;
+        | mov TMP3, STABLE:TMP1->invocation_spec;
+        | cmp TMP3, 0;
+        | jne >1;
+        | mov aword WORK[dst], 0;
+        | jmp >2;
+        |1:
+        | mov aword WORK[dst], 1;
+        |2:
```
bdw: Minor potential improvement; most jumps here can be eliminated by setting up a 1 (in TMP2 or something) and using AND with the interesting values.

src/jit/x64/emit.dasc
```diff
@@ -1924,6 +2086,20 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
...
+    case MVM_OP_getcodename: {
+        MVMint16 obj  = ins->operands[0].reg.orig;
+        MVMint16 code = ins->operands[1].reg.orig;
+        | mov TMP1, aword WORK[code];
+        | cmp_repr_id TMP1, TMP2, MVM_REPR_ID_MVMCode;
+        | je >1;
+        | test_type_object TMP1;
+        | jz >1;
+        | throw_adhoc "getcodename requires a concrete object";
+        |1:
```
bdw: This inverts the condition on the REPR ID, meaning that you'll skip the concreteness test for MVM_REPR_ID_MVMCode, and you'll execute it even if you know that it is an MVMCode object, which means that type-object MVMCode and concrete MVMArray instances (for example) will pass.

src/jit/x64/emit.dasc
Kaiepi: I'm not sure that'd work, `TMP6` can be the value of any of the `MVM_STORAGE_SPEC_BP` constants. src/core/interp.h specifically compares it against 1

src/jit/x64/emit.dasc
Kaiepi: egcc complained about casting a pointer to int when I tried this, but this works:
```
mov64 TMP2, ((uintptr_t)(MVM_6model_invoke_default));
cmp TMP2, STABLE:TMP1->invoke;
```

src/jit/x64/emit.dasc
bdw: Hmmm. Mabe not then (although we could always do a final `AND TMP6, 1`). It's not important :smile:

src/jit/x64/emit.dasc
bdw: This is how DynASM always does it, so it's correct

Kaiepi:
Updated. This is all the ops I was able to get working for now.

src/jit/x64/emit.dasc
```diff
@@ -535,6 +536,50 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
...
+    case MVM_OP_getlexrelcaller: {
+        MVMint16 dst  = ins->operands[0].reg.orig;
+        MVMint16 ctx  = ins->operands[1].reg.orig;
+        MVMint16 name = ins->operands[2].reg.orig;
+        | mov TMP5, aword WORK[ctx];
+        | cmp_repr_id TMP5, TMP6, MVM_REPR_ID_MVMContext;
+        | je >1;
+        | test_type_object TMP5;
+        | jz >1;
+        | throw_adhoc "getlexrelcaller needs a context";
```
bdw: Same as another comment I added, this allows both type-object MVMContext and concrete non-MVMObject, and it should not. Apologies for not spotting it earlier.

src/jit/x64/emit.dasc
```diff
@@ -535,6 +536,50 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
...
+        MVMint16 name = ins->operands[2].reg.orig;
+        | mov TMP5, aword WORK[ctx];
+        | cmp_repr_id TMP5, TMP6, MVM_REPR_ID_MVMContext;
+        | je >1;
+        | test_type_object TMP5;
+        | jz >1;
+        | throw_adhoc "getlexreldyn needs a context"
+        |1:
+        | mov TMP6, CONTEXT:TMP5->body.context;
+        | mov ARG1, TC
```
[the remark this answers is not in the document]
bdw: Oh, and repr id check is wrong here too.

src/jit/x64/emit.dasc
```diff
@@ -853,6 +898,37 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
...
         break;
     }
+    case MVM_OP_ctxcallerskipthunks: {
+        MVMint16 dst = ins->operands[0].reg.orig;
+        MVMint16 ctx = ins->operands[1].reg.orig;
+        | mov TMP5, aword WORK[ctx];
+        | test_type_object TMP5;
+        | jz >1;
+        | cmp_repr_id TMP5, TMP6, MVM_REPR_ID_MVMContext;
+        | je >1;
```
bdw: concrete repr id check

src/jit/x64/emit.dasc
```diff
@@ -921,14 +997,74 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
...
     }
+    case MVM_OP_captureposarg_i: {
+        MVMint16 dst = ins->operands[0].reg.orig;
+        MVMint16 src = ins->operands[1].reg.orig;
+        MVMint16 pos = ins->operands[2].reg.orig;
+        | mov TMP5, qword WORK[src];
+        | test_type_object TMP5;
+        | jnz >1;
+        | cmp_repr_id TMP5, TMP6, MVM_REPR_ID_MVMCallCapture;
+        | jne >1;
```
bdw: This one is correct, going to the failure condition if it is either a type object (!IS_CONCRETE()) or not-a-MVMCallCapture.

src/jit/x64/emit.dasc
```diff
@@ -1924,6 +2141,20 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
...
         break;
     }
+    case MVM_OP_getcodename: {
+        MVMint16 obj  = ins->operands[0].reg.orig;
+        MVMint16 code = ins->operands[1].reg.orig;
+        | mov TMP1, aword WORK[code];
+        | cmp_repr_id TMP1, TMP2, MVM_REPR_ID_MVMCode;
+        | je >1;
+        | test_type_object TMP1;
+        | jnz >1;
```
bdw: REPR id check here, too

Kaiepi:
Removed the template for getlexrelcaller since it was breaking things after more testing

PR #889, opened 2018-07-02
Implement SipHash, use as our hashing function w/ 64bit hashvals

samcv:
Implement SipHash, use as our hashing function w/ 64bit hashvals

This switches our hashing function from "Jenkins's Lookup2" to SipHash.

This adds a heavily modified version of https://github.com/majek/csiphash
which has been modified to use a struct to store the state and allow us to
add 64 bits at a time (two graphemes). It also has been modified to take
64 bit integers instead of pointers to bytes.

Rewriting the CSipHash code to take a 64 bit integer is significant because
it gives us a massive speedup due to much better inlining and less data
copying.

Some of this SipHash code was finished a month ago but only recently had I
optimized it enough so that was just as fast as our current hashing
function. It makes my hashing speed test file 0.5% slower, which is totally
acceptable given it has been studied extensively, while our previous hash fn
has not been studied for security to the best of my knowledge.

We now use a 128-bit hashSecret stored in two MVMuint64's. Previously we
only had a 32-bit length hashSecret (stored in a MVMuint32).

The hash values are now 64-bits instead of 32-bits.

### DIFF
--- a/src/strings/siphash/Makefile
+++ b/src/strings/siphash/Makefile
@@ -1,0 +1,4 @@
+
+all:
+	gcc -O3 test.c -g -Wall -Wextra -ggdb -o siphashtest -D MVM_CAN_UNALIGNED_INT64 && ./siphashtest
+	gcc -O3 test.c -g -Wall -Wextra -ggdb -o siphashtest && ./siphashtest
--- a/src/strings/siphash/test.c
+++ b/src/strings/siphash/test.c
@@ -1,0 +1,115 @@
+#include <stdio.h>
+#include <stdint.h>
+#include <assert.h>
+#include <sys/time.h>
+#define MVM_HASH_FORCE_LITTLE_ENDIAN 1
+#include "csiphash.h"
+uint64_t gettime_ns() {
+	struct timeval tv;
+	gettimeofday(&tv, NULL);
+	return (uint64_t)tv.tv_sec * 1000000000ULL + tv.tv_usec * 1000ULL;
+}
+
+#define REPEATS 20000
+
+uint64_t vectors[64] = {
+	0x726fdb47dd0e0e31LLU, 0x74f839c593dc67fdLLU, 0x0d6c8009d9a94f5aLLU, 0x85676696d7fb7e2dLLU,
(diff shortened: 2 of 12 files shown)


src/strings/siphash/Makefile
```diff
@@ -0,0 +1,4 @@
+
+all:
+	gcc -O3 test.c -g -Wall -Wextra -ggdb -o siphashtest -D MVM_CAN_UNALIGNED_INT64 && ./siphashtest
+	gcc -O3 test.c -g -Wall -Wextra -ggdb -o siphashtest && ./siphashtest
```
jnthn: Should this have been committed?

src/strings/siphash/test.c
```diff
@@ -0,0 +1,90 @@
...
+			assert(siphash24(plaintext, i, (uint64_t*)key) == vectors[i]);
+		}
+	}
+	assert(siphash24((uint8_t*)Grapheme32, 9 * sizeof(int32_t), (uint64_t*)key) == 4563223716124497198LLU);
+	testmvm();
+	t1 = gettime_ns();
+
+	printf("%i tests passed in %.3fms, %.0fns per test\n", REPEATS*65, (t1-t0)/1000000., (t1-t0)/(REPEATS*64.));
+	return 0;
+}
```
jnthn: Was it intended to include this in the commit?

jnthn:
The [Appveyor failure](https://ci.appveyor.com/project/samcv/moarvm-35u74/build/1.0.856/job/mcrgosve7odja28p) looks like a legit problem also.

Really nice work otherwise. :-)

src/strings/siphash/test.c
samcv: Yes, this was meant to be commited so we still get to keep the tests for the csiphash.h file and make sure that if we make changes to it that everything is still working correctly.

jnthn: OK, good with me. I do wonder if we want to put them under a `t/` directory or some such, but it's not a big deal (unless we start accumulating such things around the repo, and then it is :-)). :+1:

samcv:
Merging this now.

PR #891, opened 2018-07-03
Alter some exprjit templates to remove branching

jstuder-gh:
Remove branching from some templates where a bitwise op will do.

### DIFF
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -593,14 +593,12 @@
       (carg $1 ptr)) int_sz))
 
 (template: isfalse_s
-  (if
-    (nz
+  (flagval
+    (zr
       (call (^func &MVM_coerce_istrue_s)
         (arglist
           (carg (tc) ptr)
-          (carg $1 ptr)) int_sz))
-    (const 0 int_sz)
-    (const 1 int_sz)))
+          (carg $1 ptr)) int_sz))))
 
 (template: null (^vmnull))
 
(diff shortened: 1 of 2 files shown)


src/jit/core_templates.expr
```diff
@@ -640,11 +638,9 @@
...
 
 (template: isconcrete
-  (if (all
-        (nz $1)
-        (^is_conc_obj $1))
-    (const 1 int_sz)
-    (const 0 int_sz)))
+  (and
+    (flagval (nz $1))
+    (flagval (^is_conc_obj $1))))
```
bdw: This one relied on the 'short-circuit' behaviour of `all` - after the first `nz $1`, the `^is_conc_obj` (which derferences the pointer) isn't executed. The bitwise-version skips that, so this one isn't safe.

src/jit/core_templates.expr
```diff
@@ -1154,39 +1150,29 @@
...
 
 (template: isint
-  (if (all
-        (nz $1)
-        (eq (^getf (^repr $1) MVMREPROps ID) (const (&QUOTE MVM_REPR_ID_P6int) (&sizeof MVMuint32))))
-    (const 1 int_sz)
-    (const 0 int_sz)))
+  (and
+    (flagval (nz $1))
+    (flagval (eq (^getf (^repr $1) MVMREPROps ID) (const (&QUOTE MVM_REPR_ID_P6int) (&sizeof MVMuint32))))))
```
bdw: Same here.

jstuder-gh:
Darn, I should have noticed that glaring issue. @bdw, I reworked the commit and re-pushed to include the (few) safe changes.

PR #900, opened 2018-07-08
Add more exprjit templates (07/08/18)

jstuder-gh:
Add 59 more templates and nullptr macro.

They've  passed the spectest with the following env vars set:

* MVM_SPESH_NODELAY=1
* MVM_SPESH_INLINE_DISABLE=1
* MVM_SPESH_OSR_DISABLE=1

### DIFF
--- a/src/core/interp.c
+++ b/src/core/interp.c
@@ -3591,7 +3591,7 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                 goto NEXT;
             OP(bind_sk):
                 MVM_io_bind(tc, GET_REG(cur_op, 0).o,
-                    GET_REG(cur_op, 2).s, GET_REG(cur_op, 4).i64, (int)GET_REG(cur_op, 6).i64);
+                    GET_REG(cur_op, 2).s, GET_REG(cur_op, 4).i64, (MVMint32)GET_REG(cur_op, 6).i64);
                 cur_op += 8;
                 goto NEXT;
             OP(accept_sk):
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -1460,6 +1460,31 @@
           (carg $1   ptr)
           (carg $2   ptr)) int_sz))))
 
+(template: pow_I
+  (call (^func &MVM_bigint_pow)
+    (arglist
+      (carg (tc) ptr)
+      (carg $1   ptr)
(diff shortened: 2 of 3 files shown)


lizmat:
Thank you for all the hard work.  But I think these should be added as separate Pull Requests, so they can be judged separately, and bisected better in case of problems.

Now, it is an all-or-nothing Pull Request, which gets scarier to commit with each addition.

jstuder-gh:
I understand and appreciate your concern. I think that you're right that it would be better to submit in smaller, digestable chunks than with one monolithic PR. In the future, I'll definitely submit more along those lines.

I do want to mention that I have been working on this branch with bisectability in mind. Each commit only has a few related templates included at a time (if they are only calling a C function) or only one if it is a bit more involved.

I run [this script](https://gist.github.com/jstuder-gh/9e8b5736f2087221a315ed322d67ce53) as I add new commits and use it with git bisect to track down issues when they appear. This script runs the spectest and sets MVM_SPESH_NODELAY so that the jitting will occur right away (as well as MVM_SPESH_INLINE_DISABLE and MVM_SPESH_OSR_DISABLE so that there is less optimization going on other than the jitting). 

This branch passes for me when I run this with [commit 2ff4274d6482eafadce42c06b9763b6d084ab277](https://github.com/perl6/roast/commit/2ff4274d6482eafadce42c06b9763b6d084ab277) checked out for the roast (it's a bit older, but so is the head of MoarVM's master that I branched off from). I'm also using [commit cec76ff79ebb4092803d5e680e648883bda1bc55](https://github.com/perl6/nqp/commit/cec76ff79ebb4092803d5e680e648883bda1bc55) on NQP and [commit ade83c861484f5f4e4184a88d9dae79a2cdf110c](https://github.com/rakudo/rakudo/commit/ade83c861484f5f4e4184a88d9dae79a2cdf110c) on Rakudo.

This script is assuming that all the source directories have the same base directory and that MoarVM has already had Configure.pl run for it. Also I tend to run it with a local branch other than master checked out on ${basedir}/rakudo/t/spec so that the spectest won't be updated (there's probably a better way but it works).

I won't add any more commits to this branch. If need be I can break this one up too.

lizmat:
Wrt this branch, I'll leave that up to the people deciding the correctness of it.  :-)

jstuder-gh:
Closing (was merged in commit 691aa0d648486143f970b907e9a900a32e47c67d).

PR #901, opened 2018-07-08
Make a faster HASH_ITER macro and make other hash optimizations

samcv:
HASH_ITER_FAST is faster and it doesn't randomize bucket order (though
it's faster regardless of that).

Makes the biggest difference when we are iterating a hash with a LOT of
keys in it, such as adding a million keys to a hash then deleting and
adding back again. Got about 40% speed improvement on this nqp script:
https://gist.github.com/9a65a586ea21b20def8efe4aaaab55de

Also optimize GET_X_BITS by making a switch out of it. It doesn't seem
like this would be faster, but it is due to compiler optimizations.

### DIFF
--- a/src/strings/uthash.h
+++ b/src/strings/uthash.h
@@ -124,9 +124,9 @@ do {                                                                            
  * If the size of the hashv is changed we will need to change max_hashv_div_phi,
  * to be max_hashv / phi rounded to the nearest *odd* number.
  * max_hashv / phi = 2654435769 */
-const static uint32_t max_hashv_div_phi = 2654435769LLU;
+const static uint32_t max_hashv_div_phi = MVM_C_CONSTANT_U64(2654435769);
 #define DETERMINE_BUCKET_FIB(hashv, offset) \
-    (((hashv) * max_hashv_div_phi) >> (32 - offset))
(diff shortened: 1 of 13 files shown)


src/strings/uthash.h
```diff
@@ -556,20 +557,95 @@ do {
...
-
-/* obtain a count of items in the hash */
-#define HASH_CNT(hh,head) ((head)?((head)->hh.tbl->num_items):0)
 #define GET_X_BITS(bucket_rand, log2_num_buckets) \
     ((bucket_rand) >> ((sizeof(bucket_rand) * 8) - (log2_num_buckets)))
+/* obtain a count of items in the hash */
+#define HASH_CNT(hh,head) ((head)?((head)->hh.tbl->num_items):0)
+
+MVM_STATIC_INLINE unsigned GET_X_BITS_FUNCT(MVMuint64 bucket_rand, unsigned log2_num_buckets) {
+    switch (log2_num_buckets) {
```
AlexDaniel: > Also optimize GET_X_BITS by making a switch out of it. It doesn't seem
like this would be faster, but it is due to compiler optimizations.

Maybe it needs a comment here saying that it is done this way because it's faster? Otherwise someone may refactor it months later :)

PR #912, opened 2018-07-20
Add +x and shebangs for scripts.

jkramer:
Same cleanup as for rakudo and nqp repos, see rakudo/rakudo#2077.


AlexDaniel:
Thanks!

PR #914, opened 2018-07-20
Update dyncall to latest commit

Kaiepi:
Forgot to do this when I updated MoarVM/dyncall since I was using `--has-dyncall` to build MoarVM instead


AlexDaniel:
I think dyncall was updated some time ago, so there's no need in this PR. Correct me if I'm wrong.

PR #915, opened 2018-07-21
JIT neg_I and abs_I

MasterDuke17:
Rewrite the MVM_BIGINT_UNARY_OP macro the same way as was done for the
MVM_BIGINT_BINARY_OP macro so it can be JITted.

### DIFF
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -1441,14 +1441,14 @@
     (arglist
       (carg (tc) ptr)
       (carg $2   ptr)
-      (carg $1   ptr)) int_sz))
+      (carg $1   ptr)) ptr_sz))
 
 (template: abs_I
   (call (^func &MVM_bigint_abs)
     (arglist
       (carg (tc) ptr)
       (carg $2   ptr)
-      (carg $1   ptr)) int_sz))
+      (carg $1   ptr)) ptr_sz))
 
 (template: cmp_I
   (call (^func &MVM_bigint_cmp)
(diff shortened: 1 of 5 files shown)


src/jit/core_templates.expr
```diff
@@ -1436,6 +1436,20 @@
       (carg $1   ptr)
       (carg $2   ptr)) ptr_sz))
 
+(template: neg_I
+  (call (^func &MVM_bigint_neg)
+    (arglist
+      (carg (tc) ptr)
+      (carg $2   ptr)
+      (carg $1   ptr)) int_sz))
```
jstuder-gh: One small thing, looks like you accidently put a return size of int_sz instead of ptr_sz.

I don't think it makes a difference here since the size of a pointer and MVMint64 are the same on x64, but I suppose it could be an issue on other potentially JITted architectures where that is not the case.

src/jit/core_templates.expr
```diff
@@ -1436,6 +1436,20 @@
...
+
+(template: abs_I
+  (call (^func &MVM_bigint_abs)
+    (arglist
+      (carg (tc) ptr)
+      (carg $2   ptr)
+      (carg $1   ptr)) int_sz))
```
jstuder-gh: Same as above.

MasterDuke17:
Good catch.

src/math/bigintops.h
```diff
@@ -1,7 +1,7 @@
 int MVM_bigint_mp_set_uint64(mp_int * a, MVMuint64 b);
 
-void MVM_bigint_abs(MVMThreadContext *tc, MVMObject *result, MVMObject *a);
-void MVM_bigint_neg(MVMThreadContext *tc, MVMObject *result, MVMObject *a);
+MVMObject * MVM_bigint_abs(MVMThreadContext *tc, MVMObject *result, MVMObject *a);
```
timo: rename result to result_type?

PR #918, opened 2018-07-22
Allow getting file descriptors from async sockets

Kaiepi:
This is so Rakudo can provide them as properties on clients and servers
created through IO::Socket::Async.

### DIFF
--- a/src/io/asyncsocket.c
+++ b/src/io/asyncsocket.c
@@ -497,6 +497,15 @@ static void push_name_and_port(MVMThreadContext *tc, struct sockaddr_storage *na
     MVM_repr_push_o(tc, arr, port_o);
 }
 
+static void push_handle_fd(MVMThreadContext *tc, MVMObject *arr, uv_handle_t *handle) {
+    uv_os_fd_t fd;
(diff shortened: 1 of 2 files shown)


src/io/asyncsocket.c
```diff
@@ -715,8 +732,10 @@ static void on_connection(uv_stream_t *server, int status) {
             MVM_repr_push_o(tc, arr, msg_box);
             MVM_repr_push_o(tc, arr, tc->instance->boot_types.BOOTStr);
             MVM_repr_push_o(tc, arr, tc->instance->boot_types.BOOTInt);
+            MVM_repr_push_o(tc, arr, tc->instance->boot_types.BOOTInt);
```
jnthn: Why this addition?

Kaiepi: It's there as a placeholder for the listening socket's file descriptor from line 714

Having to wait until someone connects before its value can be known probably isn't the most ideal way of handling this, though

Kaiepi:
Closing this since the way I worked out how to get the server's native descriptor as soon as it gets tapped really needs to be two separate pull requests

PR #922, opened 2018-07-27
Add missing .h suffix to build/Makefile.in

### LINKED ISSUE #917: Build fails on NetBSD with default config
[jit/core_templates.h can't be found by the compiler](https://gist.github.com/Kaiepi/a70d442fbc4ca6d19071dc687ed836be) when building with make. Using gmake instead stops it from happening, but that's not the default setting and it doesn't come installed with the OS. I tested on FreeBSD and OpenBSD and didn't run into this issue.

- OS
NetBSD nbsd.vm.kennel.qt 8.0 NetBSD 8.0 (GENERIC) #0: Tue Jul 17 14:59:51 UTC 2018  mkrepro@mkrepro.NetBSD.org:/usr/src/sys/arch/amd64/compile/GENERIC amd64
- Version
2018.06-309-g428ed0e2a

Kaiepi:
This was making compiling the JIT fail on NetBSD since
src/jit/core_templates.h was never getting compiled.

Fixes #917

### DIFF
--- a/build/Makefile.in
+++ b/build/Makefile.in
@@ -489,7 +489,7 @@ DYNASM  = $(MINILUA) 3rdparty/dynasm/dynasm.lua
 DYNASM_SCRIPTS = 3rdparty/dynasm/dynasm.lua 3rdparty/dynasm/dasm_x86.lua
 DYNASM_HEADERS = 3rdparty/dynasm/dasm_proto.h 3rdparty/dynasm/dasm_x86.h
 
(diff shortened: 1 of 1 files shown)


jnthn:
Well, seems the automated checks are happy with this, which means it doesn't make any of those other `make` programs we exercise there *unhappy*. I figure 9 days is enough for somebody more make-savvy to have weighed in on the PR if there were going to, so will merge. Thanks!

PR #926, opened 2018-08-04
JIT decoder(empty|takebytes)

MasterDuke17:

This gets rid of some BAILs I saw when using Compress::Zlib. However, it seemed to actually make it slightly slower (roughly from 6.4s before to 6.6s after when `gzslurp`ing a 975k file). @bdw any idea why?

### DIFF
--- a/src/jit/graph.c
+++ b/src/jit/graph.c
@@ -235,6 +235,8 @@ static void * op_to_func(MVMThreadContext *tc, MVMint16 opcode) {
     case MVM_OP_encode: return MVM_string_encode_to_buf;
     case MVM_OP_decoderaddbytes: return MVM_decoder_add_bytes;
     case MVM_OP_decodertakeline: return MVM_decoder_take_line;
+    case MVM_OP_decodertakebytes: return MVM_decoder_take_bytes;
+    case MVM_OP_decoderempty: return MVM_decoder_empty;
 
     case MVM_OP_elems: return MVM_repr_elems;
     case MVM_OP_concat_s: return MVM_string_concatenate;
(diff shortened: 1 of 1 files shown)


timo:
not sure if the impact can be this big, but i'd assume that ensure_decoder would be inlined in the non-jit version, and is a "full" call in this one.

timo:
either make a c function that's only meant to be called by the jit that has the `ensure` function at its start, or implement it in the exprjit as a template (i.e. port the `ensure` function to expr template language)

src/jit/graph.c
```diff
@@ -2769,6 +2771,22 @@ static MVMint32 consume_ins(MVMThreadContext *tc, MVMJitGraph *jg,
...
         break;
     }
+    case MVM_OP_decodertakebytes: {
+        MVMint16 dst     = ins->operands[0].reg.orig;
+        MVMint16 decoder = ins->operands[1].reg.orig;
+        MVMint16 chomp   = ins->operands[2].reg.orig;
+        MVMint16 inc     = ins->operands[3].reg.orig;
+        MVMJitCallArg argc[] = { { MVM_JIT_INTERP_VAR, { MVM_JIT_INTERP_TC } },
+                                 { MVM_JIT_REG_VAL, { decoder } },
+                                 { MVM_JIT_LITERAL_PTR, { (MVMint64)"decodertakebytes" } } };
```
bdw: pls make this uintptr_t, we have enough unportable stuff going on

MasterDuke17: I just copied this from the other `decoder*` cases. Should I change them all?

bdw: Yes please :smile:

PR #929, opened 2018-08-05
Refactor Expression JIT to make tree manipulation easier

bdw:
This introduces a number of internal changes in the expression JIT. The purpose of these changes is to make the internal representation easier to manipulate, which should help the development of the expression optimizer.

- Assigning labels for branches now happens during tiling process, which allows us to assign absolute labels (since the compiler is already constructed at tiling time)
- We add a new function `MVM_jit_expr_apply_template_adhoc` which unifies adhoc tree manipulation with template application
- Small constants (up to 32 bits) and large constants (64 bits) are now treated separately. This simplifies assembly language of tiles using constants, and allows us to reduce the node size to 32 bit.
- `MVMJitExprNode` is no longer a separate type (although we named the enum of operators `MVMJitExprOperator`. 
- We got rid of the separate `MVMJitExprNodeInfo` parallel-tree, which needed to be maintained separately during tree manipulation, was massively wasteful (none of the nodes corresponding to arguments or links were filled in) and cache-unfriendly. 
- The same information`MVMJitExprInfo` node which is stored inline (next to the operator node; both are now 32 bits in size). As an additional result, there is no longer any difference in handling variadic versus nonvariadic nodes, which should make tree processing a bit faster as well.
- The JIT sequence number starts at 1 so the `jit-bisect.pl` tool can bisect the first compiled frame.
- Cast mode and operand type information has been removed from `expr_ops.h` into the node analysis and tree precompilation code, respectively, the only places where they are useful.
- The expression template validator has stopped destructively modifying the template array (it used to insert the variadic argument length, but this now happens during template emitting)


docs/jit/todo.org
```diff
@@ -243,8 +243,105 @@ Currently, we try to resovle the link address directly. What we should
...
+
+ And the pointer itself replaced by the index.
+
+ We could argue against dealing with 64 bit constants in general, but
+ unfortunately, const_i64 prevents us from doing that.... Ways of
+ dealing with that:
+
+ + A 'large constants' table per tree (into which we could copy both the
+   i64 constants and the function pointer constants)
+   + We could store this entire table in the data section, too
```
jnthn: Is https://github.com/MoarVM/MoarVM/pull/929/files#diff-ba7685d88421b4854627c06ca18b94b8R152 not doing this?

jnthn:
Looks like a very nice set of improvements. :+1:

docs/jit/todo.org
bdw: Yes. Documentation is out of date. (The `todo.org` was an experiment to try and see if there'd be collaboration or value out of a public roadmap, but it hasn't really worked out that way).

PR #932, opened 2018-08-07
add speshguardgetstaticcode, for closures and such

timo:
lets a spesh plugin figure out if a given object,
such as the $!do attribute of a Code, is "the same"
across invocations - ignoring what exactly it closes
over.

i implemented this so i could make the "assign" spesh
plugin in rakudo understand Proxy objects.


bdw:
@timo - is this still valuable, and if so, can you resolve the conflict above?

PR #934, opened 2018-08-07
Get the number of bytes to be processed from the current buffer and not from the header.

xelak6:
This fixes issue #2158.


MasterDuke17:
https://github.com/rakudo/rakudo/issues/2158

PR #935, opened 2018-08-08
Silence an incorrect GCC string overflow warning

MasterDuke17:
The string in question is created by MVM_string_utf8_c8_encode_C_string
for the only call of this function, so we can be assured it does in fact
have a null terminator and using strlen is safe.


niner:
Why use strncpy when you allocate the buffer based on temp_size already?

MasterDuke17:
> Why use strncpy when you allocate the buffer based on temp_size already?

@nine I'm just attempting to silence the warning, I didn't write the original code, so I couldn't say. But I guess it could be changed to just a strdup()?

PR #936, opened 2018-08-08
Increase the result buffer size.

xelak6:
The last grapheme of the previous input buffer is inserted in the
current result buffer. Therefore, the result buffer needs space
for an additional grapheme when processing the last input buffer.


zoffixznet:
:+1: Thanks!

PR #941, opened 2018-08-15
Fix Spesh Memory Leaks

bdw:
Spesh memory management is mostly simple except for the cases where it isn't.
We create spesh graphs in three ways:

- From a spesh plan (i.e. the original bytecode)
- From a spesh candidate, when inlining
- From a known invocant, when we do not have a pre-existing specialization, when inlining.

When we generate a spesh candidate, we transfer ownership of allocated memory (e.g. for inlines, handlers, local type maps, spesh slots) to that candidate.

I found that we had a number of leaks coming from the third case, and that they were not straightforward to fix. The code assumed that inlinees came from spesh candidates and that the spesh candidate would therefore own the allocated memory, and not free it. But in the third case there is no spesh candidate, so those fields would leak.

I fixed this in two ways:
- I made sure the inlinee graph was destroyed directly after inlining. The memory allocated by the region allocator (for the graph nodes that are merged into the inliner) is merged into the inliner region allocator.
- I made sure we free all allocated fields, if and only if they are different from the ones allocated for the spesh candidate, so that we don't free those that are still in use.

Further, there is a fix for the FSA, so that the per-thread FSA freelist is cleaned up before freeing the memory of the FSA allocator as a whole.

### DIFF
--- a/src/core/regionalloc.c
+++ b/src/core/regionalloc.c
@@ -43,3 +43,15 @@ void MVM_region_destroy(MVMThreadContext *tc, MVMRegionAlloc *alloc) {
     }
     alloc->block = NULL;
 }
+
+/* Link source region into target region, so they can be cleaned up as one */
+void MVM_region_merge(MVMThreadContext *tc, MVMRegionAlloc *target, MVMRegionAlloc *source) {
+    MVMRegionBlock *block = source->block;
+    while (block != NULL) {
+        MVMRegionBlock *prev = block->prev;
+        block->prev = target->block->prev;
+        target->block->prev = block;
+        block = prev;
+    }
+    source->block = NULL;
+}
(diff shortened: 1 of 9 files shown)


coke:
+1 for post-release merge.

jnthn:
This looks good, but I agree it's probably most sensible to hold off until after the release, given it's not so trivial a change that I can say "this can't possibly be wrong". :-)

I have a couple of fixes that are in the same boat, fwiw: they fix deopt bugs that could theoretically occur on master (even if I didn't ever observe that happening), but are involved enough to carry a small risk of bringing their own issue. I'm happy to lean towards "devil we know" for both those fixes and this PR.

bdw:
Thank you for the review. I will merge after the release.

Maybe for the future, it is better to have an (explicitly stable) pre-release branch than to lock development on master for an indefinite time.

AlexDaniel:
@bdw there is a post-release branch https://github.com/MoarVM/MoarVM/tree/postrelease-opts and you can merge this PR there.

bdw:
That is literally the opposite of what I meant :smile: 
Either way, there is clear need for a branch that is 'open for development', and one that is stable for release.
So whether we call the one master and the other release, or the first 'postrelease' or 'development', and the second 'master', matters not very much, but we should probably do something for it.

PR #942, opened 2018-08-18
Improve exception msg for slice op on VMArray

jstuder-gh:


src/6model/reprs/VMArray.c
```diff
@@ -863,10 +863,13 @@ static void aslice(MVMThreadContext *tc, MVMSTable *st, MVMObject *src, void *da
...
     MVMint64 elems;
 
+    if (total_elems <= 0) MVM_exception_throw_adhoc(tc, "MVMArray: Cannot slice an empty list");
+
     start = start < 0 ? total_elems + start : start;
     end   = end   < 0 ? total_elems + end   : end;
     if ( end < start || start < 0 || end < 0 || total_elems <= start || total_elems <= end ) {
-        MVM_exception_throw_adhoc(tc, "MVMArray: Slice index out of bounds");
+        MVM_exception_throw_adhoc(tc, "MVMArray: Slice index out of bounds\n"
+            "\tStart: %ld\n" "\tEnd: %ld", start, end);
```
jnthn: Is it true that you can't slice an empty list? What if (as a degenerate case) you tried to take an empty slice of an empty list?

jstuder-gh: Slicing an empty list works fine in Perl 6's interfaces, but doesn't currently work via this op. The op's current interface takes the beginning and ending index of a list:

```perl6
my @l := nqp::list(1, 2, 3);
for nqp::slice(@l, 0, 0) { say($_) }

# OUTPUT: 1
```

```perl6
my @l := nqp::list();
for nqp::slice(@l, 0, 0) { say($_) }

# OUTPUT: MVMArray: Slice index out of bounds
```
Since any index would be out of bounds given an empty list, I figured it would make sense to have the error state it explicitly.

I suppose the op could correct for out-of-bounds indices, but I had figured that it made sense to have the onus be on the HLL to verify the data is correct rather than have the VM cover for bad data. And if the data is really horribly mangled it will fail faster.

What do you think?

PR #945, opened 2018-08-21
Workaround version failure on Travis CI (MacOS)

jstuder-gh:
Travis doesn't fetch the full history (only 50 commits by default) and
we use `git fetch --unshallow` to retrieve the rest. This is necessary
to retrieve the tags used in versioning MoarVM. That process does not
appear to be working on MacOS currently.

Explicitly fetch the full history in order to work around this defect.

See [this thread](https://github.com/travis-ci/travis-ci/issues/4942)
for more information.

I don't really know if this works or not without submitting the PR and
letting Travis do it's thing :P


jnthn:
Thanks!

PR #946, opened 2018-08-21
A few more exprjit templates (08/14/18)

jstuder-gh:
+9 new templates

### DIFF
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -1823,11 +1823,24 @@
       (carg $1   ptr)
       (carg $2   ptr)) ptr_sz))
 
+(template: getpid
+  (call (^func &MVM_proc_getpid)
+    (arglist
+      (carg (tc) ptr)) int_sz))
+
 (template: assertparamcheck!
   (when (zr $0)
     (callv (^func &MVM_args_bind_failed)
       (arglist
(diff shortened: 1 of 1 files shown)


src/jit/core_templates.expr
```diff
@@ -987,6 +987,46 @@
...
+
+(template: unshift_o!
+  (callv (^getf (^repr $0) MVMREPROps pos_funcs.unshift)
+    (arglist
+      (carg (tc) ptr)
+      (carg (^stable $0) ptr)
+      (carg $0 ptr)
+      (carg (^body $0) ptr)
+      (carg $1 ptr)
+      (carg (const (&QUOTE MVM_reg_obj) int_sz) int))))
```
bdw: I don't think the `!` means, what you think it means, and given that I've implemented it, I'm probably right :wink: 
The `!` means that the template writes a value to a register (has a write operand), but doesn't the expression doesn't yield a value but writes directly to the output register. It means that the template is *destructive*, hence the `!` was chosen for similarity with the scheme convention.
Since none of the unshift operators actually have an output operand, the `!` is wrong here, and after I've merged #950 it will error out, too.

jstuder-gh: Thanks. I've updated the ops to remove the '!'.

PR #947, opened 2018-08-21
Add decodelocaltime op

pmurias:
Will help us remove nqp::p6decodelocaltime.
Not tested on windows as I don't have a dev env set up on it :(


src/platform/win32/time.c
```diff
@@ -41,3 +42,18 @@ void MVM_platform_nanosleep(MVMuint64 nanos)
         millis = (DWORD)((end - now) / E6);
     }
 }
+
+void MVM_platform_decodelocaltime(MVMint64 time, MVMint64 decoded[]) {
+    const time_t t = (time_t)time;
+    struct tm *tm = localtime(&t, &tm);
```
bdw: I'd prefer the use of `locatime_s` here, which is equivalent to `localtime_r` in the POSIX world

pmurias: How portable across windows versions is localtime_s? I'm not a windows dev so I have no idea

bdw: From my (limited) understanding, exactly as portable as `localtime` because it is part of the 'Universal C Runtime Library'.

pmurias:
I fixed the localtime_s use. I don't use windows so I haven't tested the fix :(

PR #948, opened 2018-08-22
Update libuv to 1.23.0

ugexe:
Many windows bugfixes -- notably "win,pipe: fix IPC pipe deadlock"

https://github.com/libuv/libuv/blob/90891b4232e91dbd7a2e2077e4d23d16a374b41d/ChangeLog#L3-L255


jnthn:
Will merge this after the upcoming release, so it gets proper testing.

PR #949, opened 2018-08-24
Cleanup *_I ops

MasterDuke17:
Make the ops in interp.c more consistent, and convert `expmod_I` to allocate in the function like the other `*_I` ops do.

NQP builds ok and passe `make m-test` and Rakudo builds ok and passes `make m-test m-spectest`.


MasterDuke17:
Closing this in favor of the cleaner https://github.com/MoarVM/MoarVM/pull/970.

PR #950, opened 2018-08-31
[JIT] New template compiler

bdw:
- Single-pass compilation (no separate 'validate_template' step
- Better type and format checking (operands should be expressions or references, parameters should be barewords, numbers or macros)
- Write-out parameters need to be refered with a `\$` sigil, like `\$0`, and templates without output operands should not get a '!' suffix

Already caught a few issues this way.

### DIFF
--- a/docs/jit/how-to-add-an-missing-expression-operator.org
+++ b/docs/jit/how-to-add-an-missing-expression-operator.org
@@ -13,7 +13,7 @@ The file =src/jit/expr_ops.h= contains a macro-list of all defined operators.
 Each operator is defined as such:
 
 #+BEGIN_EXAMPLE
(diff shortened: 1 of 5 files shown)


bdw:
Also added hygienic macros by eager resolution of `let:` declarations, e.g. the following ought to work, even though it is nonsense:

    (macro: ^foo (,bar) (let: (($obj (load ,bar ptr_sz))) (addr $obj ,bar))
    (template: foobar (let: (($obj (load $1 ptr_sz))) (^foo $obj))

I.e. the `$obj` in the macro `^foo` refers only to the declaration within the macro, not to the declaration in the template; and the `,bar` substitution in the application of `^foo` correctly refers to the declaration of `$obj` in the outer `let:`.

PR #953, opened 2018-09-05
Add a macro for decoder_ensure_decoder and ...

MasterDuke17:
templates for `decoder<addbytes takebytes takeline empty>`.

### DIFF
--- a/src/jit/macro.expr
+++ b/src/jit/macro.expr
@@ -89,3 +89,13 @@
 (macro: ^body (,a) (addr ,a (&offsetof MVMObjectStooge data)))
 
 (macro: ^nullptr () (const 0 ptr_sz))
+
+(macro: ^decoder_ensure_decoder (,tc ,decoder ,op)
+    (when (any
+           (ne (^getf (^repr ,decoder) MVMREPROps ID) (const (&QUOTE MVM_REPR_ID_Decoder) (&sizeof MVMuint32)))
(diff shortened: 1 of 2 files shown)


src/jit/macro.expr
```diff
@@ -89,3 +89,13 @@
...
+
+(macro: ^decoder_ensure_decoder (,tc ,decoder ,op)
+    (when (any
+           (ne (^getf (^repr ,decoder) MVMREPROps ID) (const (&QUOTE MVM_REPR_ID_Decoder) (&sizeof MVMuint32)))
+           (zr (flagval (^is_conc_obj ,decoder))))
+         (callv (^func &MVM_exception_throw_adhoc)
+           (arglist
+             (carg ,tc ptr)
+             (carg (const_ptr "Operation '%s' can only work on an object with the Decoder representation") ptr)
+             (carg ,op ptr)))))
```
bdw: Ideally, throw_adhoc is a macro by itself.. and maybe repr-id is also a macro.

bdw: A C macro that concatenates strings :smile:  - I'm trying it out now.

niner:
Shouldn't spesh be able to ensure that the prerequisites are actually met, so we don't even need the code to throw an exception in JITed code?

bdw:
Yes, in many cases it ought to... and so repr id check elimination is on the roadmap for the expression optimizer. I'm not quite sure how I'd do it, though....

PR #954, opened 2018-09-05
Implement 'cast_load_addr' tile for exprjit 

jstuder-gh:
Implement the 'cast_load_addr' tile for the exprjit template.
Logic mirrors that of the standard cast operation.

Also adds a template that uses the new tile and a macro for storage_spec.

Compiles and passes spectest, but definitely could use review. ping @bdw.

### DIFF
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -694,7 +694,7 @@
     (nz $1)
     (cast
       (^getf (^storage_spec $1) MVMStorageSpec boxed_primitive)
-      int_sz (&sizeof MVMuint16) (&QUOTE MVM_JIT_CAST_UNSIGNED))
+      int_sz (&sizeof MVMuint16) cast_unsigned)
     (const 0 int_sz)))
 
 (template: gethow
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -689,6 +689,14 @@
       (carg $2 ptr)
       (carg \$0 ptr))))
 
+(template: objprimspec
+  (if
+    (nz $1)
+    (cast
+      (^getf (^storage_spec $1) MVMStorageSpec boxed_primitive)
+      int_sz (&sizeof MVMuint16) (&QUOTE MVM_JIT_CAST_UNSIGNED))
+    (const 0 int_sz)))
+
 (template: gethow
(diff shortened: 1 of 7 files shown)


src/jit/x64/tiles.dasc
```diff
@@ -263,8 +263,68 @@ MVM_JIT_TILE_DECL(cast) {
...
 
+/* Logic is the same as cast above except loading from memory instead of reg */
+/* See comments in cast above for more detail */
 MVM_JIT_TILE_DECL(cast_load_addr) {
-    DIE("NYI");
+    MVMint32 ofs       = tile->args[0];
+    MVMint32 load_size = tile->args[1];
+    MVMint32 to_size   = tile->args[2];
+    MVMint32 from_size = tile->args[3];
+    MVMint32 cast_mode = tile->args[4];
```
bdw: Ouch. The size of `args` is limited to 4. I guess that's not enough then.

bdw: ... although you don't strictly need both from_size and load_size, so maybe there is a fix for this

src/jit/core_templates.expr
```diff
@@ -689,6 +689,14 @@
       (carg $2 ptr)
       (carg \$0 ptr))))
 
+(template: objprimspec
+  (if
+    (nz $1)
+    (cast
+      (^getf (^storage_spec $1) MVMStorageSpec boxed_primitive)
+      int_sz (&sizeof MVMuint16) (&QUOTE MVM_JIT_CAST_UNSIGNED))
```
bdw: `(&QUOTE MVM_JIT_CAST_UNSIGNED)` can be written as `cast_unsigned` (the `MVM_JIT_` is an implied prefix); the `(&QUOTE)` macro expression is intended for when this isn't suitable.

src/jit/x64/tiles.dasc
jstuder-gh: If we set the args array size to something higher (6 might be a good number?), would that interfere with any code expecting a certain size or alignment of the structure?

bdw: Yes, that is actually the easiest and cheapest fix.

jstuder-gh:
@bdw, thanks for the review! I've added another commit extending the size of the args array to 6.

PR #964, opened 2018-09-15
Fix wrong field in exprjit cast template and add macro

jstuder-gh:
In the 'objprimbits' template, accidently used the wrong field name for
cast's from_size value ('boxed_primitive' instead of 'bits').

In this case, 'boxed_primitive' and 'bits' are both the same size, but
that doesn't make it less wrong just less painfully wrong 😉.

Added a macro that makes this idiom more DRY and less prone to this type
of error.

### DIFF
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -748,9 +748,7 @@
 (template: objprimspec
   (if
     (nz $1)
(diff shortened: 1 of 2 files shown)


bdw:
Awesome :+1:

PR #967, opened 2018-09-16
Safe fork in MoarVM

bdw:
The `fork()` system call poses problems for multithreaded programs such as MoarVM, on account that there is only one thread that is inherited by the child process, and that the other threads in the parent process might be modifying heap memory (e.g. processing `malloc`) while the `fork()` call occured. So having multiple threads while forking might leave the child process in an inconsistent state.
Fortunately, we have some degree of control over the threads started by the VM itself, meaning we can try to stop them (and temporarily become a single-threaded process) prior to the call to `fork()`, and reinstall them afterwards.

As a nice side effect, this also means that we can shut down the spesh and IO threads before exit (in `--full_cleanup` mode), which means that this crashes only sometimes rather than always.

This means, though, that the state used by the system threads (IO eventloop and spesh worker) must become attached to the VM instance, and that we must be able to send effective stop signals. The eventloop thread provides a convenient means by the `uv_loop_stop` call and a `uv_async_send` wakeup. The spesh worker will now recognize VMNull on the work queue as a stop signal. To ensure that this signal is processed speedily, I implemented an `unshift` operation on the MVMConcBlockingQueue REPR.

Other changes are:
- the IO eventloop now uses a per-MVMInstance uv_loop_t structure, rather than the default per-threadcontext structure. Since we now only use the per-thread loop for synchronous file system operations, we might get rid of it entirely.
  - I had to fix several bugs in clients of the eventloop to use the correct (passed in) uv_loop_t structure, rather than the redundant one assigned to the threadcontext.
-  The entire MVMConcBlockingQueueBody structure is now malloc'd, rather than jsut the locks sections; this significantly simplifies code.
- The spesh worker thread is now published as an MVMInstance variable. No attempt is made to protect it, but it is started when the process is single-threaded (or holding a mutex)

Whether this has more practical value than hack value is open for debate :smile: - but at least the non-crashing full-cleanup should be welcome.


ugexe:
Does this avoid the need for http://docs.libuv.org/en/v1.x/loop.html#c.uv_loop_fork ?

bdw:
Actually, that's a good question. I think that is still necessary, yes.

PR #972, opened 2018-09-27
Remove JIT log file

bdw:
Using the new spesh graph comments, let's get rid of the adhoc JIT log.


lizmat:
So, how would one find out which blocks are not getting JITted?

jnthn:
@lizmat The information is now folded into the specialization log.

PR #973, opened 2018-09-28
Remove som dead code.

patzim:
The two #ifndef branches were equal. Just remove the #ifndef.


jnthn:
Nice catch, thanks! Will leave the CI to run before merging.

PR #976, opened 2018-09-30
Exprjit Templates (09/20/18)

jstuder-gh:
+12 templates


bdw:
Merged in as 0939a10cdd

PR #979, opened 2018-10-02
Better takedispatcher opt

timo:
takedispatcher was already turned into a null instruction, but nothing downstream was taking advantage of this.

quoth jonathan at http://colabti.org/irclogger/irclogger_log/moarvm?date=2018-10-02#l50

> jnthn: Uh, yeah, you can't optimize much inside inlines at the moment :)
> jnthn: Because there's no information on what is retained for deopt available
> jnthn: So the chances of screwing things up is very high.
> timotimo: i'll create a PR for the changes, i suppose?
> jnthn: Yeah, but I'm going to reject it on principle in the immediate.
> jnthn: Or well, defer it until we have the things needed to do it safely.
> […]
> jnthn: fwiw, I think the better approah on this will be to do a normal optimize_bb pass on the inlinee - when we're got to the point where we can do it safely

### DIFF
--- a/src/spesh/inline.c
+++ b/src/spesh/inline.c
@@ -492,7 +492,10 @@ MVMSpeshBB * merge_graph(MVMThreadContext *tc, MVMSpeshGraph *inliner,
                 /* If the inlining code doesn't set a dispatcher, takedispatcher
                  * in the inlinee can be assumed to always return null. */
                 if (!inliner->sets_dispatcher && !inlinee->sets_dispatcher) {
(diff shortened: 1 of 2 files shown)


timo:
i've found this to cause some random spec test crashes, fwiw.

bdw:
@timo do you still want this merged in? Or shall I close the branch?

jnthn:
We now retain deopt info and run the optimizer over the inlined code, which rips the takedispatcher branch out fully. That's a more general solution (covers more opts on the inlinee). So I'd say it supersedes this. Thanks, though.

Issue #981, opened 2018-10-06
Probable Issue with exprjit bit shifting tiles


jstuder-gh:
The left and right bit shift tiles I submitted recently appear to be incorrect. I mistakenly used the ```reg[1]``` as an immediate value in the shift instruction, but somehow overlooked that that would not be using the contents of the register in the operation.

```
     | sal Rq(reg[0]), reg[1];
```

Before submitting, I had run spectests (which include many bit shifting operations) with MVM_SPESH_NODELAY on, so I thought that it was working as intended. However, it appears that it is just never consistently not being invoked due to the exprjit bailing before then. As such it is not currently affecting results but could in the future as templates are built out more.

I've made a commit that I believe could work ([see here](https://github.com/jstuder-gh/MoarVM/commit/151413dc1b663bbd70c7cd04115aa50af7d0c2d4)), but I don't really know of any way to test it. Plus it is likely it could fail in some instances (I've included a comment on the commit)

I believe the safest thing to do be to remove the tiles and templates altogether for now, until sometime in the future when more of the surrounding templates are in place and it is easier to test for. What do you think?

Ping @bdw.

bdw:
Oops! I'm an idiot for not seeing it during review. But I'm not sure why you're implementing the move to cl in the patch.

jstuder-gh:
I tried  ```sal Rq(reg[0]) Rq(reg[1])``` but unfortunately the instruction will not accept two registers as operands (with the exception of CL) (http://www.felixcloutier.com/x86/SAL:SAR:SHL:SHR.html).

bdw:
Replicated with this test, thanks for the report: https://gist.github.com/bdw/f1c4db2a029c8e601025d0caa23ade0c

PR #982, opened 2018-10-06
Remove exprjit shift operations for now

jstuder-gh:
See [Github Issue #981](https://github.com/MoarVM/MoarVM/issues/981)


bdw:
You're quite correct. Because shl/shr don't take operand other than from `cl` (hysterical raisins?), we can't implement this correctly without implementing register requirements.

bdw:
Replication that demonstrates the problem: https://gist.github.com/bdw/f1c4db2a029c8e601025d0caa23ade0c

PR #985, opened 2018-10-19
Use larger buffer for gethostname when available

ugexe:


ugexe:
Fixes the following:
```
$ sudo hostname abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz01234567
$ hostname
abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz01234567
$ perl6 -e 'say gethostname()'
abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqrstuvwxyz01
```

timo:
i'm not sure if it was exactly that, but in the past i've had trouble with MSVC, uint8_t foo[123]; and sizeof(foo). we'll need to double-check if it works on appveyor and jnthn's machines, and perhaps use the constant again?

jnthn:
AppVeyor seems happy, so I figure it's good on Windows too. Thanks!

PR #986, opened 2018-10-19
Comment out problematic case in speshing deconting of _n

zoffixznet:
This appears to fix R#2395 https://github.com/rakudo/rakudo/issues/2395 but I've no idea why.

I've absolutely no idea about this area of the codebase or the impact of this change.

I'm making this PR for last-resort type of situation, if we have to fix up R#2395 for the release, but I hope someone who knows this part of the code will figure out what's going on.


zoffixznet:
I removed the blocker label (for [reasons](https://github.com/rakudo/rakudo/issues/2395#issuecomment-431848737) from the ticket so I don't think this PR is needed anymore.

PR #989, opened 2018-11-01
 Add exprjit integer multiplication op

jstuder-gh:
+1 template

Only adding a tile that takes two registers for now, as that is the only
way the op is being used currently.

### DIFF
--- a/src/jit/x64/tiles.dasc
+++ b/src/jit/x64/tiles.dasc
@@ -567,6 +567,13 @@ MVM_JIT_TILE_DECL(and_load_idx) {
     }
 }
 
+MVM_JIT_TILE_DECL(mul_reg) {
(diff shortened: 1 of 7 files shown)


src/jit/x64/tiles.dasc
```diff
@@ -567,6 +567,13 @@ MVM_JIT_TILE_DECL(and_load_idx) {
     }
 }
 
+MVM_JIT_TILE_DECL(mul_reg) {
+    MVMint8 reg[2];
+    ensure_two_operand_pre(tc, compiler, tile, reg);
+    | imul Rq(reg[0]), Rq(reg[1]);
```
jstuder-gh: Two dynamic registers this time, I swear :P

PR #994, opened 2018-11-09
Make nqp::serializetobuffer return a buffer instead of string

pmurias:
This fixes the case where the serialized data is not directly stashed into the bytecode


pmurias:
That should fix it

PR #1001, opened 2018-11-26
Include system library paths in Configure.pl if configured to use them

### LINKED ISSUE #1000: Configure.pl fails when including system libraries with clang
```
bastille% ./Configure.pl --has-libtommath --cc=clang 
Welcome to MoarVM!

Updating submodules .................................... OK
Configuring native build environment ................... 
    trying to compile a simple C program ............... YES
    Testing pkgconfig .................................. 0.27.1
OK

    probing whether your compiler thinks that it is gcc  Can't compile simple gcc probe, so something is badly wrong at build/probe.pm line 147.
```

The same error happens with `--has-dyncall`, `--has-libuv`, etc. I'm not sure if this happens or not on Linux, I tested on OpenBSD 6.4 with the GENERIC.MP kernel

Kaiepi:
Fixes #1000

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -212,20 +212,12 @@ if (-e "$config{pkgconfig}") {
 
 # conditionally set include dirs and install rules
 $config{cincludes} = '' unless defined $config{cincludes};
+$config{lincludes} = '' unless defined $config{lincludes};
 $config{install}   = '' unless defined $config{install};
 if ($args{'has-libuv'}) {
     $defaults{-thirdparty}->{uv} = undef;
     unshift @{$config{usrlibs}}, 'uv';
-    if ($config{pkgconfig_works}) {
-        my $result = `$config{pkgconfig} --cflags libuv`;
-        if ( $? == 0 ) {
-            $result =~ s/\n/ /g;
-            $config{cincludes} .= ' ' . "$result";
-            print("Adding extra include for libuv: $result\n");
-        } else {
-            print("Error occured when running $config{pkgconfig} --cflags libuv.\n");
-        }
-    }
+    setup_native_library('libuv') if $config{pkgconfig_works};
 }
 else {
     $config{cincludes} .= ' ' . $defaults{ccinc} . '3rdparty/libuv/include'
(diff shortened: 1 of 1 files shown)


Configure.pl
```diff
@@ -851,6 +834,35 @@ sub wsl_bash_on_win {
...
+        }
+    }
+    else {
+        print("Error occured when running $config{pkgconfig} --cflags $library.\n");
+    }
+    my $result_libs = `$config{pkgconfig} --libs-only-L $library`;
+    if ( $? == 0 ) {
+        $result_libs =~ s/\n//g;
+        if (index($config{lincludes}, $result_libs) == -1) {
+            $config{lincludes} .= ' ' if $config{lincludes};
```
samcv: This is better written as `$config{lincludes} = join(' ', $config{lincludes}, $result_libs);`

Configure.pl
```diff
@@ -328,6 +302,14 @@ sub uniq {
     $defaults{-thirdparty}->{dcb} = undef;
     $defaults{-thirdparty}->{dl}  = undef;
     $config{nativecall_backend} = 'dyncall';
+    if (index($config{cincludes}, '-I/usr/local/include') == -1) {
+        $config{cincludes} .= ' ' if $config{cincludes};
```
samcv: Same as mentioned below about using a join instead of concatenation.

Configure.pl
```diff
@@ -328,6 +302,14 @@ sub uniq {
...
+        $config{cincludes} .= '-I/usr/local/include';
+    }
+    if (index($config{lincludes}, '-L/usr/local/lib') == -1) {
+        $config{lincludes} .= ' ' if $config{lincludes};
```
samcv: Same as mentioned below about using a join instead of concatenation.

Configure.pl
```diff
@@ -851,6 +834,35 @@ sub wsl_bash_on_win {
...
 }
 
+# set up cincludes and ldusr flags for a native library
+sub setup_native_libraries {
+    my $library = shift;
+    my $result_cflags = `$config{pkgconfig} --cflags $library`;
+    if ( $? == 0 ) {
+        $result_cflags =~ s/\n//g;
+        if (index($config{cincludes}, $result_cflags) == -1) {
+            $config{cincludes} .= ' ' if $config{cincludes};
```
samcv: Same as mentioned below about using a join instead of concatenation.

PR #1008, opened 2018-12-05
Fix double free in platform/sys.c

Kaiepi:
libuv expects the info struct to have already been initialized.

See https://github.com/rakudo/rakudo/issues/2520

### DIFF
--- a/src/platform/sys.c
+++ b/src/platform/sys.c
@@ -4,9 +4,10 @@
 MVMuint32 MVM_platform_cpu_count(void) {
     int count;
     uv_cpu_info_t *info;
+    int e;
 
-    uv_cpu_info(&info, &count);
-    uv_free_cpu_info(info, count);
+    e = uv_cpu_info(&info, &count);
+    if (e == 0) uv_free_cpu_info(info, count);
 
     return count;
 }


niner:
I don't think your assessment is correct. libuv docs claim: "int uv_cpu_info(uv_cpu_info_t** cpu_infos, int* count) Gets information about the CPUs on the system. The cpu_infos array will have count elements and needs to be freed with uv_free_cpu_info()."

So we're actually not talking about an uv_cpu_info* but a whole array of those pointers. Which means that initializing with a single one is an odd solution already. And indeed the source shows that an array is allocated by uv_free_cpu_info():
https://github.com/libuv/libuv/blob/7ebb26225f2eaae6db22f4ef34ce76fa16ff89ec/src/unix/openbsd.c#L265

I think what happens is that the sysctl in line 262 fails and uv_cpu_info returns errno before actually allocating the array. That's why the uv_free_cpu_info() call explodes. I think a more correct fix is to check uv_cpu_info's return value and only call uv_free_cpu_info() if count > 0.

cygx:
From what I can gather, the correct solution would be to check the return value of `uv_cpu_info()` and only call `uv_free_cpu_info()` on success (ie return value 0). On failure, the former will clean up after itself.

cygx:
This of course poses a follow-up question: What value should `MVM_platform_cpu_count()` if `uv_cpu_info()` fails?

Kaiepi:
Maybe an exception should be thrown if that happens?

niner:
Note that uv_cpu_info() can also return errors after it allocated the cpu_infos array but it will set count right after successfully allocating that array. So really, we must only call uv_free_cpu_info() if count is > 0

cygx:
Cf eg [line 274](https://github.com/libuv/libuv/blob/7ebb26225f2eaae6db22f4ef34ce76fa16ff89ec/src/unix/openbsd.c#L274) or [line 285](https://github.com/libuv/libuv/blob/7ebb26225f2eaae6db22f4ef34ce76fa16ff89ec/src/unix/openbsd.c#L85): Here, the memory gets freed after count has already been set. So under certain conditions, using count as condition variable would result in a double free.

PR #1018, opened 2018-12-11
keep end of string bytes \x0000

### LINKED ISSUE #1017: utf16 encoding does not have end of string bytes
This is related to https://stackoverflow.com/questions/53691576/perl6-nativecall-with-str-is-encodedutf16-got-randomly-corrupted-result

xlat:
Fixes #1017
This is also related to https://stackoverflow.com/questions/53691576/perl6-nativecall-with-str-is-encodedutf16-got-randomly-corrupted-result

This fix add the missing space for the end of string char which is 2 bytes for utf16.

### DIFF
--- a/src/strings/utf16.c
+++ b/src/strings/utf16.c
@@ -400,7 +400,7 @@ char * MVM_string_utf16_encode_substr_main(MVMThreadContext *tc, MVMString *str,
     if (!output_size)
         output_size = &scratch_space;
     *output_size = (char *)result_pos - (char *)result;
-    result = MVM_realloc(result, *output_size);
+    result = MVM_realloc(result, *output_size + 2);
     MVM_free(repl_bytes);
     return (char *)result;
 }
--- a/src/strings/utf16.c
+++ b/src/strings/utf16.c
@@ -397,6 +397,7 @@ char * MVM_string_utf16_encode_substr_main(MVMThreadContext *tc, MVMString *str,
         }
     }
     result_pos[0] = 0;
+    result_pos[1] = 0;
     if (!output_size)
         output_size = &scratch_space;
     *output_size = (char *)result_pos - (char *)result;


src/strings/utf16.c
```diff
@@ -400,7 +400,7 @@ char * MVM_string_utf16_encode_substr_main(MVMThreadContext *tc, MVMString *str,
     if (!output_size)
         output_size = &scratch_space;
     *output_size = (char *)result_pos - (char *)result;
-    result = MVM_realloc(result, *output_size);
+    result = MVM_realloc(result, *output_size + 2);
```
jnthn: There's no certainty that the 2 bytes here will be zeroed, which I think is important?

xlat: Yes it is important as a single \x00 byte is for an utf8/ansi c-style string; that's function is supposed to return a valid c-style string utf16, isn't it ?
My local tests does not got corrupted results, I may be lucky 😅. 
So we should add a `result_pos[1] = 0;` too ?

cygx: The original pull-request was fine: As `result_pos` is of type `MVMuint16*`, the assignment [on line 399](https://github.com/MoarVM/MoarVM/blob/master/src/strings/utf16.c#L399) cleared two bytes.

xlat: So now how to drop commit 21e3601 from this PR ?

jnthn: Urgh, I see. This is rather confusing code: it looked like it was doing `realloc` to make space for the extra null bytes, and so should then be zeroing those out afterwards. Whereas it seems that the `realloc` here is just to try and make the string consume less memory after the encoding, and I guess the `while ((alloc_size - 2 * (result_pos - result)) < bytes_needed)` is trying to keep space for the extra 2 bytes at the end. So yes, I think the original PR was right after all.

@xlat Just `git reset --hard HEAD~1` and `git push -f`

jnthn:
Cherry-picked the desired commit into `master`; thanks!

samcv:
Do we need a test for this?

PR #1033, opened 2019-01-12
Trim array returned by nqp::decodelocaltime

pmurias:
Remove elements that are unused by rakudo.
This makes it work the same as on other backends.


lizmat:
FWIW, I'd rather see the other fields documented and supported on other platforms.  Is that difficult for some reason?  Also, I would like to see a nqp::gmtime() function as well, but that's another issue I guess.  :-)

pmurias:
Determining if the daylight saving time is in effect won't be trivial to do correctly on the js backend (but I guess not impossible), the other two fields don't seem to be a problem. 
If we actually need them they can be needed but if they won't I don't see a reason to keep useless and untested code lying around.

PR #1037, opened 2019-01-20
Add uname op

ugexe:
* Updates libuv to 1.25.0 for `uv_os_uname`.
* Adds `uname` op.

This would allow rakudo to avoid calling out to `bin/uname`:
https://github.com/rakudo/rakudo/blob/a5b9fc2de5eac6734d700724fe46b2c3ab0d82fe/src/core/Kernel.pm6#L13
https://github.com/rakudo/rakudo/blob/e9351cbaa961c4d1e4ef8b4cb52418d55766a8d6/src/core/Distro.pm6#L61

Similar libuv functionality should find its way into nodejs in the near future. I'm not familiar with a JVM solution, but I can't imagine it would be too difficult. We can continue to use `bin/uname` as a fallback either way.


src/platform/sys.c
```diff
@@ -11,3 +11,21 @@ MVMuint32 MVM_platform_cpu_count(void) {
...
+    int r = uv_os_uname(&uname);
+    if (r != 0)
+        MVM_exception_throw_adhoc(tc, "Unable to uname: %s", uv_strerror(r));
+    if (REPR(result)->ID != MVM_REPR_ID_VMArray || !IS_CONCRETE(result) ||
+            ((MVMArrayREPRData *)STABLE(result)->REPR_data)->slot_type != MVM_ARRAY_STR) {
+        MVM_exception_throw_adhoc(tc, "uname needs a concrete string array.");
+    }
+
+    MVMROOT(tc, result, {
+        MVM_repr_bind_pos_s(tc, result, 0, MVM_string_utf8_decode(tc, tc->instance->VMString, uname.sysname, strlen((char *)uname.sysname)));
```
jnthn: This has a subtle problem: we root the local `result`, but if we're passing it as an argument then it will be potentially read and placed on the stack or into a register. Then the `decode` can allocate, and leave the register/stack location that `result` was earlier read into outdated, leading to problems. So the two calls need to be split over two lines, with a temporary.

src/core/interp.c
```diff
@@ -5543,6 +5543,10 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                 cur_op += 8;
                 goto NEXT;
             }
+            OP(uname):
+                MVM_platform_uname(tc, GET_REG(cur_op, 0).o);
+                cur_op += 2;
+                goto NEXT;
```
jnthn: I think I'd find this a bit more natural if we allocated a `BOOTStrArray`, populated it, and set it as the result, rather than relying on one being passed in; I don't think we'd really need the caller to pick the type here.

ugexe: Understood. I was cribbing from `getrusage`... there is a [commit](https://github.com/MoarVM/MoarVM/commit/fdb5e4d6bb15a14caa27bb35db8d751d4c24d131#diff-bb3587478c6b68f62c75192b00b05f0fR4543) that changes to passing in an array, so I cargo culted that through to this.

jnthn: That's a fair point. That particular op can be called repeatedly (as is by the scheduler), and I guess this was done to avoid regular allocations; I'm not that fond of the exception, but it's pragmatic. I don't think the information from `uname` changes, so there'd be little reason to repeatedly call it, so we can do things the usual way.

ugexe:
Closing in favor of https://github.com/MoarVM/MoarVM/pull/1049

PR #1038, opened 2019-01-25
Add smrt intify op

MasterDuke17:
This will be needed for an about-to-land NQP PR that makes NQP default to using int instead of num behavior for variables and ops.


src/core/interp.c
```diff
@@ -6376,7 +6385,7 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
             OP(DEPRECATED_31):
                 MVM_exception_throw_adhoc(tc, "The syncpipe op was removed in MoarVM 2017.07.");
             OP(DEPRECATED_32):
-                MVM_exception_throw_adhoc(tc, "The close_fhi op was removed in MoarVM 2017.07.");
+                MVM_exception_throw_adhoc(tc, "The foobar op was removed in MoarVM 2017.07.");
```
jnthn: This change looks accidental.

MasterDuke17: Woops, force pushed a fix.

src/core/coerce.c
```diff
@@ -378,6 +378,42 @@ void MVM_coerce_smart_numify(MVMThreadContext *tc, MVMObject *obj, MVMRegister *
         else
             MVM_exception_throw_adhoc(tc, "cannot numify this");
     }
+    {
+        MVMint64 t = (MVMint64)res_reg->n64;
+        if ((MVMnum64)t != res_reg->n64) {
+            //fprintf(stderr, "got an actual non-int value %g\n", res_reg->n64);
+        }
+    }
```
jnthn: This block looks like a dead code debug leftover; I must have missed it in the first review, sorry.

MasterDuke17: Argh, can't believe I committed that. Anyway, force pushed another update.

MasterDuke17:
Rebased to HEAD on master.

src/spesh/optimize.c
```diff
@@ -929,19 +929,21 @@ static void optimize_repr_op(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpeshBB
 /* smrt_strify and smrt_numify can turn into unboxes, but at least
```
niner: You could mention smrt_intify here

MasterDuke17: Yeah, updated a couple other comments also and force pushed.

PR #1042, opened 2019-01-28
Allow CStruct pass-by-value in Nativecall

scovit:
Second version of the patch, @bdw ++

This feature can be implemented only on libffi.

This is the initial implementation of the pass-by-value
functionality for nativecall, it recognizes the 'copy' key on
parameter/return hashes. Then it build up the libffi parameter
definition by analyzing the STables and REPRs of the members. It
currently works only on P6num P6int and CStruct member types but
will be extended upon review. This functionality can be implemented
only on the libffi nativecall backend, and might render dyncall
obsolete for MoarVM.

### DIFF
--- a/src/6model/bootstrap.c
+++ b/src/6model/bootstrap.c
@@ -543,6 +543,7 @@ static void string_consts(MVMThreadContext *tc) {
     string_creator(attribute, "attribute");
     string_creator(of, "of");
     string_creator(rw, "rw");
+    string_creator(copy, "copy");
     string_creator(type, "type");
     string_creator(typeobj, "typeobj");
     string_creator(free_str, "free_str");
--- a/src/6model/reprs/CStruct.c
+++ b/src/6model/reprs/CStruct.c
@@ -588,7 +588,15 @@ static void bind_attribute(MVMThreadContext *tc, MVMSTable *st, MVMObject *root,
                         if (REPR(value)->ID != MVM_REPR_ID_MVMCStruct)
                             MVM_exception_throw_adhoc(tc,
                                 "Can only store CStruct attribute in CStruct slot in CStruct");
-                        cobj = ((MVMCStruct *)value)->body.cstruct;
+                        if (repr_data->attribute_locations[slot] & MVM_CSTRUCT_ATTR_INLINED) {
+                            memcpy((char *)body->cstruct + repr_data->struct_offsets[slot],
+                                   ((MVMCStruct *)value)->body.cstruct,
+                                   ((MVMCStructREPRData *)STABLE(value)->REPR_data)->struct_size);
+                            ((MVMCStruct *)value)->body.cstruct = (char *)body->cstruct + repr_data->struct_offsets[slot];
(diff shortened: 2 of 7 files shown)


bdw:
Hi @scovit  - why did you close this? I haven't had the opportunity to look at it yet 😄

scovit:
Hi @bdw ,
Thank you for the interest in my patch, I am working on MoarVM with that applied locally, so I am keeping it updated.
 
It is here: https://github.com/scovit/MoarVM/tree/cstruct-by-value
And here is the associated simple rakudo patch: https://github.com/scovit/rakudo/tree/cstruct-by-value
 
When I did a rebase, the pull-request was automatically closed. I can open another one, if it helps. By the way I didn’t understand if it is a big issue the fact that this feature will be libffi only?

I honestly do not understand why we keep both libffi and dyncall as slow-path backends, on top of the JIT integration?

niner:
@scovit NativeCall is an important part of Perl 6 and Rakudo and everything improving our coverage is very welcome!
I, too was confused by this PR being closed before I had a chance to do a review. Didn't know that Github will do that automatically.

I'm not sure if the "is copy" trait is the most appropriate here. For other types the "rw" trait expresses something getting passed as pointer and it's absence signifies passing by value. Of course, we are already a bit late since CStructs are passed as pointer by default and an "rw" trait doesn't mean anything. The same is actually true for arrays. That may be something to clean up in a NativeCall v2. The important bit is that MoarVM will have to support both variants.

As for libffi vs. dyncall I guess that's just because no one has asked that question before. I'm not even aware of the reasons for having two backends in the first place. In theory we could get rid of both if someone completes the JIT support. The biggest obstace for that is support for callbacks.

scovit:
Hi @niner, thank you for reviewing my code, I think we can keep the discussion open here!

I am not sure about the "rw" reasoning part: if I understand correctly, in Perl6, objects are always references. For this reason, I think that defaulting to pass-by-value for CStruct is ambiguous and should be avoided, while that is totally fine for native types. Instead, adding another trait, such as "is copy", will match better the way Perl 6 treats objects in arguments passing for Perl 6 functions.

scovit:
Regarding the point that "objects are always references", I honestly find very weird the uppercase HAS syntax for Array/CStruct inlining inside a CStruct. I think that a specific trait, such as "is inlined", could be more consistent with to the "is rw" trait specific to declaring accessorized attributes.

patzim:
IIRC the reason for having both libffi and dyncall was that libffi doesn't support Windows without mingw/cygwin and dyncall misses many less used platforms. Thus support for both was added.

scovit:
@patzim it is my understanding, from https://github.com/libffi/libffi , that libffi can compile with MSVC and requires mingw/cygwin only as part of the build process

jnthn:
> In theory we could get rid of both if someone completes the JIT support.

Only if they do the JIT support for every platform that we might wish to support. :-)

bdw:
So much work though!

PR #1043, opened 2019-02-02
Jit (open|read|close)_dir

MasterDuke17:
These were causing some `JIT: bailed completely because of <...>` in the code from gfldex++'s blog post.


MasterDuke17:
@timo good catch, updated.

PR #1045, opened 2019-02-04
Configure.pl: fix cross-compilation

ffontaine:
Do not add -I/usr/local/include or -L/usr/local/lib as this will break
cross-compilation.

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -260,11 +260,13 @@ else {
 if ($args{'has-libtommath'}) {
     $defaults{-thirdparty}->{tom} = undef;
     unshift @{$config{usrlibs}}, 'tommath';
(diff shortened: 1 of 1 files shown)


Kaiepi:
`-I/usr/local/include` and `-L/usr/local/lib` are added because ld fails to find the libraries on *BSD without them. Only adding those flags if it's known that we're not cross-compiling might work better

Kaiepi:
That or checking if the OS is a *BSD before adding them. I'm not sure if Solaris would also need this or not

ffontaine:
I updated the PR to disable these paths when cross-compiling

PR #1049, opened 2019-02-12
Add uname op

ugexe:
This would allow rakudo to avoid calling out to `bin/uname`:
https://github.com/rakudo/rakudo/blob/a5b9fc2de5eac6734d700724fe46b2c3ab0d82fe/src/core/Kernel.pm6#L13
https://github.com/rakudo/rakudo/blob/e9351cbaa961c4d1e4ef8b4cb52418d55766a8d6/src/core/Distro.pm6#L61

Similar libuv functionality should find its way into nodejs in the near future. I'm not familiar with a JVM solution, but I can't imagine it would be too difficult. We can continue to use `bin/uname` as a fallback either way.


ugexe:
Closing/opening to rebuild

ugexe:
This requires a re-run of `tools/update_ops.p6` to resolve conflicts. Since there is another op related PR I will wait on re-running it (or whoever merges this can re-run it themselves).

PR #1053, opened 2019-02-13
Relocatable MoarVM

patzim:
This PR makes moar search for libmoar relative to its location by default, essentially making MoarVM relocatable. A Configure option --no-relocatable switches back to the old behavior.
Also MVM_exepath() is added to the public API. That function is required by the new perl6 runners in rakudo.
AIX very likely still needs work.
This PR is mostly about getting feedback from other devs. I do not expect this PR (and the respective PRs in NQP and Rakudo) to be merged right away.
I tested this on Linux+GCC, Windows10+MinGW and Windows10+VC.


build/setup.pm
```diff
@@ -132,10 +132,11 @@ our %TC_POSIX = (
...
 
-    ccshared   => '-fPIC',
-    ldshared   => '-shared @ccshared@',
-    moarshared => '',
-    ldrpath    => '-Wl,-rpath,"/@libdir@"',
+    ccshared            => '-fPIC',
+    ldshared            => '-shared @ccshared@',
+    moarshared          => '',
+    ldrpath             => '-Wl,-rpath,"/@libdir@"',
+    ldrpath_relocatable => '-Wl,-z,origin,-rpath,\'$$ORIGIN/../lib\'',
```
manchicken: I don't think that the `-z` flag to `ld` is terribly portable. It looks like the macOS linker is not a fan of that flag.

patzim:
This branch is now confirmed to also work on OpenBSD and MacOS. I know of no other blockers so I'm fine with this PR (and the respectives in NQP and rakudo) to be merged post release.
I can do a rebase on head before the merge if that is in any way helpful.

Issue #334, opened 2016-01-14
Build failure on Mac OS X: file was built for archive which is not the architecture being linked (x86_64)


briandfoy:
I'm trying to build Moar from the latest on GitHub and the download on MoarVM.org. I keep getting an error about third-party libraries:

ld: warning: ld: warning: ld: warning: ignoring file 3rdparty/libatomic_ops/src/libatomic_ops.a, file was built for archive which is not the architecture being linked (x86_64): 3rdparty/libatomic_ops/src/libatomic_ops.aignoring file 3rdparty/tinymt/libtinymt.a, file was built for archive which is not the architecture being linked (x86_64): 3rdparty/tinymt/libtinymt.aignoring file 3rdparty/sha1/libsha1.a, file was built for archive which is not the architecture being linked (x86_64): 3rdparty/sha1/libsha1.a

Let me know what you'd like to know about my system; it's a basic Mac OS X setup.

FROGGS:
Hi, can we get the entire build output, starting with the output of Configure.pl?

coke:
Also output of ```sw_vers``` and any options passed to Configure.pl (just did a build here on 10.10.5 with no issues)

raiph:
geekosaur's comments from http://irclog.perlgeek.de/moarvm/2016-01-16#i_11896132

you might run lipo -detailed_info against the archive

to see what it says

but: there si a case macports has seen where someone had the wrong dev tools on their system and ended up with archives that had the correct objects in them but no symbol table. ld erroneously reports this as "which is not the architecture being linked"

if lipo confirms that the archive has the right arch objects in it, see if "otool -S" produces output.

if it doesn't, then look for extra (incorrect) versions of ar, ranlib, and/or ld. (also check that Configure.pl is finding the ones it should be) (the specific case we saw was someone had installed the iOS dev kit incorrectly and some things were finding the ARM cross-dev versions of ar and ranlib)

...actually just ranlib was the wrong one

but you can also see this if building 64-bit objects and then somehow have a 32-bit-only ranlib or ar sitting around; ar will build valid archives without a symbol table, ranlib won't add a symbol table

...ld will report the wrong thing

and that 32-bit thing can happen pretty easily if this is a system upgraded from Tiger or Leopard because xcode 4 and later do not remove earlier 32-bit xcode that lived under /Developer

and random stuff will use the old /Developer 32-bit versions :/

[...]

PR #1055, opened 2019-02-17
Adding some safeties for macOS

manchicken:
After spending more than two days chasing down issues with MoarVM and rakudobrew, I found a simple problem with the Homebrew binutils package being incompatible with the Xcode toolchain.

I added a quick check to make sure that those on macOS know what they're getting themselves in to.


manchicken:
Much credit for help in troubleshooting to @jnthn and this issue here: https://github.com/MoarVM/MoarVM/issues/334

jnthn:
Thanks!

PR #1054, opened 2019-02-13
use _wopen on windows

apapas:

PR #1057, opened 2019-02-23
Use _wopen on windows via uv_fs_open

ugexe:
Inspired by https://github.com/MoarVM/MoarVM/pull/1054 this uses uv_fs_open instead of open(2) (which uses _wopen/CreateFileW on Windows)

### DIFF
--- a/src/io/syncfile.c
+++ b/src/io/syncfile.c
@@ -440,6 +440,7 @@ MVMObject * MVM_file_open_fh(MVMThreadContext *tc, MVMString *filename, MVMStrin
     int fd;
     int flag;
     STAT_t statbuf;
+    uv_fs_t req;
 
     /* Resolve mode description to flags. */
     char * const fmode  = MVM_string_utf8_encode_C_string(tc, mode);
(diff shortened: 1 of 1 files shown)


src/io/syncfile.c
```diff
@@ -451,12 +452,9 @@ MVMObject * MVM_file_open_fh(MVMThreadContext *tc, MVMString *filename, MVMStrin
...
 
     /* Try to open the file. */
-#ifdef _WIN32
-    flag |= _O_BINARY;
-#endif
-    if ((fd = open((const char *)fname, flag, DEFAULT_MODE)) == -1) {
+    if ((fd = uv_fs_open(NULL, &req, (const char *)fname, flag, DEFAULT_MODE, NULL)) < 0) {
         char *waste[] = { fname, NULL };
-        const char *err = strerror(errno);
+        const char *err = uv_strerror(req.result);
```
MasterDuke17: Should this be included in `waste`?

ugexe: Auditing the use of `MVM_exception_throw_adhoc_free` I did not find that to be the case. If it is then there are many exceptions that can be fixed.

PR #1058, opened 2019-02-24
Use preadv(2) instead of read(2) via uv_fs_read

ugexe:


ugexe:
Feel free to deny this PR since I cannot actually see the advantages of `readv` or `pread` being used here anytime soon. This was primarily done as part of using more libuv.

ugexe:
I suppose we'll have to wait to hear the reasoning for https://github.com/MoarVM/MoarVM/commit/f17adad2be068b5fd338b6fa6931ce937311b334 which contains a nearly identical implementation 😅

Perhaps the use of `tc->loop` made it not thread safe? Or to remove a level of indirection to make use of additional `MVM_gc_mark_thread_blocked` easier/possible?

ugexe:
It seems to be ok to replace everything with libuv equivalents except `read/uv_fs_read` and `write/uv_fs_write` due to tty/eof handling on windows. So I'm going to close this for now, although I intend to poke further to figure out why libuv isn't working as expected here.

PR #1067, opened 2019-03-08
 Improve check for used toolchain on macOS 

vrurg:
Don't strictly test for make compliance because XCode is using GNU make
though earlier version of it. This shouldn't break the build process
unless a feature from later version of GNU make would make it into the
Makefile which won't work in XCode's version of make. But this would
draw the whole XCode build broken anyway.

Still, to be on the safe side, this commit makes Configure.pl forcingly
choose /usr/bin/make (which always belongs to XCode) unless GNU
toolchain is requested.


vrurg:
@bdw It does work for me. I'd be happy to see the author of the original commit to test it in his setup too though. Pinged him under the corresponding commit.

With regard to the hardcoded `make`, neither would I like to have it this way. But as long as it exclusively for macOS and only for XCode toolchain – that'd be ok.

manchicken:
I can check that out tonight. I don’t see Xcode using GNU make. Are you sure?

manchicken:
Oh, you can test this yourself by `brew install binutils` and then adding that into the beginning of your `$PATH`.

vrurg:
@manchicken This is the point: I don't use homebrew but MacPorts. I can't `brew install` and wouldn't mix up the two.

With regard to what is being used, here is `/usr/bin/make` output of my XCode:

```
GNU Make 3.81
Copyright (C) 2006  Free Software Foundation, Inc.
This is free software; see the source for copying conditions.
There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
PARTICULAR PURPOSE.

This program built for i386-apple-darwin11.3.0
```

It's old, partially re-branded, but still GNU make. This what MacPorts have installed for me:

```
GNU Make 4.2.1
Built for x86_64-apple-darwin18.0.0
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

So far, the Makefile generated is perfectly compatible with both. So, my hack for using the hardcoded path for XCode toolchain is merely a way to have _everything_ taken from the same basket just for the sake of predictability. Even though I'm not sure if there is any real sense in doing so.

manchicken:
I just tested it, this works for me. Thanks, @vrurg!

vrurg:
Thanks @manchicken ! Then we're good to go.

PR #1069, opened 2019-03-16
Give MVM_thread_new a stack size parameter

ugexe:
With the addition of uv_thread_create_ex in libuv 1.26 we can now
give use a custom stack size when creating a new thread. See
http://docs.libuv.org/en/v1.x/threading.html#c.uv_thread_create_ex
for more information on the libuv implementation.


ugexe:
Note that this cannot pass CI tests since there are nqp changes that go with this.

AlexDaniel:
Is this still relevant? Also, what's the justification for adding it?

PR #1072, opened 2019-03-20
Ask malloc to free kernel pages on GC collection

samcv:
This reduces the max memory usage used by asking malloc to free memory
pages. We set it to 128KiB which is glibc's malloc default. This should
hopefully ensure it does get collected.

Credit to:
https://www.joyfulbikeshedding.com/blog/2019-03-14-what-causes-ruby-memory-bloat.html

PS: just want to say this isn't ready for merge yet, as not all libc's (though most) have malloc_trim

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -518,6 +518,7 @@ if ($config{cc} eq 'cl') {
 build::probe::C_type_bool(\%config, \%defaults);
 build::probe::computed_goto(\%config, \%defaults);
 build::probe::pthread_yield(\%config, \%defaults);
+build::probe::check_fn_malloc_trim(\%config, \%defaults);
 if ($^O eq 'aix') {
     build::probe::numbits(\%config, \%defaults);
     $config{ldflags} = join(',', $config{ldflags}, '-bmaxdata:0x80000000')
--- a/build/Makefile.in
+++ b/build/Makefile.in
@@ -237,6 +237,7 @@ OBJECTS2 = src/6model/reprs/MVMDLLSym@obj@ \
           src/platform/sys@obj@ \
           src/platform/random@obj@ \
           src/platform/memmem32@obj@ \
+          src/platform/malloc_trim@obj@ \
           src/moar@obj@ \
           @platform@ \
           @jit_obj@
@@ -401,6 +402,7 @@ HEADERS = src/moar.h \
           src/platform/sys.h \
           src/platform/setjmp.h \
           src/platform/memmem.h \
+          src/platform/malloc_trim.h \
           src/platform/random.h \
           src/platform/fork.h \
           src/jit/graph.h \
--- a/build/config.h.in
+++ b/build/config.h.in
@@ -33,6 +33,10 @@
 #define MVM_HAS_PTHREAD_YIELD @has_pthread_yield@
 #endif
 
+#if @has_fn_malloc_trim@
+#define MVM_HAS_FN_MALLOC_TRIM @has_fn_malloc_trim@
+#endif
+
 /* How this compiler does static inline functions. */
 #define MVM_STATIC_INLINE @static_inline@
 
--- a/build/probe.pm
+++ b/build/probe.pm
@@ -407,6 +407,29 @@ EOT
     $config->{cancgoto} = $can_cgoto || 0
 }
 
+sub check_fn_malloc_trim {
+    my ($config) = @_;
+    my $restore = _to_probe_dir();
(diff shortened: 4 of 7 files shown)


AlexDaniel:
So, is there any before↔after comparison?

samcv:
@AlexDaniel 
I
 go from 8.7-8.2 max to 6.3-7.2 with my script that cats a linux iso 5 times doing `malloc_trim(128*1024);`

samcv:
https://gist.github.com/samcv/fce3ecdb5e8927d6a13092ed7b33f255

AlexDaniel:
OK, nice! But why 128KiB?

>  621   The `pad' argument to malloc_trim represents the amount of free
 622   trailing space to leave untrimmed. If this argument is zero,
 623   only the minimum amount of memory to maintain internal data
 624   structures will be left (one page or less). Non-zero arguments
 625   can be supplied to maintain enough trailing space to service
 626   future expected allocations without having to re-obtain memory
 627   from the system.

Given how memory-hungry rakudo is, wouldn't it be better to have that argument much larger (probably in megabytes)?

AlexDaniel:
Not sure how to benchmark it, maybe a loop that creates a lot of objects that are thrown away relatively quickly? I'm guessing it'd be a bit faster with higher pad sizes.

samcv:
@AlexDaniel the size used is the default. So no real reason why it should be that size. But the conditions that it releases memory to the kernel are not guaranteed to run with any regularity.

AlexDaniel:
> A GNU extension, called malloc_trim(), exists for releasing memory from the top of the heap, but it can be painfully slow. It hurts real bad for a lot of small objects, so it should be used sparingly.

[source](https://marek.vavrusa.com/memory/)

See also: https://dev.to/evilmartians/cables-vs-malloctrim-or-yet-another-ruby-memory-usage-benchmark-3emo

How often do we run gc? Do we actually need to run malloc_trim as often?

We really need a way to benchmark this…

Kaiepi:
There is no `malloc_trim` on *BSD.
On OpenBSD, you could use `freezero` to discard *all* memory, including what's cached in memory pages.
FreeBSD and NetBSD use jemalloc. I'll take a look at how it works tomorrow and see if I can find any way to solve this

Kaiepi:
`freezero` appears to fix the leak with the test code in the Rakudo issue when I build moar to use it in `src/io/asyncsocket.c` and `src/strings/utf8.c`:
```
bastille% perl6 test.p6
153.445313MiB maxrss – before doing anything!!
｢4b32fcab2192054b82312cde4c235adda0b8ac4d.zst｣
186.664063MiB maxrss 
｢eb3123e5e60fa1635ed1ee121cec1290c290044a.zst｣
203.773438MiB maxrss 
｢f66f8be09c6023b3b53e32f7571ea536ad87b87f.zst｣
212.054688MiB maxrss 
｢b2a3441749878e338b0861b14b3b9433cc902f42.zst｣
219.449219MiB maxrss 
｢78980ed447cceff82f5efef16dbe9ee437aae809.zst｣
232.742188MiB maxrss 
｢aa94ffc33712c6f3068c6472b79a74fa70aa7b2e.zst｣
232.742188MiB maxrss 
｢a0a28432f54d608130f17247f9202f4c4939dfff.zst｣
232.742188MiB maxrss 
```

A couple problems though:
- It can't be used everywhere in MoarVM currently since it requires a size argument, making it impossible to use with `void *` variables unless you already know what type they were before they were cast
- Where would it be appropriate to use `free` instead of `freezero`? Not zeroing out the memory with `freezero` is supposed to be useful in cases where the same memory gets allocated again.

This part of the fix might be better off in a separate pullreq.

Kaiepi:
Sorry, should've commented on the Rakudo issue, not here.

samcv:
@bdw regarding the status of this ticket. We will need to implement in Configure.{pl,pm} the ability to check if malloc_trim exists. This is the only reliable way to do this, due to various libC's/malloc implementations. This way the function will only do something in case the malloc_trim function exists. On my way to Bern I may have time to try looking at this. @jnthn we have existing code to check if a function exists or not?

jnthn:
@samcv I believe the stuff in `build/probe.pm` is doing that kind of check (invokving a C compiler, checking the result, etc.)

samcv:
I have updated the code so it checks for the malloc_trim() function. It should now work properly on all libc's, using malloc_trim only if it exists.

PR #1078, opened 2019-03-23
Add indexingoptimized exprjit template

ugexe:


src/jit/core_templates.expr
```diff
@@ -2341,6 +2341,12 @@
         (carg (tc) ptr)
         (carg $1 ptr)) int_sz)))
 
+(template: indexingoptimized
+  (call (^func &MVM_string_indexing_optimized)
+    (arglist
+      (carg (tc) ptr)
+      (carg $1 int)) ptr_sz))
```
MasterDuke17: Shouldn't this be `(carg $1 ptr)`?

PR #1081, opened 2019-03-23
Add objectid exprjit template

ugexe:


MasterDuke17:
Not sure about this. The interp.c implementation stores the MVMint64 casted result into a .i64. https://github.com/MoarVM/MoarVM/blob/6c7810ce7ca905d772ac2a3e47e73cf7c7c41ed8/src/core/interp.c#L4361-L4364

PR #1083, opened 2019-03-24
Add decodeconf and decoderepconf legojit

ugexe:


src/jit/graph.c
```diff
@@ -3751,6 +3752,20 @@ static MVMint32 consume_ins(MVMThreadContext *tc, MVMJitGraph *jg,
         jg_append_call_c(tc, jg, op_to_func(tc, op), 6, args, MVM_JIT_RV_PTR, dst);
         break;
     }
+    case MVM_OP_decodeconf: {
+        MVMint16 dst = ins->operands[0].reg.orig;
+        MVMint16 blob = ins->operands[1].reg.orig;
+        MVMint16 encoding = ins->operands[2].reg.orig;
+        MVMint16 replacement = ins->operands[3].reg.orig;
```
MasterDuke17: Shouldn't this be `NULL`?

ugexe: Ah yes. On a positive note it'll work for `decoderepconf`
https://github.com/MoarVM/MoarVM/blob/47bb50118e66fbe804ddd10ae08907a8bbb7bfc1/src/jit/core_templates.expr#L2464-L2471

PR #1085, opened 2019-03-24
Add backtrace and backtracestrings lego and exprjit template

ugexe:


MasterDuke17:
How come no template for `backtracestrings`?

ugexe:
> How come no template for backtracestrings?

It already exists. Its not next to this, but I guess thats a result of `tools/sort-expr-templates-by-oplist-order.p6`

MasterDuke17:
> It already exists. Its not next to this, but I guess thats a result of `tools/sort-expr-templates-by-oplist-order.p6`

Ah, ok.

PR #1088, opened 2019-03-31
Remove unnecessary (void *) cast in MVM_free_null

Kaiepi:
Code noise etc. etc.


Kaiepi:
Can't fix the commit title since I don't have access to git for a few more days

PR #1090, opened 2019-04-09
Add flag for UV_UDP_REUSEADDR

Garland-g:
This change allows the UV_UDP_REUSEADDR flag to be set on a udp socket.


lizmat:
Feels to me some comments are missing.  As to what bit means what in ssi->flags?  Could be missing before already.

Garland-g:
As far as I could tell, only 0x1 is used for setting broadcast on the socket.

I'd be happy to change/add relevant documentation, but I didn't see any comments relating to the flag(s), so I'm not sure where to put that.

PR #1098, opened 2019-04-26
Add various uniprop legojit

ugexe:


src/jit/graph.c
```diff
@@ -1958,7 +1965,46 @@ static MVMint32 consume_ins(MVMThreadContext *tc, MVMJitGraph *jg,
         jg_append_primitive(tc, jg, ins);
         break;
     }
-        /* Unspecialized parameter access */
+    case MVM_OP_getuniname: {
+        MVMint32 dst = ins->operands[0].reg.orig;
+        MVMJitCallArg args[] = { { MVM_JIT_INTERP_VAR, { MVM_JIT_INTERP_TC } },
+                                 { MVM_JIT_REG_VAL, { dst } } };
```
jnthn: This seems to be passing the destination as the argument, but it should be passing operand 1 (an integer codepoint) instead?

Issue #1068, opened 2019-03-09
SEGV when using --full-cleanup


MasterDuke17:
```
[dan@alexandria p6]$ valgrind /home/dan/Source/perl6/install/bin/moar --full-cleanup --execname="$0" --libpath="/home/dan/Source/perl6/install/share/nqp/lib" --libpath="/home/dan/Source/perl6/install/share/perl6/lib" --libpath="/home/dan/Source/perl6/install/share/perl6/runtime" /home/dan/Source/perl6/install/share/perl6/runtime/perl6.moarvm -e 'my $cmd = run « cat hs_ref_GRCh37.p5_chr1.fa », :out; $cmd.out.close'
==20944== Memcheck, a memory error detector
==20944== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==20944== Using Valgrind-3.14.0 and LibVEX; rerun with -h for copyright info
==20944== Command: /home/dan/Source/perl6/install/bin/moar --full-cleanup --execname=-bash --libpath=/home/dan/Source/perl6/install/share/nqp/lib --libpath=/home/dan/Source/perl6/install/share/perl6/lib --libpath=/home/dan/Source/perl6/install/share/perl6/runtime /home/dan/Source/perl6/install/share/perl6/runtime/perl6.moarvm -e my\ $cmd\ =\ run\ __\ cat\ hs_ref_GRCh37.p5_chr1.fa\ __,\ :out;\ $cmd.out.close
==20944== 
==20944== Invalid read of size 1
==20944==    at 0x4B63078: uv_loop_close (uv-common.c:660)
==20944==    by 0x4A6A766: MVM_io_eventloop_destroy (eventloop.c:297)
==20944==    by 0x4B272A8: MVM_vm_destroy_instance (moar.c:531)
==20944==    by 0x109563: main (main.c:312)
==20944==  Address 0xbe1b8b8 is 88 bytes inside a block of size 128 free'd
==20944==    at 0x48389AB: free (vg_replace_malloc.c:530)
==20944==    by 0x4A6A74F: MVM_free (alloc.h:40)
==20944==    by 0x4A6A74F: MVM_io_eventloop_destroy (eventloop.c:293)
==20944==    by 0x4B272A8: MVM_vm_destroy_instance (moar.c:531)
==20944==    by 0x109563: main (main.c:312)
==20944==  Block was alloc'd at
==20944==    at 0x483777F: malloc (vg_replace_malloc.c:299)
[...]

PR #1103, opened 2019-05-21
Move freeing of event_loop_wakeup to after uv_loop_close

pamplemoussecache:
Addresses part of the issue in #1068 . uv_loop_close tries to use instance->event_loop_wakeup, resulting in an invalid access of recently freed memory.

### DIFF
--- a/src/io/eventloop.c
+++ b/src/io/eventloop.c
@@ -292,11 +292,13 @@ void MVM_io_eventloop_destroy(MVMThreadContext *tc) {
 
     if (instance->event_loop) {
         uv_close((uv_handle_t*)instance->event_loop_wakeup, NULL);
-        instance->event_loop_wakeup = NULL;
 
         /* Not sure we can always do this */
         uv_loop_close(instance->event_loop);
+       
         MVM_free(instance->event_loop_wakeup);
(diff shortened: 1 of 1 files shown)


src/io/eventloop.c
```diff
@@ -292,11 +292,11 @@ void MVM_io_eventloop_destroy(MVMThreadContext *tc) {
 
     if (instance->event_loop) {
         uv_close((uv_handle_t*)instance->event_loop_wakeup, NULL);
-        MVM_free(instance->event_loop_wakeup);
         instance->event_loop_wakeup = NULL;
 
         /* Not sure we can always do this */
         uv_loop_close(instance->event_loop);
+        MVM_free(instance->event_loop_wakeup);
```
jnthn: I think the `instance->event_loop_wakeup = NULL;` would also need to move; otherwise, we set it to NULL, and then do `MVM_free(NULL)`, which silently does nothing.

pamplemoussecache: Sorry about that. Should be fixed now.

bdw:
@jnthn - if no objection, I'll merge this now

jnthn:
@bdw None from me.

PR #1106, opened 2019-05-26
Update libuv to version 1.29.1

dogbert17:
Spectested clean on Linux Mint 64 bit


jnthn:
Seems the next Rakudo release is still a couple of weeks off, so let's merge this.

PR #1108, opened 2019-05-31
Default to doing a non-relocatable build

patzim:
Non relocatability is the more common case with less caveats.
Remove the --no-relocatable config, add --relocatable instead.

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -47,7 +47,7 @@ GetOptions(\%args, qw(
     has-dyncall has-libffi pkgconfig=s
     build=s host=s big-endian jit! enable-jit
     prefix=s bindir=s libdir=s mastdir=s
-    no-relocatable make-install asan ubsan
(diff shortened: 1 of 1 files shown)


Configure.pl
```diff
@@ -95,7 +95,7 @@ sub uniq {
 $args{debug}        = 3 if defined $args{debug} and $args{debug} eq "";
 
 # Relocatability is not supported on AIX and OpenBSD.
-$args{'no-relocatable'} = 1 if $^O eq 'aix' || $^O eq 'openbsd';
+$args{'relocatable'} = 0 if $^O eq 'aix' || $^O eq 'openbsd';
```
jnthn: If somebody passed it explicitly, and we can't do it, should we not complain rather than silently override their choice?

patzim: Changed accordingly. Thanks for the hint!

PR #1111, opened 2019-06-04
Remove profiling overhead from profiles

MasterDuke17:
By calculating the average cost of executing the profiling functions and
then subtracting that times the number of entries from the exclusive
cost of an HLL block.

This gist (https://gist.github.com/MasterDuke17/ab7108070054f92e4cab5313033ea3f8) has some data for what the values were when https://github.com/MasterDuke17/MoarVM/blob/remove_profiling_overhead_from_profile/src/profiler/instrument.c#L399 and https://github.com/MasterDuke17/MoarVM/blob/remove_profiling_overhead_from_profile/src/profiler/instrument.c#L418 branches were taken when profiling `(^5_000).grep(*.is-prime).tail.say; my $a = (^100).pick(2).join; my $b; for ^100_000 { ++$b if $a ~~ / $_ / }; say $b`.

### DIFF
--- a/src/profiler/profile.c
+++ b/src/profiler/profile.c
@@ -8,8 +8,34 @@ void MVM_profile_start(MVMThreadContext *tc, MVMObject *config) {
     if (MVM_repr_exists_key(tc, config, tc->instance->str_consts.kind)) {
         MVMString *kind = MVM_repr_get_str(tc,
             MVM_repr_at_key_o(tc, config, tc->instance->str_consts.kind));
-        if (MVM_string_equal(tc, kind, tc->instance->str_consts.instrumented))
+        if (MVM_string_equal(tc, kind, tc->instance->str_consts.instrumented)) {
+            MVMuint32 i;
+            MVMuint64 s, e;
             MVM_profile_instrumented_start(tc, config);
+
+            /* Call the profiling functions a bunch of times and record how long they took. */
+            s = uv_hrtime();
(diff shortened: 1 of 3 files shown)


src/profiler/profile.c
```diff
@@ -8,8 +8,34 @@ void MVM_profile_start(MVMThreadContext *tc, MVMObject *config) {
...
             MVM_profile_instrumented_start(tc, config);
+
+            /* Call the profiling functions a bunch of times and record how long they took. */
+            s = uv_hrtime();
+            for (i = 0; i < 1000; i++) {
+                MVM_profile_log_enter(tc, tc->cur_frame->static_info, MVM_PROFILE_ENTER_NORMAL);
+                MVM_profile_log_exit(tc);
+            }
+            e = uv_hrtime();
+            tc->instance->profiling_overhead = (MVMuint64) ((e - s) / 1000) * 0.9;
```
jnthn: What's the `* 0.9` here?

MasterDuke17:
Just a scaling factor in case the measurement is a little high.

jnthn:
> Just a scaling factor in case the measurement is a little high.

Hm, from the loop overhead, perhaps? It's quite hard to reason about, since the C compiler may also do inlining and cancel the cost of the function call overhead.

I guess let's give this a try, though; I think even slightly off compensation is going to give better numbers than none at all.

PR #1114, opened 2019-06-07
Implement freemem and totalmem ops

Kaiepi:


Kaiepi:
See https://github.com/rakudo/rakudo/pull/2959

src/platform/sys.c
```diff
@@ -1,17 +1,25 @@
...
-    int count;
+MVMint64 MVM_platform_cpu_count(void) {
+    MVMint64       count;
     uv_cpu_info_t *info;
-    int e;
+    int            e;
 
-    e = uv_cpu_info(&info, &count);
-    if (e == 0) uv_free_cpu_info(info, count);
+    e = uv_cpu_info(&info, (int *)(&count));
```
jnthn: Why was the type of `count` changed from `int`? This could cause problems on big endian, maybe? I'd change `count` back to `int` and then just have it upcast to `MVMint64` implicitly at return.

src/core/interp.c
```diff
@@ -5175,7 +5175,15 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
...
-                GET_REG(cur_op, 0).i32 = MVM_platform_cpu_count();
+                GET_REG(cur_op, 0).i64 = MVM_platform_cpu_count();
+                cur_op += 2;
+                goto NEXT;
+            OP(freemem):
+                GET_REG(cur_op, 0).i64 = MVM_platform_free_memory();
+                cur_op += 2;
+                goto NEXT;
+            OP(totalmem):
+                GET_REG(cur_op, 0).i64 = MVM_platform_total_memory();
```
jnthn: New ops should be added to `interp.c` in the same order as they appear in `oplist`, so these new ones should come after `uname`.

src/platform/sys.c
Kaiepi: I was naively thinking MoarVM only supported amd64 because of the legojit, which would've made `int` 64 bits. It was a dumb assumption on my part. I'll change it back

jnthn: We only JIT on that platform, but we try - with varying degrees of success - to run on others. :-)

PR #1116, opened 2019-06-08
fix rakudo/rakudo#2965

kazcw:
Exception rather than segfault in response to an attempt to HAS a type before it's defined (rakudo/rakudo#2965)


kazcw:
After this patch, the following throws 3 compile-time exceptions instead of segfaulting:

```
class OopsUnion is repr(｢CUnion｣) { HAS OopsUnion $.more; }
class OopsStruct is repr(｢CStruct｣) { HAS OopsStruct $.more; }
class OopsPPStruct is repr(｢CPPStruct｣) { HAS OopsPPStruct $.more; }
```

I'm trying to figure out how to write a test that passes when it throws an exception at compile time

MasterDuke17:
`throws-like` in an EVAL maybe?

kazcw:
Hm, this causes the new exception to be thrown upon attempts to HAS objects of types that aren't CArray/CUnion/CP*STruct. I should move the check into the appropriate switch branches so it fails right for the never-HASable cases...

kazcw:
Ok, I made the check more fine grained so the new error will never occur when a null-derived pointer dereference didn't before

PR #1117, opened 2019-06-08
fix warning in Configure in non-relocatable builds

kazcw:
Use of uninitialized value $args{"relocatable"} in numeric eq (==) at ./Configure.pl line 98.


kazcw:
This also disables overriding the test with `--relocatable=force`, which I don't think was an intentional feature?

jnthn:
> This also disables overriding the OpenBSB check with --relocatable=force, which I don't think was an intentional feature?

Good question; @patzim ?

patzim:
> This also disables overriding the OpenBSB check with `--relocatable=force`, which I don't think was an intentional feature?

No this was not intentional. Good find, kazcw++! I aprove of this PR being merged.

Issue #1087, opened 2019-03-30
MoarVM panic when using PKafka


kbucheli:
We want to use Perl 6 as Kafka client. When I want to save the offset in the queue after I processed the message I get some MoarVM panics:

    MoarVM panic: Internal error: zeroed target thread ID in work pass

or sometimes

    MoarVM panic: Internal error: invalid thread ID 8614688 in GC work pass

This only happens after running it for some time (< 1 minute)
Code below:

```
use PKafka::Consumer;
use PKafka::Message;
use PKafka::Config;

sub MAIN () {
   my $exception-channel = Channel.new;
   for (0..5) -> $partition {

       say "start $partition";

       my $kafka-config = PKafka::Config.new({
               'metadata.broker.list' => 'kafka-dev-1:5551, kafka-dev-2:5551, kafka-dev-3:5551',
               'security.protocol' => 'ssl',
               'ssl.ca.location' => '/etc/ssl/ca_server_cert.pem',
               'ssl.certificate.location' => '/etc/ssl/client_cert.pem',
               'ssl.key.location' => '/etc/ssl/client_key.pem',
               'ssl.key.password' => 'password',
               'group.id' => 'metrics-feed-consumer',
               'enable.auto.offset.store' => 'false',
       });
       my $kafka-consumer = PKafka::Consumer.new(
            topic => 'monitoring.metrics', brokers => 'kafka-dev-1:5551, kafka-dev-2:5551, kafka-dev-3:5551',
            config => $kafka-config
       );
       $kafka-consumer.messages.tap(-> $msg {
           given $msg {
               when PKafka::Message {
                   say "$partition : got {$msg.payload-str}";
                   $kafka-consumer.save-offset($msg); # <--- guilty line
               }
           }
       });

       my $worker-done = $kafka-consumer.consume-from-last(partition => $partition);

       $exception-channel.send($worker-done);
   }

   react {
       # handle exceptions
       whenever $exception-channel -> $promise {
           whenever $promise {}
       }

       # fininsh up properly with signals
       whenever signal(SIGINT,SIGKILL) { done }
   }
}
```

Could that an MoarVM issue? Could it be some threading issue withing librdkafka? Or in the interaction between both?

```
# perl6 -v
This is Rakudo version 2019.03.1 built on MoarVM version 2019.03
implementing Perl 6.d.
```

bdw:
Yes, this is threading related.

AlexDaniel:
[...]

Issue #800, opened 2018-02-10
MoarVM panic: Internal error: zeroed target thread ID in work pass


domq:
The attached script [bugreport.pl.txt](https://github.com/MoarVM/MoarVM/files/1713295/bugreport.pl.txt) crashes most of the time with the error message in the issue's title, but something with others (see below). May or may not be related to #473.

My platform: Mac OS X Sierra 10.12.6, perl6 from homebrew.

```
Mac:~$ perl6 --version
This is Rakudo Star version 2018.01 built on MoarVM version 2018.01
implementing Perl 6.c.
```

I unfortunately have no idea how to determine the version of Gumbo and XML::XPath that I have installed ([this](https://stackoverflow.com/questions/44883887/how-do-i-find-the-version-and-authority-of-a-perl-6-module) turns up nothing); but I installed both with `zef` minutes ago.

More version info:
```
Mac:~$ brew info gumbo-parser
gumbo-parser: stable 0.10.1 (bottled)
C99 library for parsing HTML5
https://github.com/google/gumbo-parser
/usr/local/Cellar/gumbo-parser/0.10.1 (10 files, 754.8KB) *
  Poured from bottle on 2018-02-10 at 10:27:34
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/gumbo-parser.rb
Mac:~$ brew info libxml2
libxml2: stable 2.9.7 (bottled), HEAD [keg-only]
GNOME XML library
http://xmlsoft.org
/usr/local/Cellar/libxml2/2.9.4_2 (277 files, 9.8MB)
  Poured from bottle on 2017-05-17 at 14:18:56
/usr/local/Cellar/libxml2/2.9.4_3 (281 files, 10.4MB)
  Poured from bottle on 2017-07-03 at 20:06:46
/usr/local/Cellar/libxml2/2.9.7 (281 files, 10.4MB)
  Poured from bottle on 2017-11-15 at 14:52:17
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/libxml2.rb
==> Options
--HEAD
	Install HEAD version
[snip]
```

The results of running the script 10 times on my platform:

```
Mac:~$ for i in $(seq 1 10); do perl6 ./bugreport.pl ; done
Parsed
MoarVM panic: Internal error: zeroed target thread ID in work pass
Parsed
Type check failed in binding to parameter '$start'; expected XML::Node but got Array ($[(my \XML::Element_1...)
[...]

Issue #751, opened 2017-11-13
MoarVM panic: Internal error: invalid thread ID <some-random-ID> in GC work pass


azawawi:
The following script generates a panic on native call buffer read access when the final binary blob is created. Please see [here](https://github.com/azawawi/perl6-msgpack/blob/master/lib/MsgPack/Packer.pm6#L25). On random occasions, it also generates a `Heap corruption error`. The error is replicated every time on my Ubuntu 16.04 16-GB laptop with libmsgpack v1.4.2 and v2.1.5 (latest). How can I debug this?

```
use v6;
use MsgPack;

say "MsgPack version: " ~ MsgPack::version;

my $data     = [1 xx 200_000];
my $packed   = MsgPack::pack($data);

say "data     : " ~ $data.elems;
say "packed   : " ~ $packed.elems;
```

zoffixznet:
Can you please include the `perl6` version this happens in? Run `perl6 -v`

AlexDaniel:
This is Rakudo version 2017.10-176-g68b2891d9 built on MoarVM version 2017.10-57-gd257df23
implementing Perl 6.c.

azawawi:
@zoffixznet The latest as usual.

jnthn:
> How can I debug this?

1. Try running it under `perl6-valgrind-m` and report the results here
2. Build MoarVM with `MVM_GC_DEBUG` set to 1 in `src/gc/debug.h`, then run it under `perl6-gdb-m` with a breakpoint on `MVM_panic`. If it panics, paste the `bt` here also.

azawawi:
@jnthn Sorry for the late reply. I think I found part of the problem. Please see their simple buffer (sbuffer) implementation [here](https://github.com/msgpack/msgpack-c/blob/master/include/msgpack/sbuffer.h#L80). Are we caching CArray pointers in NativeCall (See [here](https://github.com/azawawi/perl6-msgpack/blob/master/lib/MsgPack/Native.pm6#L16)]?

AlexDaniel:
@azawawi can you retest? Should be fixed now.

MasterDuke17:
I just tested and the script doesn't panic, closing as fixed.

PR #1119, opened 2019-06-08
enforce wb in get_attribute of C reprs

kazcw:
Fixes non-deterministic crashes in anything that non-trivially uses
CStruct/CPPStruct/CUnion reprs.

The type of failure this fixes usually matches these 3 patterns, alternately:
- ｢Segmentation fault｣
- ｢MoarVM panic: Internal error: zeroed target thread ID in work pass｣
- /'MoarVM panic: Internal error: invalid thread ID '\d+' in GC work pass'/

Also fixed a separate bug in CUnion: it wasn't caching child objects for
attributes that are inlined.

Tested to fix the issue in MoarVM #800, Skarsnik/perl6-gumbo#5 and [perl6 RT 131003](https://rt.perl.org/Public/Bug/Display.html?id=131003).
Probably also fixes MoarVM #751 and #1087

### DIFF
--- a/src/6model/reprs/CPPStruct.c
+++ b/src/6model/reprs/CPPStruct.c
@@ -483,7 +483,7 @@ static void get_attribute(MVMThreadContext *tc, MVMSTable *st, MVMObject *root,
                                 obj = MVM_repr_box_str(tc, typeobj, str);
                             });
                         }
-                        child_objs[real_slot] = obj;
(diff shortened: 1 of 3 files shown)


AlexDaniel:
This is amazing!!!! I just tested it, and indeed gumbo no longer crashes.

@kazcw by any chance can you add tests? Also, how did you figure it out? I've been struggling with this issue for years and multiple people have looked at it, we never got a fix until you showed up.

PR #1122, opened 2019-06-16
Fix MVM_file_isexecutable on Windows

hankache:
Fix https://github.com/rakudo/rakudo/issues/2610

### DIFF
--- a/src/io/fileops.c
+++ b/src/io/fileops.c
@@ -229,13 +229,15 @@ MVMint64 MVM_file_isexecutable(MVMThreadContext *tc, MVMString *filename, MVMint
                     MVMString *fileext = MVM_string_substring(tc, filename, n, -1);
                     char *ext  = MVM_string_utf8_c8_encode_C_string(tc, fileext);
                     char *pext = getenv("PATHEXT");
-                    int plen   = strlen(pext);
-                    int i;
-                    for (i = 0; i < plen; i++) {
-                        if (0 == stricmp(ext, pext++)) {
-                             r = 1;
-                             break;
+                    char *delim = ";";
+                    char *ptr = strtok(pext, delim);
+
+                    while(ptr != NULL) {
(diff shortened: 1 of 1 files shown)


src/io/fileops.c
```diff
@@ -229,13 +229,15 @@ MVMint64 MVM_file_isexecutable(MVMThreadContext *tc, MVMString *filename, MVMint
...
                     char *ext  = MVM_string_utf8_c8_encode_C_string(tc, fileext);
                     char *pext = getenv("PATHEXT");
-                    int plen   = strlen(pext);
-                    int i;
-                    for (i = 0; i < plen; i++) {
-                        if (0 == stricmp(ext, pext++)) {
-                             r = 1;
-                             break;
+                    char *delim = ";";
+                    char *ptr = strtok(pext, delim);
```
niner: From the man page of strtok:
Be cautious when using these functions.  If you do use them, note that:
* These functions modify their first argument.
* These functions cannot be used on constant strings.
* The identity of the delimiting byte is lost.
* The strtok() function uses a static buffer while parsing, so it's not thread safe.  Use strtok_r() if this matters to you.

All of these except for the bit about constant strings are real issues. E.g. this file test will only work once and then only find matches for the very first extension in PATHEXT. There's also gonna be issues if any other thread uses strtok for any purpose concurrently (which may happen with poorly written native libraries which are not aware that they are used in a multi threaded program)

ugexe:
Seems a bit weird to have done this at all. `PATHEXT` tells cmd.exe what extensions to try when launching some command/file without an extension, but `CreateProcess` only tries `.exe`. Neither of these relate to file permissions. Perhaps removal is a better option?

Here is the commit where it was added: https://github.com/MoarVM/MoarVM/commit/532f2620cbb5545b5a468ac28a0e1d1ae17e9ee2

PR #1126, opened 2019-06-21
[IP6NS Grant] Make socket family handling portable

Kaiepi:
MVM_io_resolve_host_name assumes a 64-bit system where the values of the
AF constants are the same as those on Linux. This breaks domain name
resolution on 32-bit systems and OSes where the values of AF constants
aren't the same as Linux's, such as FreeBSD.

This makes the socket family a separate parameter from the port and uses
platform-independent socket family constants, which are translated to
the correct AF constant by MVM_io_resolve_host_name to use with
getaddrinfo.

Related to https://github.com/rakudo/rakudo/issues/3007


Kaiepi:
The original commit message for this is wrong. This doesn't break on 32 bit systems since `int64_t` is being used anyway. But what was also happening was Rakudo was never setting the family to the upper bits of the port passed to `MVM_io_resolve_host_name`, meaning that its value could've been anything depending on the compiler.

PR #1128, opened 2019-06-24
Add missing concreteness checks to most array ops

Kaiepi:
This fixes a ton of segfaults when type objects are passed to array ops,
along with one case in particular where splicing a type object on index
0, inserting 0 elements from another type object would cause moar to
attempt to allocate an 18 quintillion element array.


jnthn:
Makes sense to me; thanks!

PR #1130, opened 2019-07-02
Throw if MVM_bigint_is_prime called with too many rounds

MasterDuke17:
Valid range is 0..PRIME_SIZE (defined in libtommath).

Fixes https://github.com/perl6/nqp/issues/560.

### DIFF
--- a/src/math/bigintops.c
+++ b/src/math/bigintops.c
@@ -1203,8 +1203,10 @@ MVMint64 MVM_bigint_is_prime(MVMThreadContext *tc, MVMObject *a, MVMint64 b) {
             return 0;
         }
         else {
-            int result;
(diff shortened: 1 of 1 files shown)


AlexDaniel:
The error message is oddly specific even though it checks for just `MP_OKAY`. What about using `MP_ITER` and having a more generic error message if something else happens (though I'm not sure if there's any other error possible)?

jnthn:
@AlexDaniel That could be worth doing also; didn't see the comment until after hitting merge, but either way, this is an improvement on silent failure...

AlexDaniel:
Case in point: https://colabti.org/irclogger/irclogger_log/perl6-dev?date=2019-08-31#l45

PR #1131, opened 2019-07-04
Build with spaces in path

patzim:
The only bit missing was some quoting to get the build install the include
files successfully.
This makes spaces in path work on Windows only. On Linux the build still fails because the dyncall build fails when path has spaces in it. I did notify the dyncall team about that.

See [rakudo#3031](https://github.com/rakudo/rakudo/pull/3031).

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -272,10 +272,10 @@ if ($args{'has-libuv'}) {
 else {
     $config{cincludes} .= ' ' . $defaults{ccinc} . '3rdparty/libuv/include'
                         . ' ' . $defaults{ccinc} . '3rdparty/libuv/src';
-    $config{install}   .= "\t\$(MKPATH) \$(DESTDIR)\$(PREFIX)/include/libuv\n"
-                        . "\t\$(MKPATH) \$(DESTDIR)\$(PREFIX)/include/libuv/uv\n"
-                        . "\t\$(CP) 3rdparty/libuv/include/*.h \$(DESTDIR)\$(PREFIX)/include/libuv\n"
-                        . "\t\$(CP) 3rdparty/libuv/include/uv/*.h \$(DESTDIR)\$(PREFIX)/include/libuv/uv\n";
+    $config{install}   .= "\t\$(MKPATH) \"\$(DESTDIR)\$(PREFIX)/include/libuv\"\n"
+                        . "\t\$(MKPATH) \"\$(DESTDIR)\$(PREFIX)/include/libuv/uv\"\n"
+                        . "\t\$(CP) 3rdparty/libuv/include/*.h \"\$(DESTDIR)\$(PREFIX)/include/libuv\"\n"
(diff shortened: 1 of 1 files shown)


AlexDaniel:
I'm curious, so this makes spaces work but `"` characters in paths are now illegal, right? This is fine on Windows because `"` is not allowed in filenames, but on linux that'd just sweep things under the rug. Last time I checked, Makefiles were really stupid when it came to proper handling of filenames. However, that is in Configure.pl, so can't we do some proper escaping there?

patzim:
In Rakudo and NQP the quoting in Makefiles is abstracted in a few macros (thanks to the shiny templating system vrurg introduced), so there the quoting can change depending on the `make` flavour. Moar hasn't yet received that treatment. Thus the plain "put stuff in quotes" solution.
I estimate that having spaces in the path on Windows is more common than having quotes in the path on Linux. So I'd say let's just go with the quotes for now and once moar gets the same shiny template system as rakudo and nqp now have, we get the full solution.

Configure.pl
```diff
@@ -286,25 +286,25 @@ sub uniq {
...
+                        . "\t\$(CP) 3rdparty/libatomicops/src/atomic_ops/sysdeps/*.h \"$lao/atomic_ops/sysdeps\"\n"
+                        . "\t\$(CP) 3rdparty/libatomicops/src/atomic_ops/sysdeps/armcc/*.h \"$lao/atomic_ops/sysdeps/armcc\"\n"
+                        . "\t\$(CP) 3rdparty/libatomicops/src/atomic_ops/sysdeps/gcc/*.h \"$lao/atomic_ops/sysdeps/gcc\"\n"
+                        . "\t\$(CP) 3rdparty/libatomicops/src/atomic_ops/sysdeps/hpc/*.h \"$lao/atomic_ops/sysdeps/hpc\"\n"
+                        . "\t\$(CP) 3rdparty/libatomicops/src/atomic_ops/sysdeps/ibmc/*.h \"$lao/atomic_ops/sysdeps/ibmc\"\n"
+                        . "\t\$(CP) 3rdparty/libatomicops/src/atomic_ops/sysdeps/icc/*.h \"$lao/atomic_ops/sysdeps/icc\"\n"
+                        . "\t\$(CP) 3rdparty/libatomicops/src/atomic_ops/sysdeps/loadstore/*.h \"$lao/atomic_ops/sysdeps/loadstore\"\n"
+                        . "\t\$(CP) 3rdparty/libatomicops/src/atomic_ops/sysdeps/msftc/*.h \"$lao/atomic_ops/sysdeps/msftc\"\n"
+                        . "\t\$(CP) 3rdparty/libatomicops/src/atomic_ops/sysdeps/sunc/*.h \"$lao/atomic_ops/sysdeps/sunc\"\n";
```
bdw: Ideally maybe this would've been a Makefile include, that would've made it decidedly less messy.

PR #1134, opened 2019-07-08
JIT - Floating Point Support for Expression JIT

bdw:
This PR adds basic floating point support for the expression JIT.
Some of the design decisions:
- no different operators for arithmetic operations, store, copy etc. between integer and floating point operands. (Basically, every operator whose implementation and output types are determined completely by the input type, is treated as generic, because the tiler can sort out the difference perfectly). Types are checked by the expression template compiler and to a lesser extent by the JIT.
- in the register allocator (and emitter), treat register sets consistently as bitmaps, register identifiers as numbers. In x86-64, we treat the register set as a 32-bit bitmap (16 GPR, 16 SSE registers, ignoring x87 and MMX registers for the moment!). Live ranges also get a set of permissible registers - this helps to handle register requirerements.
- Register allocator now enforces 'two-operand' order - i.e, the output argument should be the same, as the first input argument, for binary operators.


bdw:
Oh, it also pulls in an instruction encoding fix for DynASM.

src/jit/core_templates.expr
```diff
@@ -171,12 +193,25 @@
...
   (call (^func &MVM_coerce_s_i)
     (arglist
       (carg (tc) ptr)
       (carg $1   ptr)) int_sz))
 
+(template: coerce_sn
+  (calln (^func &MVM_coerce_s_n)
+    (arglist
+      (carg (tc) ptr)
+      (carg $1   ptr))))
```
MasterDuke17: This shouldn't be `(carg $1   ptr)) num_sz))`?

ugexe: > This shouldn't be `(carg $1 ptr)) num_sz))`?

Note this uses the new `calln`, not `call`

src/jit/x64/tiles.dasc
```diff
@@ -827,34 +852,78 @@ MVM_JIT_TILE_DECL(cmp) {
     }
 }
 
+MVM_JIT_TILE_DECL(cmp_num) {
+    MVMint8 left = REG_NUM(tile->values[1]);
+    MVMint8 right = REG_NUM(tile->values[2]);
+    | ucomisd xmm(left), xmm(right);
```
bdw: There's multiple variants of these, and the IR can't express the difference

PR #1137, opened 2019-07-11
Implement NativeCall wide string support

Kaiepi:
See the Rakudo pullreq for info on what this does


Kaiepi:
Looks like AppVeyor is unhappy with my use of `stringapiset.h`, which isn't available before Vista... I'll need to find some other way to transcode wide strings to UTF-8 since Windows didn't support it until very recently. Since Windows wide strings are UTF-16, I'll probably have to make some changes to `src/strings/utf16.c` to support transcoding strings of types other than `MVMString` and `char *`.

src/strings/wide.c
```diff
@@ -0,0 +1,41 @@
...
+    return str;
+}
+
+MVMwchar * MVM_string_wide_encode(MVMThreadContext *tc, MVMString *str, MVMuint64 *output_size) {
+          MVMwchar *wstr;
+    const char     *cstr;
+          size_t    size;
+
+#ifdef _MSC_VER
+    MVMwchar *wide = (MVMwchar *)MVM_string_utf16_encode_substr(tc, str, output_size, 0, -1, NULL, 0);
```
zhuomingliang: looks like `wide` is not used.

Kaiepi: Oop, that should be `wstr`

patzim:
Is there a reason why this isn't merged yet? Do the core devs need a poke?

lizmat:
It feels like all of Kaiepi's work will be lost unless someone reimagines this work.

@patrickbkr @timo @niner

lizmat:
What would be needed to salvage this PR that had a lot of work put into it?

zhuomingliang:
> What would be needed to salvage this PR that had a lot of work put into it?


If it doesn't break test, I would suggest merging into main branch after resolving conflicts, and improving it later, so these work won't be lost.

PR #1139, opened 2019-07-13
Add context to error messages

MasterDuke17:

### DIFF
--- a/src/core/args.c
+++ b/src/core/args.c
@@ -760,7 +760,7 @@ static void flatten_args(MVMThreadContext *tc, MVMArgProcContext *ctx) {
             MVMStorageSpec  lss   = REPR(list)->pos_funcs.get_elem_storage_spec(tc, STABLE(list));
 
             if ((MVMint64)new_arg_pos + count > 0xFFFF) {
-                MVM_exception_throw_adhoc(tc, "Too many arguments (%"PRId64") in flattening array, only %"PRId32" allowed.", (MVMint64)new_arg_pos + count), 0xFFFF;
+                MVM_exception_throw_adhoc(tc, "Too many arguments (%"PRId64") in flattening array, only %"PRId32" allowed.", (MVMint64)new_arg_pos + count, 0xFFFF);
             }
 
             for (i = 0; i < count; i++) {
--- a/src/core/args.c
+++ b/src/core/args.c
@@ -409,25 +409,13 @@ void MVM_args_assert_nameds_used(MVMThreadContext *tc, MVMArgProcContext *ctx) {
     MVMuint16 i;
     if (size > 64) {
         for (i = 0; i < size; i++)
-            if (!ctx->named_used.byte_array[i]) {
-                char *c_param = MVM_string_utf8_encode_C_string(tc,
-                    ctx->args[ctx->num_pos + 2 * i].s);
-                char *waste[] = { c_param, NULL };
-                MVM_exception_throw_adhoc_free(tc, waste,
-                    "Unexpected named argument '%s' passed",
-                    c_param);
-            }
+            if (!ctx->named_used.byte_array[i])
+                MVM_args_throw_named_unused_error(tc, ctx->args[ctx->num_pos + 2 * i].s);
     }
     else {
         for (i = 0; i < size; i++)
-            if (!(ctx->named_used.bit_field & ((MVMuint64)1 << i))) {
-                char *c_param = MVM_string_utf8_encode_C_string(tc,
-                    ctx->args[ctx->num_pos + 2 * i].s);
-                char *waste[] = { c_param, NULL };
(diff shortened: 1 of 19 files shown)


src/core/args.c
```diff
@@ -409,25 +409,13 @@ void MVM_args_assert_nameds_used(MVMThreadContext *tc, MVMArgProcContext *ctx) {
...
-            if (!(ctx->named_used.bit_field & ((MVMuint64)1 << i))) {
-                char *c_param = MVM_string_utf8_encode_C_string(tc,
-                    ctx->args[ctx->num_pos + 2 * i].s);
-                char *waste[] = { c_param, NULL };
-                MVM_exception_throw_adhoc_free(tc, waste,
-                    "Unexpected named argument '%s' passed",
-                    c_param);
-            }
+            if (!(ctx->named_used.bit_field & ((MVMuint64)1 << i)))
+                MVM_args_throw_named_unused_error(tc, ctx->args[ctx->num_pos + 2 * i].s);
```
bdw: In C, please use explicit brackets by default; as a side effect, that would've reduced the scope of the diff to contain only the block; which would've made it easier to review :smile:

src/core/interp.c
```diff
@@ -489,8 +489,11 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                             found = 1;
                         }
                     }
-                    if (!found)
-                        MVM_exception_throw_adhoc(tc, "setstaticlex given invalid lexical name");
+                    if (!found) {
+                        char *c_name = MVM_string_utf8_encode_C_string(tc, name);
+                        char *waste[] = { c_name, NULL };
+                        MVM_exception_throw_adhoc_free(tc, waste, "setstaticlex given invalid lexical name '%s'", c_name);
```
bdw: This is such a common pattern, I wonder whether it warrants a helper function

src/core/interp.c
```diff
@@ -5715,9 +5744,10 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                     target->o = check;
                 goto NEXT;
             }
-            OP(sp_rebless):
-                if (!REPR(GET_REG(cur_op, 2).o)->change_type) {
-                    MVM_exception_throw_adhoc(tc, "This REPR cannot change type");
+            OP(sp_rebless): {
+                MVMObject *obj = GET_REG(cur_op, 2).o;
+                if (!REPR(obj)->change_type) {
+                    MVM_exception_throw_adhoc(tc, "REPR %s (%s) cannot change type", REPR(obj)->name, MVM_6model_get_debug_name(tc, obj));
```
bdw: Why not refactor the rest of the references to obj as well?

src/strings/utf8.c
```diff
@@ -384,8 +384,25 @@ MVMuint32 MVM_string_utf8_decodestream(MVMThreadContext *tc, MVMDecodeStream *ds
...
+                    }
+                    else if (bufsize == 1) {
+                        MVMGrapheme32 a = buffer[pos];
+                        MVM_free(buffer);
+                        MVM_exception_throw_adhoc(tc, "Malformed UTF-8 near byte %02x", a);
+                    }
+                    else {
+                        MVM_free(buffer);
+                        MVM_exception_throw_adhoc(tc, "Malformed UTF-8");
+                    }
```
bdw: I don't like the repeated free of `buffer` here... you could do a `char first[3]; memcpy(first, buffer, MAX(bufsize, sizeof(first))` and use that? (Alternatively, add `buffer` as `waste` argument to `MVM_exception_throw_adhoc_free`)

src/core/args.c
MasterDuke17: I'm generally a fan of explicit brackets, but the codebase has a **lot** of `if`s without them. I'd almost say without is more common. Should I change all the ones in this PR?

src/core/interp.c
MasterDuke17: Possibly, but that seems like it would belong in a separate PR.

src/core/interp.c
MasterDuke17: Hm. Or should I just partially revert and use `GET_REG(cur_op, 2).o` in the arguments to `MVM_exception_throw_adhoc`?

src/strings/utf8.c
MasterDuke17: Addressed in ab86372.

src/core/args.c
bdw: What I meant to say is that, if you had used, or rather left, the brackets around the block, the diff would've contained only the actual contents of the block. Which makes it easy to assert that the conditional itself hadn't changed

MasterDuke17: Oh yeah, I see now. Want me to add the braces back?

PR #1145, opened 2019-07-19
Fixed some small bugs exposed by `-Wall`.

Aearnus:
Hello!

This pull request fixes a couple minor bugs that showed up when I enabled `-Wall` (and one that showed up without it). The bugs include an `printf` formatting issue, an improperly specified return value, and a couple values which could end up uninitialized or strange depending on the code path taken.

I also added explicit defaults on a couple switch statements to be more explicit about their intents.

### DIFF
--- a/src/io/procops.c
+++ b/src/io/procops.c
@@ -737,7 +737,7 @@ static void spawn_setup(MVMThreadContext *tc, uv_loop_t *loop, MVMObject *async_
             MVMObject *error_cb;
             MVMString *msg_str;
 
-            snprintf(error_str, 127, "Failed to spawn process %s: %s (error code %d)",
+            snprintf(error_str, 127, "Failed to spawn process %s: %s (error code %ld)",
                     si->prog, uv_strerror(spawn_result), spawn_result);
 
             msg_str = MVM_string_ascii_decode_nt(tc,
--- a/src/jit/compile.c
+++ b/src/jit/compile.c
@@ -339,7 +339,8 @@ MVMint32 MVM_jit_spill_memory_select(MVMThreadContext *tc, MVMJitCompiler *compi
         idx = compiler->spills_free[bucket];
         compiler->spills_free[bucket] = compiler->spills[idx].next;
     } else {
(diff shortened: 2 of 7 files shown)


src/io/procops.c
```diff
@@ -737,7 +737,7 @@ static void spawn_setup(MVMThreadContext *tc, uv_loop_t *loop, MVMObject *async_
             MVMObject *error_cb;
             MVMString *msg_str;
 
-            snprintf(error_str, 127, "Failed to spawn process %s: %s (error code %d)",
+            snprintf(error_str, 127, "Failed to spawn process %s: %s (error code %ld)",
```
cygx: given that `uv_spawn()` returns an `int` (cf [uv.h](https://github.com/libuv/libuv/blob/c4e9657d59723a6273ce68bda935680e1b3596c5/include/uv.h#L1000)), personally I'd adjust the declaration of `spawn_result` instead...

Aearnus: I agree -- fixed that & changed the type.

src/jit/compile.c
```diff
@@ -339,7 +339,8 @@ MVMint32 MVM_jit_spill_memory_select(MVMThreadContext *tc, MVMJitCompiler *compi
         idx = compiler->spills_free[bucket];
         compiler->spills_free[bucket] = compiler->spills[idx].next;
     } else {
-        MVM_VECTOR_ENSURE_SPACE(compiler->spills, idx = compiler->spills_num++);
+        idx = compiler->spills_num; compiler->spills_num++;
```
jnthn: It'd be clearer if this were split over two lines.

src/spesh/dump.c
```diff
@@ -90,6 +90,7 @@ static void dump_bb(MVMThreadContext *tc, DumpStr *ds, MVMSpeshGraph *g, MVMSpes
     MVMSpeshIns *cur_ins;
     MVMint64     i;
     MVMint32     size = 0;
+    MVMuint32    line_number;
```
jnthn: What warning does this fix? Generally, it's better to declare variables in the narrowest scope they'll be used, and so this change would seem to make things worse.

AlexDaniel:
@DataKinds any updates? :)

PR #1148, opened 2019-07-22
Init debug socket for Windows

Altai-man:
To avoid error during socket creation, we need to init it on Windows. I suspect we do it for user code, but debugserver does not have strong coverage and was not working on Windows until this fix.

### DIFF
--- a/src/debug/debugserver.c
+++ b/src/debug/debugserver.c
@@ -2565,11 +2565,24 @@ static void debugserver_worker(MVMThreadContext *tc, MVMCallsite *callsite, MVMR
     vm->debugserver->thread_id = tc->thread_obj->body.thread_id;
 
     {
+#ifdef _WIN32
+        WORD wVersionRequested;
+        WSADATA wsaData;
+#endif
         char portstr[16];
         struct addrinfo *res;
         int error;
 
         snprintf(portstr, 16, "%"PRIu64, port);
+
+#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


ugexe:
See https://github.com/libuv/libuv/commit/af31d01492b2714a56a9b86943e8aeb9711ee019 -- is something like this also needed?

Altai-man:
Looking there: I don't see why it should. As I understand it, related libuv commit refers to a fact that different parts of libuv will be aborted if WSAStartup fails, which is not desired in safe mode.
However, our debugserver must have a socket to talk with the client protocol, it is pointless to have a debugserver thread that can't speak with the process, hence the panic.
Of course, if in future we'll get support of something other than sockets to communicate, the situation will be different, but then the code must be different too.

On the other hand I am not really a C pro, so I expect for someone more experienced with libuv to look at this commit.

Also, I will build a full nqp/rakudo stack on Windows, not only moarvm, after getting some sleep and will see if this truly enables debugging features for us or just moves failure point forward.

jnthn:
> However, our debugserver must have a socket to talk with the client protocol, it is pointless to have a debugserver thread that can't speak with the process, hence the panic.

Yes, agree with this. This code path is only followed if we're starting in debug mode. I can imagine it's important if people write applications that don't do networking stuff and should be able to work in safe mode, so it's good libuv does that, but I don't think it's applicable in our situation.

PR #1150, opened 2019-07-28
Generate conversion table from a source file

ZhongnianTao:
This PR updates the conversion table of GB2312 with unicode.org-mappings/EASTASIA/GB/GB2312.TXT from https://haible.de/bruno/charsets/conversion-tables/GB2312.html, instead of using python libraries to generate the table.


AlexDaniel:
Two ideas:
1. Nothing in the source file indicates how this table was generated. I think it would be helpful to leave a few comments for the next person who has to work on this.
2. I see at least four relatively large sections of just zeros (I counted ≈3534, ≈1216, ≈3553, ≈12179 zeros). Given that these are int32, that's ≈82 kB of wasted memory for an encoding that most programs won't even use. Wouldn't it be better to remove large sections with zeros and instead shift the index accordingly on the fly?

ZhongnianTao:
Comments about how the tables are generated have added.
The four largest sections of zeros are removed, and the function is adjusted, which saves memory for 41131 int32 (~161KB). The remaining sections are small (333, 316, 295, 246, and 234) so further saving memory might not compensate increased time coefficient (if-else branches) of the conversion function.

AlexDaniel:
Awesome!

samcv:
Looks good. Merging this. Rakudo and roast have also had their PR merged.

PR #1152, opened 2019-07-31
[IP6NS Grant] Retry connecting over IPv4 when IPv6 connections fail in certain cases

Kaiepi:
On systems where IPv6 is enabled, but misconfigured, hostnames can
resolve to IPv6 addresses when AF_UNSPEC is specified, but actually
trying to connect with them returns an error of some sort. Don't throw
when this happens; instead, keep track of whether or not IPv6 actually
works in the MoarVM instance, and when it doesn't, resolve the given
hostname to an IPv4 address before reattempting to connect, if
possible.

This comes with some minor refactors to make being able to check errors
like this possible in the future if needed when binding sockets. That I
currently have no way of testing since I don't own any domain names.

This alongside https://github.com/MoarVM/MoarVM/pull/1126 fixes https://github.com/rakudo/rakudo/issues/3007 and https://github.com/rakudo/rakudo/issues/2162, but the two cannot be merged at the same time as either one will need to be rebased after the other gets merged. The other pullreq should be merged first.

Passes nqp's `make test` and Rakudo's `make test` and `make spectest`.

### DIFF
--- a/src/core/instance.h
+++ b/src/core/instance.h
@@ -373,6 +373,11 @@ struct MVMInstance {
     /* Flags indicating the signals available on the host system */
     MVMuint64       valid_sigs;
 
+    /* Whether or not IPv6 actually works. This is necessary for systems where
+     * IPv6 is enabled, but misconfigured. */
+    uv_mutex_t      mutex_ipv6;
+    MVMuint8        ipv6;
+
     /************************************************************************
      * Caching and interning
      ************************************************************************/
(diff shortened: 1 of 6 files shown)


Kaiepi:
Rebased.

I tested various ways to break IPv6 over loopback, but none of them actually ended up making binding sockets return an error (I did find some new ways for connections to do so though). There's nothing left for me to do here before review.

Kaiepi:
This is the wrong approach to fixing the issues. There are multiple, more general problems with how sockets and host name resolution are being handled combined that are causing them, not just one specific to IPv6.

PR #1153, opened 2019-08-05
Fix bogus "Decoder may not be used concurrently" after exceptions

niner:
If stream decoding throws an exception (like on bad UTF-8) the deocder's
in_use flag would stay set, preventing any further use of the decoder.
This can lead to great action-at-a-distance effects.

Fix by extending the MVM_tc_(set|release)_ex_release_mutex mechanism to
support such flags in addition to fully blown mutexes. This way we can
keep the light weight in_use flag for the cost of a single bit test and
branch when throwing exceptions.

### DIFF
--- a/src/6model/reprs/Decoder.c
+++ b/src/6model/reprs/Decoder.c
@@ -112,11 +112,13 @@ void MVM_decoder_ensure_decoder(MVMThreadContext *tc, MVMObject *decoder, const 
 static void enter_single_user(MVMThreadContext *tc, MVMDecoder *decoder) {
     if (!MVM_trycas(&(decoder->body.in_use), 0, 1))
        MVM_exception_throw_adhoc(tc, "Decoder may not be used concurrently");
+    MVM_tc_set_ex_release_atomic(tc, &(decoder->body.in_use));
 }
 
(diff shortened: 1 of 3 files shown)


src/6model/reprs/Decoder.c
```diff
@@ -112,11 +112,13 @@ void MVM_decoder_ensure_decoder(MVMThreadContext *tc, MVMObject *decoder, const
 static void enter_single_user(MVMThreadContext *tc, MVMDecoder *decoder) {
     if (!MVM_trycas(&(decoder->body.in_use), 0, 1))
        MVM_exception_throw_adhoc(tc, "Decoder may not be used concurrently");
+    MVM_tc_set_ex_release_mutex(tc, (uv_mutex_t *)((uintptr_t)&(decoder->body.in_use) | 1));
```
jnthn: I think it'd be cleaner to have a function `MVM_tc_set_ex_release_atomic` that takes an `AO_t` and does the `| 1` and casting. That way, we can more easily use it from other places we might need it.

niner:
Added the suggested `MVM_tc_set_ex_release_atomic` and merged. Thanks for the review!

PR #1158, opened 2019-08-15
Revamp DLL subsystem

cygx:
The DLL subsystem (namespace `MVM_dll_*`) was never integrated into NativeCall and only ever used to load extops.

This pull request kills off the obsolete `DLLSym` repr and re-purposes some ops that were never exposed to NQP, changing

    freelib             r(str)
    findsym             w(obj) r(str) r(str)
    dropsym             r(obj)

into

    droplib             r(obj)
    getlib              w(obj) r(str) r(obj)
    findsym             w(int64) r(obj) r(str)

As stated, these ops appear to have been unused, so I don't think we need to deprecate the former first?

A corresponding Perl6-level interface could look like this:

    class DLL is repr<CPointer> {
        method new { !!! }

        multi method load(DLL:U: Str:D $name, Str:D $lib --> DLL) {
            nqp::loadlib($name, $lib);
            nqp::getlib($name, self);
        }

        multi method load(DLL:U: Str:D $lib --> DLL) {
            self.load($lib, $lib);
        }

        multi method load(DLL:U: IO::Path:D $path --> DLL) {
            my $abspath := $path.absolute;
            self.load($abspath, $abspath);
        }

        multi method get(DLL:U: Str:D $name --> DLL) {
            nqp::getlib($name, self);
        }

        multi method get(DLL:U: Whatever --> DLL) {
            nqp::box_i(0, self);
        }

        method drop(DLL:D: --> Nil) {
            nqp::droplib(self);
        }

        method findsym(DLL:D: Str:D $symbol --> Pointer) {
            my int $address = nqp::findsym(self, $symbol);
            $address ?? Pointer.new($address) !! Nil;
        }

        multi method gist(DLL:D: --> Str) {
            my int $i = nqp::unbox_i(self);
            "DLL|{$i ?? $i.base(16) !! '*'}";
        }
    }

Example usage:

    my $lib = DLL.load('moar.dll');
    LEAVE $lib.drop;

    say $lib.findsym('MVM_vm_create_instance');

or

    my $sym = DLL.get(*).findsym('puts');
    nativecast(:(Str --> int32), $sym).('hello world');

That's working code, assuming you build NQP with

    QAST::MASTOperations.add_core_moarop_mapping('loadlib', 'loadlib');
    QAST::MASTOperations.add_core_moarop_mapping('droplib', 'droplib');
    QAST::MASTOperations.add_core_moarop_mapping('getlib', 'getlib');
    QAST::MASTOperations.add_core_moarop_mapping('findsym', 'findsym');

added to `vm/moar/QAST/QASTOperationsMAST.nqp`.

This seems to be a useful thing to have if we want to fatpack bytecode/resources into executables.


cygx:
Turns out this probably should go through the whole deprecation thing - otherwise, trying to run new bytecode with old VM versions will give unhelpful validation errors...

PR #1161, opened 2019-08-23
Pull out top-level API into standalone header

cygx:
Our installed headers are broken: While one would expect something like 

    #include <moar/moar.h>

to work, it does not as we basically dump all out header files into the public include directory without any care for relative paths.

This will take some time to fix, but as a stopgap measure, this pull request adds `api.h` as a standalone header declaring the functions in the `MVM_vm_*` namespace.

The signature of the newly introduced `MVM_vm_run_bytecode()` function has been changed to no longer refer to any `MVM*` types (besides `MVMInstance`).

### DIFF
--- a/src/api.h
+++ b/src/api.h
@@ -1,0 +1,40 @@
+#ifndef MVM_API_H
+#define MVM_API_H
+
+#include <stddef.h>
+
+#ifndef MVM_TYPES_INCLUDED
+#include <moar/types.h>
+#endif
+
+#ifndef MVM_PUBLIC
+#ifdef _WIN32
+#define MVM_PUBLIC __declspec(dllimport)
+#else
+#define MVM_PUBLIC
+#endif
+#endif
+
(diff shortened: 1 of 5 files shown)


src/api.h
```diff
@@ -0,0 +1,40 @@
+#ifndef MVM_API_H
+#define MVM_API_H
+
+#include <stddef.h>
+
+#ifndef MVM_TYPES_INCLUDED
```
jnthn: Any reason not to check this in types.h?

cygx: Not really - I could put a standard inclusion guard

    #ifndef MVM_TYPES_H
    #define MVM_TYPES_H
    // ...
    #endif

around `types.h`. I didn't because all other checks were done in `api.h`, but I've no preference either way.

jnthn: Yes, I'd prefer it that way.

niner:
@cygx you still working on this?

PR #1166, opened 2019-08-27
[IP6NS Grant] Improve hostname resolution

Kaiepi:
This makes a few different changes to hostname resolution. First, it makes it so when resolving hostnames on systems where no IPv6 addresses actually exist on the interface used, an IPv6 address will never be returned by `MVM_io_resolve_host_name`, and likewise for IPv4. Second, rather than `MVM_io_resolve_host_name` only returning the first address returned by `getaddrinfo`, it returns all of them instead, and functions using its return value now attempt to use all of them before they actually throw. Finally, `MVM_io_resolve_host_name` now takes parameters for socket types, protocols, and passivity, which makes it so it doesn't return addresses that can't actually be used when those are specified.

This is currently broken, as the `uv_udp_t` handle bound when `MVM_io_async_socket_udp` gets called is zeroed out in `write_setup`. Interestingly, if I don't set `handle->data` to `NULL`, the `SocketSetupInfo` struct there ends up being the only struct member not to be zeroed out. Maybe the GC's involved getting involved here somehow?

Fixes rakudo/rakudo#3007
Fixes rakudo/rakudo#2162

### DIFF
--- a/src/io/asyncsocketudp.c
+++ b/src/io/asyncsocketudp.c
@@ -526,6 +526,7 @@ typedef struct {
     uv_loop_t        *loop;
     MVMObject        *async_task;
     uv_udp_t         *handle;
+    int               bind;
     struct addrinfo  *records;
     struct addrinfo  *cur_record;
     MVMint64          flags;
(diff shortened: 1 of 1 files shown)


Kaiepi:
OK, this should be ready for review now

Kaiepi:
I can't find any memory leaks with OpenBSD's `malloc_gdump` anymore with the last commit (Valgrind complains about a missing syscall wrapper and bails with sockets tests on the only box I can run it on).

The last commit also includes some fixes to `MVMROOT`s in the async sockets files, since they weren't always being given every pointer known by whatever function it's used from that the GC keeps track of (hopefully I'm not on the wrong track here).

src/io/asyncsocket.c
```diff
@@ -31,124 +31,150 @@ static void free_on_close_cb(uv_handle_t *handle) {
...
-            res_buf->body.ssize    = buf->len;
-            res_buf->body.elems    = nread;
-            MVM_repr_push_o(tc, arr, (MVMObject *)res_buf);
-
+            MVMROOT(tc, arr, {
+                MVMObject *seq_boxed = MVM_repr_box_int(tc,
+                    tc->instance->boot_types.BOOTInt, ri->seq_number++);
+                MVM_repr_push_o(tc, arr, seq_boxed);
+                /* Produce a buffer and push it. */
+                MVMROOT(tc, seq_boxed, {
```
timo: since seq_boxed doesn't get used in the code after this point, and it's already pushed into the array which makes it reachable, this MVMROOT can be dropped

src/io/asyncsocket.c
```diff
@@ -199,19 +225,20 @@ static MVMAsyncTask * read_bytes(MVMThreadContext *tc, MVMOSHandle *h, MVMObject
...
     ri              = MVM_calloc(1, sizeof(ReadInfo));
     MVM_ASSIGN_REF(tc, &(task->common.header), ri->buf_type, buf_type);
     MVM_ASSIGN_REF(tc, &(task->common.header), ri->handle, h);
     task->body.data = ri;
+    task->body.ops  = &read_op_table;
 
     /* Hand the task off to the event loop. */
-    MVMROOT(tc, task, {
+    MVMROOT6(tc, h, queue, schedulee, buf_type, async_type, task, {
         MVM_io_eventloop_queue_work(tc, (MVMObject *)task);
```
timo: i would assume the first five arguments here are being kept alive by the caller of this function, and since the pointers don't get used after this point, they could perhaps be left unrooted here

PR #1168, opened 2019-08-27
change MVM_VECTOR growth factor from 2 to 1.5

rbnx:
1.5 is usually a better growth factor for dynamic arrays

### DIFF
--- a/src/core/vector.h
+++ b/src/core/vector.h
@@ -45,8 +45,8 @@
 #define MVM_VECTOR_ENSURE_SIZE(x, size) do {\
         size_t _s = (size); \
         if (_s >= (x ## _alloc)) {    \
-            size_t newsize = (x ## _alloc) * 2 + 2; \
-            while (_s >= newsize) newsize *= 2; \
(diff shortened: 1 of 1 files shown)


bdw:
@rbnx curious, why?

bdw:
@rbnx thank you, I learn something new every day. Travis and AppVeyor seem to hate it, though, but that may have been because of the recent perl5.20 thing. So I'll run it locally to verify.

bdw:
There seems to be an infinite loop bug here...

rbnx:
Oops that was happening here: `MVM_VECTOR_INIT(xx,0)`.
fixed it by making the default `newsize` 8 when `x_alloc` is 0

niner (changes requested):
Can you please extend the commit message to tell the gist of why 1.5 is a better number than 2 and a link to the original research? Any benchmarks of how much this helps would be greatly appreciated as well.

A commit message must always answer the question of why a change was made. Otherwise someone in the future will come across this piece of code and wonder why we don't just double the size and reduce the number of allocations that way.

PR #1169, opened 2019-08-27
Separate includes from submodules from system-wide includes

vrurg:
Currently configuration variable cincludes contains `-I` flags for both
includes from `3rdparty` for for those found with `pkg-config`. This is
not correct as `cincludes` can be used by `NQP` and `Rakudo` for their
C compilations. An example of a situation when this is necessary is
rakudo/rakudo#3136 where `ffi.h` cannot be found even though `moar` is
being compiled with it.


AlexDaniel:
Makes sense to me.

PR #1176, opened 2019-09-13
Deoptimization in the expression JIT compiler

bdw:
Adds deoptimization guard support in the expression JIT compiler.
To achieve this, I made the argument to deopt_one be the deopt_idx, using which we lookup the deopt_target, rather than using the deopt_target itself. This saves another search in the materialization logic. This makes the implementation of the `sp_guard` opcodes non-special as far as the template preprocessor is concerned, except that we do have to introduce a global sync.


jnthn:
Really nice set of improvements! :-)

PR #1184, opened 2019-09-29
Fix access to freed memory in resolve_using_guards

niner:
Commit acb04a448bd7b067f1a0943b6fd521e21057c96e fixed outdated pointers in the
guard set, but it was still possible that the guard set was freed while still
in use by resolve_using_guards as the next GC safe point may occur in
evaluate_guards. Fix this by turning SpeshPluginState into a 6model object, so
we can let the GC figure out whether the state is still in use somewhere or not


niner:
Has been merged already

PR #1190, opened 2019-10-08
[IP6NS Grant] Close synchronous sockets properly

Kaiepi:
`close` should not be used to close sockets on Windows; that's meant for
file handles, not sockets.


Kaiepi:
This PR is a bit premature. I need to mull over how sync sockets get closed (or shutdown, eventually) more before making any changes.

Issue #1172, opened 2019-09-10
possible uid/gid confusion


robertlemmen:
while trying to understand build failures with libuv 1.30.1, I came across this (unrelated to the failure) code in src/io/fileops.c:
```
#define FILE_IS(name, rwx) \
    MVMint64 MVM_file_is ## name (MVMThreadContext *tc, MVMString *filename, MVMint32 use_lstat) { \
        if (!MVM_file_exists(tc, filename, use_lstat)) \
            return 0; \
        else { \
            uv_stat_t statbuf = file_info(tc, filename, use_lstat); \
            MVMint64 r = (statbuf.st_mode & S_I ## rwx ## OTH) \
                      || (statbuf.st_uid == geteuid() && (statbuf.st_mode & S_I ## rwx ## USR)) \
                      || (statbuf.st_uid == getegid() && (statbuf.st_mode & S_I ## rwx ## GRP)); \
            return r ? 1 : 0; \
        } \
    }
```
I am pretty certain that the comparison with getegid() should be with statbuf.st_gid

patzim:
I honestly don't understand either the new or the old code.
What I understand the code does currently is compare the users group to the files group.

My understanding of groups in Linux:
- The user group is just the group new files are assigned when the user creates them
- To check whether a user has permissions to do something with a file one needs to compare the user to all the users that are in the files group

Am I totally confused?
@dogbert17, @robertlemmen

robertlemmen:
My understanding of how this works is slightly different, a simple example:

> ls -l
-rw-rw-r-- 1 robertle devgrp   213 Sep 22 19:14 Makefile

so this file is owned by user 'robertle' and group 'devgrp'. permissions for the user are 'rw-', for the group 'rw-' and for all other users 'r--'. so it would be writable by 'robertle', which is what the statbuf.st_uid == geteuid() line checks for.

it is also writable by all users in the 'devgrp' group, I suspect that's what the getegid line is trying to check, but it does compare to the user-owner of the file, which seems like a (now fixed) bug.

you are correct though that getegid() returns *one* group id, but the caller could be in multiple groups. so instead of getegid() we should use group_member() or something like that!

patzim:
@robertlemmen Can you have a look at whether PR #1193 fixed this issue completely?

robertlemmen:
the PR looks good to me, and fixes the one problem I could see. but I never had a full set of tests or any way to really verify the behavior...

PR #1193, opened 2019-10-14
Make MVM_file_is_rwx honor root and groups

### LINKED ISSUE #1179: File permission check ops ignore root
The `filewritable` OP currently checks the EUID and EGID to determine whether a file is writeable. It doesn't check whether the UID is 0 (we are root) though. So being root does not convince rakudo that we are allowed to write to a file currently. I think it should.

Documentation is silent as to what the file tests of `IO::Path` should actually check ([see here](https://docs.perl6.org/type/IO::Path#File_test_operators)).

As a point of reference the `-w` operator of Perl 5 does respect UID = 0.

The bit of implementation where this happens is: [fileops.c:255](https://github.com/MoarVM/MoarVM/blob/master/src/io/fileops.c#L255)

patzim:
Fixes #1179 and #1172 

I tested this with the attached not pretty Raku script. Previous Moar results in:

```
foo 000 user1:lonely --          user1 CAN'T read.      Raku agrees
foo 400 user1:lonely --          user1 can read.        Raku agrees
foo 000 user2:user1 --           user1 CAN'T read.      Raku agrees
foo 040 user2:user1 --           user1 can read.        Raku DISAGREES.
foo 000 user2:user1supp --       user1 CAN'T read.      Raku agrees
foo 040 user2:user1supp --       user1 can read.        Raku DISAGREES.
foo 000 user2:lonely --          user1 CAN'T read.      Raku agrees
foo 004 user2:lonely --          user1 can read.        Raku agrees
foo 000 user2:lonely --          root can read.         Raku DISAGREES.
```

With this PR applied it returns:

```
foo 000 user1:lonely --          user1 CAN'T read.      Raku agrees
foo 400 user1:lonely --          user1 can read.        Raku agrees
foo 000 user2:user1 --           user1 CAN'T read.      Raku agrees
foo 040 user2:user1 --           user1 can read.        Raku agrees
foo 000 user2:user1supp --       user1 CAN'T read.      Raku agrees
foo 040 user2:user1supp --       user1 can read.        Raku agrees
foo 000 user2:lonely --          user1 CAN'T read.      Raku agrees
foo 004 user2:lonely --          user1 can read.        Raku agrees
foo 000 user2:lonely --          root can read.         Raku agrees
```

```
shell('sudo rm foo 2>/dev/null').so;
shell('sudo userdel -r user1 2>/dev/null').so;
shell('sudo groupdel user1supp 2>/dev/null').so;
shell('sudo userdel -r user2 2>/dev/null').so;
shell('sudo groupdel user2supp 2>/dev/null').so;
shell('sudo groupdel lonely 2>/dev/null').so;

shell 'sudo groupadd lonely';
shell 'sudo groupadd user1supp';
shell 'sudo useradd -U -M -G user1supp user1';
shell 'sudo groupadd user2supp';
shell 'sudo useradd -U -M -G user2supp user2';


sub check-access($user, $perms, $owners) {
    shell 'sudo touch foo';
    shell 'sudo sh -c "echo word > foo"';
    shell "sudo chown $owners foo";
    shell "sudo chmod $perms foo";

    my $p1 = shell("su $user -c 'cat foo' 2>/dev/null", :out);
    my $ok = $p1.so && ($p1.out.get eq 'word');

    my $p2 = shell("su $user -c '$*EXECUTABLE -e \"say «foo».IO.r\"'", :out);
    my $ok2 = trim($p2.out.get) eq 'True';

    say "foo $perms $owners --   \t $user " ~ ($ok ?? 'can' !! 'CAN\'T') ~ " read. \tRaku " ~ ($ok2 == $ok ??     +'agrees' !! 'DISAGREES.');
    shell 'sudo rm foo';
}

check-access 'user1', '000', 'user1:lonely';    # nok
check-access 'user1', '400', 'user1:lonely';    # ok

check-access 'user1', '000', 'user2:user1';     # nok
check-access 'user1', '040', 'user2:user1';     # ok

check-access 'user1', '000', 'user2:user1supp'; # nok
check-access 'user1', '040', 'user2:user1supp'; # ok

check-access 'user1', '000', 'user2:lonely';    # nok
check-access 'user1', '004', 'user2:lonely';    # ok

check-access 'root',  '000', 'user2:lonely';    # ok

shell 'sudo userdel -r user1 2>/dev/null';
shell 'sudo groupdel user1supp';
shell 'sudo userdel -r user2 2>/dev/null';
shell 'sudo groupdel user2supp';
shell 'sudo groupdel lonely'
```

### DIFF
--- a/src/io/fileops.c
+++ b/src/io/fileops.c
@@ -246,6 +246,35 @@ MVMint64 MVM_file_isexecutable(MVMThreadContext *tc, MVMString *filename, MVMint
     }
 }
 #else
+
+static int are_we_group_member(MVMThreadContext *tc, gid_t group) {
+    int len;
+    gid_t *gids;
+    int res;
+    int i;
+    /* Check the user group. */
+    if (getegid() == group)
+        return 1;
+    /* Check the supplementary groups. */
+    len  = getgroups(0, NULL);
+    if (len == 0)
+        return 0;
+    gids = MVM_malloc(len * sizeof(gid_t));
+    res = getgroups(len, gids);
+    if (res < 0) {
+        MVM_free(gids);
+        MVM_exception_throw_adhoc(tc, "Failed to retrieve groups: %s", strerror(errno));
+    }
+    res = 0;
+    for (i = 0; i < len; i++) {
+        if (gids[i] == group) {
+            res = 1;
+            break;
+        }
+    }
+    MVM_free(gids);
+    return res;
+}
 #define FILE_IS(name, rwx) \
     MVMint64 MVM_file_is ## name (MVMThreadContext *tc, MVMString *filename, MVMint32 use_lstat) { \
         if (!MVM_file_exists(tc, filename, use_lstat)) \
@@ -254,7 +283,8 @@ MVMint64 MVM_file_isexecutable(MVMThreadContext *tc, MVMString *filename, MVMint
             uv_stat_t statbuf = file_info(tc, filename, use_lstat); \
             MVMint64 r = (statbuf.st_mode & S_I ## rwx ## OTH) \
                       || (statbuf.st_uid == geteuid() && (statbuf.st_mode & S_I ## rwx ## USR)) \
-                      || (statbuf.st_gid == getegid() && (statbuf.st_mode & S_I ## rwx ## GRP)); \
+                      || (geteuid() == 0) \
+                      || (are_we_group_member(tc, statbuf.st_gid) && (statbuf.st_mode & S_I ## rwx ## GRP)); \
             return r ? 1 : 0; \
         } \
     }


src/io/fileops.c
```diff
@@ -246,6 +246,32 @@ MVMint64 MVM_file_isexecutable(MVMThreadContext *tc, MVMString *filename, MVMint
     }
 }
 #else
+
+static int are_we_group_member(gid_t group) {
+    int len;
+    gid_t *gids = 0;
```
jnthn: `NULL`

src/io/fileops.c
```diff
@@ -246,6 +246,32 @@ MVMint64 MVM_file_isexecutable(MVMThreadContext *tc, MVMString *filename, MVMint
...
+    int res;
+    int i;
+    // Check the user group.
+    if (getegid() == group) {
+        return 1;
+    }
+    // Check the supplementary groups.
+    do {
+        len  = getgroups(0, 0);
```
jnthn: First arg should be `NULL`.

src/io/fileops.c
```diff
@@ -246,6 +246,32 @@ MVMint64 MVM_file_isexecutable(MVMThreadContext *tc, MVMString *filename, MVMint
...
+        gids = MVM_realloc(gids, len * sizeof(gid_t));
+        res  = getgroups(len, gids);
+    } while (res < -1);
```
jnthn: Why the loop? The spec I found for `getgroups` says:

> If size is zero, list is not modified, but the total number of
> supplementary group IDs for the process is returned.  This allows the
> caller to determine the size of a dynamically allocated list to be
> used in a further call to getgroups().

Which would imply one call to get the length and another to populate the list would suffice. No loop = no risk of loop gone infinite by accident.

patzim: The loop was my solution to the question of what to do with the return value. Should I just ignore the return value (according to doc the function can error when given an invalid pointer or `len` being to small) or somehow try to handle the error? How could a sane error handling look?

jnthn: > How could a sane error handling look?

Throw an exception (`MVM_exception_throw_adhoc`).

> the function can error when given an invalid pointer or len being to small

Given the pointer came straight out of `MVM_realloc`, then we can trust that. Given that we asked for the length above, then `len` being too small can only occur upon a race; I think we can just disregard such cases (e.g. ignore errors that are about the length, and throw on any others).

Also, if the first call returns `0`, then we can exit swiftly without the whole allocate step.

patzim: Thanks for the detailed explanation! I changed it accordingly.

src/io/fileops.c
patzim: Done.

src/io/fileops.c
```diff
@@ -246,6 +246,35 @@ MVMint64 MVM_file_isexecutable(MVMThreadContext *tc, MVMString *filename, MVMint
...
+        return 1;
+    // Check the supplementary groups.
+    len  = getgroups(0, NULL);
+    if (len == 0)
+        return 0;
+    gids = MVM_realloc(gids, len * sizeof(gid_t));
+    res = getgroups(len, gids);
+    if (res < 0) {
+        MVM_free(gids);
+        MVM_exception_throw_adhoc(tc, "Failed to retrieve groups: %s", res);
```
jnthn: `res` is an `int`, so using it with `%s` will promptly segfault. I think you need to call `strerror(res)` to turn it into a string.

src/io/fileops.c
```diff
@@ -246,6 +246,35 @@ MVMint64 MVM_file_isexecutable(MVMThreadContext *tc, MVMString *filename, MVMint
...
+    int res;
+    int i;
+    // Check the user group.
+    if (getegid() == group)
+        return 1;
+    // Check the supplementary groups.
+    len  = getgroups(0, NULL);
+    if (len == 0)
+        return 0;
+    gids = MVM_realloc(gids, len * sizeof(gid_t));
```
jnthn: This can simply be `MVM_malloc`, which is an iota cheaper.

src/io/fileops.c
jnthn: Oh, sorry, I should read docs more carefully. It's actually `errno` that you need to pass to `strerror`. :-)

src/io/fileops.c
```diff
@@ -246,6 +246,35 @@ MVMint64 MVM_file_isexecutable(MVMThreadContext *tc, MVMString *filename, MVMint
     }
 }
 #else
+
+static int are_we_group_member(MVMThreadContext *tc, gid_t group) {
+    int len;
+    gid_t *gids = NULL;
+    int res;
+    int i;
+    // Check the user group.
```
MasterDuke17: I thought these style comments actually caused problems in CI somewhere, but regardless, I believe there is at least a convention to only use ‘/* */‘ comments.

PR #1196, opened 2019-10-16
Show correct values in encoding errors

MasterDuke17:
Fixes the valgrind errors @dogbert17++ found in t/spec/S17-procasync/encoding.t (https://gist.github.com/dogbert17/e4cb7b4d003f9f7177a70f73a49c7727). Also show chars near the bad bytes in another encoding error path.

### DIFF
--- a/src/strings/utf8.c
+++ b/src/strings/utf8.c
@@ -240,15 +240,15 @@ MVMString * MVM_string_utf8_decode(MVMThreadContext *tc, const MVMObject *result
                     MVM_free(buffer);
                     if (error_pos >= 3) {
                         unsigned char a = utf8[error_pos - 2], b = utf8[error_pos - 1], c = utf8[error_pos];
-                        MVM_exception_throw_adhoc(tc, "Malformed UTF-8 near bytes %02x %02x %02x at line %u col %u", a, b, c, line, col);
+                        MVM_exception_throw_adhoc(tc, "Malformed UTF-8 near bytes %02hhx %02hhx %02hhx at line %u col %u", a, b, c, line, col);
                     }
                     else if (error_pos == 2) {
                         unsigned char a = utf8[error_pos - 1], b = utf8[error_pos];
-                        MVM_exception_throw_adhoc(tc, "Malformed UTF-8 near bytes %02x %02x at line %u col %u", a, b, line, col);
+                        MVM_exception_throw_adhoc(tc, "Malformed UTF-8 near bytes %02hhx %02hhx at line %u col %u", a, b, line, col);
                     }
                     else if (error_pos == 1) {
                         unsigned char a = utf8[error_pos];
-                        MVM_exception_throw_adhoc(tc, "Malformed UTF-8 near byte %02x at line %u col %u", a, line, col);
+                        MVM_exception_throw_adhoc(tc, "Malformed UTF-8 near byte %02hhx at line %u col %u", a, line, col);
                     }
                     else {
                         MVM_exception_throw_adhoc(tc, "Malformed UTF-8 at line %u col %u", line, col);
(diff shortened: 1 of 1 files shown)


src/strings/utf8.c
```diff
@@ -384,17 +384,18 @@ MVMuint32 MVM_string_utf8_decodestream(MVMThreadContext *tc, MVMDecodeStream *ds
                     break;
                 }
                 case UTF8_REJECT: {
-                    char *waste[] = { (char *)buffer, NULL };
-                    if (bufsize >= 3) {
-                        MVMGrapheme32 a = buffer[pos - 2], b = buffer[pos - 1], c = buffer[pos];
+                    MVMint32 error_pos = pos - 1;
+                    char *waste[] = { (char *)bytes, NULL };
+                    if (error_pos >= 3) {
+                        MVMGrapheme32 a = bytes[error_pos - 2], b = bytes[error_pos - 1], c = bytes[error_pos];
```
niner: Why MVMGrapheme32? char and MVMGrapheme32 are both signed, so converting the value will sign extend, i.e. every char with the high bit set will be printed as e.g. ffffff80.

src/strings/utf8.c
```diff
@@ -384,17 +384,18 @@ MVMuint32 MVM_string_utf8_decodestream(MVMThreadContext *tc, MVMDecodeStream *ds
...
                         MVM_exception_throw_adhoc_free(tc, waste, "Malformed UTF-8 near bytes %02x %02x %02x", a, b, c);
```
niner: To avoid the mentioned sign extension, the format specifier should be %02hhx:
```
hh      Specifies that a following d, i, o, u, x, or X conversion specifier applies to a signed char or unsigned char argument  (the  argument
               will  have  been promoted according to the integer promotions, but its value shall be converted to signed char or unsigned char before
               printing);
```

src/strings/utf8.c
```diff
@@ -384,17 +384,18 @@ MVMuint32 MVM_string_utf8_decodestream(MVMThreadContext *tc, MVMDecodeStream *ds
                     break;
                 }
                 case UTF8_REJECT: {
-                    char *waste[] = { (char *)buffer, NULL };
-                    if (bufsize >= 3) {
-                        MVMGrapheme32 a = buffer[pos - 2], b = buffer[pos - 1], c = buffer[pos];
+                    MVMint32 error_pos = pos - 1;
+                    char *waste[] = { (char *)bytes, NULL };
```
niner: `bytes` is already declared as `char*`, no need to type cast it here.

src/strings/utf8.c
```diff
@@ -448,17 +449,18 @@ MVMuint32 MVM_string_utf8_decodestream(MVMThreadContext *tc, MVMDecodeStream *ds
                     break;
                 }
                 case UTF8_REJECT: {
-                    char *waste[] = { (char *)buffer, NULL };
-                    if (bufsize >= 3) {
-                        MVMGrapheme32 a = buffer[pos - 2], b = buffer[pos - 1], c = buffer[pos];
+                    MVMint32 error_pos = pos - 1;
```
niner: This is the exact same code as above, so please move it to its own function. As this is an error path anyway, the call overhead will not matter and the whole decodestream function is already quite long.

MasterDuke17:
Whoops, those MVMGrapheme32s are left-over from when I was extracting from
`buffer`, not `bytes`.

Should I change to `%02hhx` or just assign into `unsigned char` as I did in
the top-most code?

niner:
> Should I change to `%02hhx` or just assign into `unsigned char` as I did in
> the top-most code?

I think both are needed to make it really correct (and not just work) and to 
unambiguously communicate what's supposed to happen to the reader of the code.

MasterDuke17:
@niner how's that look?

PR #1197, opened 2019-10-17
Update to libuv version 1.33.0

dogbert17:
Clean spectest on Linux. According to tests by vrurg++ it also fixes
M#1178 a Mac specific problem wrt file change notifications.


jnthn:
Thanks!

PR #1198, opened 2019-10-18
Add libzstd-dev to travis...

MasterDuke17:
so new heap snapshot code is tested.


MasterDuke17:
Huh, it looks like it has zstd, but maybe its version is too old? @timo

PR #1201, opened 2019-10-20
Avoid using preprocessor inside MVMROOT

### LINKED ISSUE #1200: MoarVM broken on windows
In https://github.com/MoarVM/MoarVM/commit/c61236f96ec6952e7c0a2f58e35c1a5b9db2b189 the following is now inside an `MVMROOT`
https://github.com/MoarVM/MoarVM/blob/c61236f96ec6952e7c0a2f58e35c1a5b9db2b189/src/spesh/plugin.c#L340-L344

This breaks windows in the following way -- https://ci.appveyor.com/project/samcv/moarvm-35u74/builds/28238735/job/ckp9y0ugrva7i42n#L368

ugexe:
MSVC appears to have a problem with macro expansion when using a preprocessor inside a MVMROOT.

Resolves #1200


niner:
This PR would have put the MVMROOT in the wrong place. I've pushed 0f0e6d3a27c03c6d2761e76351d56c92db80c70c which should fix the issue

PR #1202, opened 2019-10-20
Fix obscure deadlock

ugexe:


ugexe:
I have no idea why this fixes https://github.com/rakudo/rakudo/issues/3157 -- I discovered using `end_time` was causing the deadlock and ended up with this workaround/solution.

PR #1203, opened 2019-10-23
[WIP] Sample code for stringification and printing of internal structures

vrurg:
While trying to resolve rakudo/rakudo#2897 I stumbled upon the necessity of tracing context serialization. Unfortunately, it is virtually impossible to distinguish one frame from another while debugging based on their objects in memory. `MVMString` complexity contributes to the problems of finding frame name when it exists and checking what lexicals frame contains.

What I'm trying to implement is support for stringification of objects from inside of MoarVM itself. The code is currently in 'skeleton' state, unstructured and doesn't even compile. It is purely for demoing of what approximately I'm trying to achieve. 

_NOTE_ There is some garbage left from very early experiments. It's too late to clean it up, so I leave it for a while. It includes `dump_ctx_info` function and declarations of two `*_string_table` functions.

# Debugging 

More or less centralized support for debug prints. Debug would be compiled in with `MVM_DEBUG` defined. Debug prints are activated if either of three is true:

- global debug is turned on
- thread-level debug is turned on
- a dedicated environment variable is set.

For the first two it could be useful to provide NQP ops to turn debugging on/off for the purpose of tracing specific areas of NQP code from MoarVM point of view.

# Output

- support for `printf` variants which produce `MVMString`
- support for printing into a file handle
- `MVM_string_note` – similar for `MVM_string_say`

# Stringification

Perhaps it should be part of REPR.

### DIFF
--- a/build/Makefile.in
+++ b/build/Makefile.in
@@ -113,6 +113,7 @@ OBJECTS1 = src/core/callsite@obj@ \
           src/core/fixedsizealloc@obj@ \
           src/core/regionalloc@obj@ \
           src/debug/debugserver@obj@ \
+          src/debug/stringify@obj@ \
           src/gen/config@obj@ \
           src/gc/orchestrate@obj@ \
           src/gc/allocation@obj@ \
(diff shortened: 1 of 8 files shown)


vrurg:
Closing by after some [IRC discussion](https://colabti.org/irclogger/irclogger_log/moarvm?date=2019-10-23#l49).

PR #1207, opened 2019-10-24
Revert "Merge pull request #1166 from Kaiepi/ipv6-unsupported"

Kaiepi:
While only the change to make all addresses get used when connecting or
binding sockets needs to be reverted from this, not every commit had
only one change included with it, making it impossible to tell if it had
truly been reverted or not. They will be remade later.

This reverts commit 39f5774386bb889050d48c17c8e8a053655fa623, reversing
changes made to 01f85c8bf833785bb4bd1f965bb717ca84fb045f.

### DIFF
--- a/src/io/asyncsocket.c
+++ b/src/io/asyncsocket.c
@@ -8,12 +8,11 @@ typedef struct {
 
 /* Info we convey about a read task. */
 typedef struct {
-    MVMThreadContext *tc;
-    int               work_idx;
     MVMOSHandle      *handle;
     MVMObject        *buf_type;
     int               seq_number;
-    int               error;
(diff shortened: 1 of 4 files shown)


AlexDaniel:
Currently it doesn't pass CI tests. 
```
src\io\asyncsocketudp.c(557) : error C2039: 'async_task' : is not a member of 'SocketSetupInfo'
         src\io\asyncsocketudp.c(487) : see declaration of 'SocketSetupInfo'
```

Kaiepi:
This now passes `make spectest`.

AlexDaniel:
Should be good now, I think!

Thanks!

PR #1209, opened 2019-10-25
Fix run-away CORE context on closures

vrurg:
The problem was caused by the fact that serialization strips off
contexts without static code effectively serializing only those
belonging to the closure's compunit. Upon deserialization only those
serialized contexts are restores and bound to the closure being
deserialized resulting in closures loosing their CORE outers.

To fix the problem I used the fact that new contexts are created as
clones of static code which actually has all its outers stacked up
properly. Therefore, if a context without deserialized outer is
encountered, I check if it's source static code has one and if it does
then use that outer for the new context.

Fixes rakudo/rakudo#2897 and a number of similar issues.


vrurg:
The most time-expensive fix in my life. Three weeks per three lines...

diakopter:
impressive

vrurg:
Yep. I knew about a couple, but the search came up with the above list. Not sure if all are found.

PR #1214, opened 2019-11-25
[IP6NS Grant] Clean up and make MVM_io_resolve_host_name more portable

Kaiepi:
How `MVM_io_resolve_host_name` handles creating addresses for UNIX sockets would only work properly on Linux. Make how it does this portable, and while we're at it, make it use less code to do the same work.

This was part of the changes to DNS resolution that were reverted.


src/io/syncsocket.c
```diff
@@ -306,41 +293,57 @@ struct sockaddr * MVM_io_resolve_host_name(MVMThreadContext *tc, MVMString *host
...
-    hints.ai_addrlen = 0;
-    hints.ai_addr = NULL;
-    hints.ai_canonname = NULL;
-    hints.ai_next = NULL;
-
     snprintf(port_cstr, 8, "%d", (int)port);
 
     MVM_gc_mark_thread_blocked(tc);
-    error = getaddrinfo(host_cstr, port_cstr, &hints, &result);
+    e = getaddrinfo(host_cstr, port_cstr, &hints, &result);
```
niner: Why the rename of the variable? Seems to make the code less readable for no good reason.

src/io/syncsocket.c
```diff
@@ -306,41 +293,57 @@ struct sockaddr * MVM_io_resolve_host_name(MVMThreadContext *tc, MVMString *host
...
+            MVM_free(host_cstr);
+            MVM_exception_throw_adhoc(tc, "UNIX sockets are not supported by MoarVM on this platform");
+#else
+            static const size_t MAX_SUN_LEN = sizeof(((struct sockaddr_un *)NULL)->sun_path);
+            size_t              sun_len     = strnlen(host_cstr, MAX_SUN_LEN);
+
+            if (sun_len >= MAX_SUN_LEN) {
+                char *waste[] = { host_cstr, NULL };
+                MVM_exception_throw_adhoc_free(
+                    tc, waste, "Socket path '%s' is too long (max supported by this platform is %zu)",
```
niner: This can be a bit confusing. Your length check correctly takes the terminating NULL character into account (not counted by strnlen, but offset by using `>=` instead of `>` in the comparison), so it will show an error message citing the 108 character path as too long due to a max of 108 characters supported.

src/io/syncsocket.c
Kaiepi: I had the impression naming it `e` is better style when writing C, but checking how the rest of MoarVM names errors little else names it like this.

src/io/syncsocket.c
Kaiepi: Should it say something like `"Socket path '...' is too long (max length supported by this platform is 107)"` instead?

niner: Yes, since that is the actual number that we allow.

PR #1218, opened 2019-11-30
Fix sometimes uninitialized warning

MasterDuke17:


niner:
While this silences the compiler, do we actually know that we never end up with `graph` left at its default value? If yes, it would be nice to have a comment explaining why.

MasterDuke17:
@ZhongnianTao ping

MasterDuke17:
I no longer get warnings about this, from either gcc 10.1.0 or clang 10.0.0.

PR #1221, opened 2019-12-03
Switch our libtommath to upstream's 1.2.0 tag

MasterDuke17:
Their new release incorporates or removes the need for all the commits
we added to our fork, so the intent of this commit is to switch to
using just their release tag.

Assuming I did everything correctly, NQP built ok and passed `make m-test` and Rakudo built ok and passes `make m-test m-spectest`.

### DIFF
--- a/src/6model/reprs/P6bigint.c
+++ b/src/6model/reprs/P6bigint.c
@@ -12,7 +12,7 @@ static MVMint64 mp_get_int64(MVMThreadContext *tc, mp_int * a) {
 
     /* For 64-bit 2's complement numbers the positive max is 2**63-1, which is 63 bits,
      * but the negative max is -(2**63), which is 64 bits. */
-    if (MP_NEG == SIGN(a)) {
+    if (MP_NEG == a->sign) {
         if (bits > 64) {
             MVM_exception_throw_adhoc(tc, "Cannot unbox %d bit wide bigint into native integer", bits);
         }
@@ -31,7 +31,7 @@ static MVMint64 mp_get_int64(MVMThreadContext *tc, mp_int * a) {
         MVM_exception_throw_adhoc(tc, "Cannot unbox %d bit wide bigint into native integer", bits);
     }
 
-    return MP_NEG == SIGN(a) ? -res : res;
+    return MP_NEG == a->sign ? -res : res;
 }
 
 MVMint64 MVM_p6bigint_get_int64(MVMThreadContext *tc, MVMP6bigintBody *body) {
@@ -137,7 +137,7 @@ static MVMuint64 get_uint(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, 
     MVMP6bigintBody *body = (MVMP6bigintBody *)data;
     if (MVM_BIGINT_IS_BIG(body)) {
         mp_int *i = body->u.bigint;
-        if (MP_NEG == SIGN(i))
+        if (MP_NEG == i->sign)
             MVM_exception_throw_adhoc(tc, "Cannot unbox negative bigint into native unsigned integer");
         else
             return mp_get_uint64(tc, i);
--- a/src/6model/reprs/P6bigint.c
+++ b/src/6model/reprs/P6bigint.c
@@ -202,7 +202,7 @@ static void serialize(MVMThreadContext *tc, MVMSTable *st, void *data, MVMSerial
         MVMString *str;
         mp_radix_size(i, 10, &len);
         buf = (char *)MVM_malloc(len);
-        mp_toradix(i, buf, 10);
+        mp_to_decimal(i, buf, sizeof(buf));
 
         /* len - 1 because buf is \0-terminated */
         str = MVM_string_ascii_decode(tc, tc->instance->VMString, buf, len - 1);
--- a/src/6model/reprs/P6bigint.c
+++ b/src/6model/reprs/P6bigint.c
@@ -77,8 +77,11 @@ static void copy_to(MVMThreadContext *tc, MVMSTable *st, void *src, MVMObject *d
     MVMP6bigintBody *src_body = (MVMP6bigintBody *)src;
     MVMP6bigintBody *dest_body = (MVMP6bigintBody *)dest;
     if (MVM_BIGINT_IS_BIG(src_body)) {
+        mp_err err;
         dest_body->u.bigint = MVM_malloc(sizeof(mp_int));
-        mp_init_copy(dest_body->u.bigint, src_body->u.bigint);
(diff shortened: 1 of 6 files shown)


src/6model/reprs/P6bigint.c
```diff
@@ -196,13 +207,18 @@ static void gc_free(MVMThreadContext *tc, MVMObject *obj) {
...
         int len;
         char *buf;
         MVMString *str;
-        mp_radix_size(i, 10, &len);
+        if ((err = mp_radix_size(i, 10, &len)) != MP_OKAY) {
+            MVM_exception_throw_adhoc(tc, "Error calculating the size of a big integer: %s", mp_error_to_string(err));
+        }
         buf = (char *)MVM_malloc(len);
-        mp_toradix(i, buf, 10);
+        if ((err = mp_to_decimal(i, buf, sizeof(buf))) != MP_OKAY) {
```
sjaeckel: this should use `len` instead of `sizeof(buf)`

src/math/bigintops.c
```diff
@@ -370,8 +376,12 @@ MVMObject * MVM_bigint_##opname(MVMThreadContext *tc, MVMObject *result_type, MV
     ia = force_bigint(tc, ba, 0); \
     ib = force_bigint(tc, bb, 1); \
     ic = MVM_malloc(sizeof(mp_int)); \
-    mp_init(ic); \
-    mp_##opname(ia, ib, ic); \
+    if ((err = mp_init(ic)) != MP_OKAY) { \
+        MVM_exception_throw_adhoc(tc, "Error initializing a big integer: %s", mp_error_to_string(err)); \
+    } \
+    if ((err = mp_##opname(ia, ib, ic)) != MP_OKAY) { \
+        MVM_exception_throw_adhoc(tc, "Error performing ##opname with big integers: %s", mp_error_to_string(err)); \
```
sjaeckel: `##opname` won't be replaced by the preprocessor in a string, you should either do `"Error performing "##opname" with big integers: %s"` or `"Error performing %s with big integers: %s", ##opname ...` where I'd prefer the first

src/math/bigintops.c
```diff
@@ -339,10 +339,15 @@ MVMObject * MVM_bigint_##opname(MVMThreadContext *tc, MVMObject *result_type, MV
...
+            mp_err err; \
             mp_int *ia = ba->u.bigint; \
             mp_int *ib = MVM_malloc(sizeof(mp_int)); \
-            mp_init(ib); \
-            mp_##opname(ia, ib); \
+            if ((err = mp_init(ib)) != MP_OKAY) { \
+                MVM_exception_throw_adhoc(tc, "Error initializing a big integer: %s", mp_error_to_string(err)); \
+            } \
+            if ((err = mp_##opname(ia, ib)) != MP_OKAY) { \
+                MVM_exception_throw_adhoc(tc, "Error performing ##opname with a big integer: %s", mp_error_to_string(err)); \
```
sjaeckel: same as below

src/math/bigintops.c
```diff
@@ -380,12 +390,17 @@ MVMObject * MVM_bigint_##opname(MVMThreadContext *tc, MVMObject *result_type, MV
...
     ia = force_bigint(tc, ba, 0); \
     ib = force_bigint(tc, bb, 1); \
     ic = MVM_malloc(sizeof(mp_int)); \
-    mp_init(ic); \
-    mp_##opname(ia, ib, ic); \
+    if ((err = mp_init(ic)) != MP_OKAY) { \
+        MVM_exception_throw_adhoc(tc, "Error initializing a big integer: %s", mp_error_to_string(err)); \
+    } \
+    if ((err = mp_##opname(ia, ib, ic)) != MP_OKAY) { \
+        MVM_exception_throw_adhoc(tc, "Error performing ##opname with big integers: %s", mp_error_to_string(err)); \
```
sjaeckel: same as above

src/math/bigintops.c
```diff
@@ -405,8 +421,12 @@ MVMObject * MVM_bigint_##opname(MVMThreadContext *tc, MVMObject *result_type, MV
         ia = force_bigint(tc, ba, 0); \
         ib = force_bigint(tc, bb, 1); \
         ic = MVM_malloc(sizeof(mp_int)); \
-        mp_init(ic); \
-        mp_##opname(ia, ib, ic); \
+        if ((err = mp_init(ic)) != MP_OKAY) { \
+            MVM_exception_throw_adhoc(tc, "Error initializing a big integer: %s", mp_error_to_string(err)); \
+        } \
+        if ((err = mp_##opname(ia, ib, ic)) != MP_OKAY) { \
+            MVM_exception_throw_adhoc(tc, "Error performing ##opname with big integers: %s", mp_error_to_string(err)); \
```
sjaeckel: same

src/math/bigintops.c
```diff
@@ -1161,16 +1184,8 @@ MVMObject * MVM_bigint_rand(MVMThreadContext *tc, MVMObject *type, MVMObject *b)
...
-#ifdef MP_NEW_LTM_VERSION
-            mp_digit p = MP_GEN_RANDOM_MAX;
-            mp_rand_digit(&p);
-            result_int = p;
-#else
-            result_int = MP_GEN_RANDOM();
-#endif
-            result_int = result_int % smallint_max;
-            if(have_to_negate)
+            mp_digit result_int = rand() % smallint_max;
```
sjaeckel: I'm pretty sure you shouldn't use `rand()` here

MasterDuke17: I definitely wasn't 100% sure about this change (and it was a long enough time ago that I don't remember all that I tried), but `MP_GEN_RANDOM()` seems to be gone from libtommath v1.2.0 and `mp_rand_digit` is deprecated in favor of `mp_rand`, but I just want a random digit, not a whole `mp_int`. What do you suggest instead of `rand()`?

sjaeckel: can't you use `MVM_getrandom()`?

MasterDuke17: Doh! Don't know why I didn't do that before. Just pushed a change to do so.

src/math/bigintops.c
```diff
@@ -1184,7 +1204,9 @@ MVMObject * MVM_bigint_rand(MVMThreadContext *tc, MVMObject *type, MVMObject *b)
 
     if (use_small_arithmetic) {
         if (MP_GEN_RANDOM_MAX >= abs(smallint_max)) {
-            mp_digit result_int = rand() % smallint_max;
+            mp_digit result_int;
+            MVM_getrandom(tc, &result_int, MP_GEN_RANDOM_MAX);
```
sjaeckel: I guess you want to check the return value ... (just in case something's really broken)

PR #1224, opened 2019-12-25
Fix compiler warnings

MasterDuke17:

### DIFF
--- a/src/profiler/telemeh.c
+++ b/src/profiler/telemeh.c
@@ -282,7 +282,7 @@ void serializeTelemetryBuffer(FILE *outfile)
     lastSerializedIndex = serializationEnd;
 }
 
-void *backgroundSerialization(void *outfile)
+void backgroundSerialization(void *outfile)
 {
     while(continueBackgroundSerialization) {
         MVM_sleep(500);
@@ -290,8 +290,6 @@ void *backgroundSerialization(void *outfile)
     }
 
     fclose((FILE *)outfile);
-
-    return NULL;
 }
 
 MVM_PUBLIC void MVM_telemetry_init(FILE *outfile)
@@ -314,7 +312,7 @@ MVM_PUBLIC void MVM_telemetry_init(FILE *outfile)
 
     beginningEpoch = epochRecord->u.epoch.time;
 
-    threadCreateError = uv_thread_create(&backgroundSerializationThread, (uv_thread_cb)backgroundSerialization, (void *)outfile);
+    threadCreateError = uv_thread_create((uv_thread_t *)&backgroundSerializationThread, backgroundSerialization, (void *)outfile);
     if (threadCreateError != 0)  {
         telemetry_active = 0;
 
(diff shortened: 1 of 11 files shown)


src/strings/utf8.c
```diff
@@ -353,7 +353,7 @@ MVMuint32 MVM_string_utf8_decodestream(MVMThreadContext *tc, MVMDecodeStream *ds
     MVMDecodeStreamBytes *cur_bytes = NULL;
     MVMDecodeStreamBytes *last_accept_bytes     = ds->bytes_head;
     MVMDecodeStreamBytes *lag_last_accept_bytes = NULL;
-    MVMint32 last_accept_pos, lag_last_accept_pos, ready, at_start;
+    MVMint32 last_accept_pos, lag_last_accept_pos = 0, ready, at_start;
```
niner: In what case can `lag_last_accept_pos` actually be used uninitialized like the compiler warns us about and why is initializing with 0 the right thing to do as compared to some change to the algorithm?

src/profiler/telemeh.c
```diff
@@ -314,7 +312,7 @@ MVM_PUBLIC void MVM_telemetry_init(FILE *outfile)
 
     beginningEpoch = epochRecord->u.epoch.time;
 
-    threadCreateError = uv_thread_create(&backgroundSerializationThread, (uv_thread_cb)backgroundSerialization, (void *)outfile);
+    threadCreateError = uv_thread_create((uv_thread_t *)&backgroundSerializationThread, backgroundSerialization, (void *)outfile);
```
niner: Why the type cast to `uv_thread_t *`? `backgroundSerializationThread` seems to already be a `uv_thread_t`.

src/strings/utf8.c
MasterDuke17: There are combinations of conditionals that could lead to it being used uninitialized. I haven't checked whether those combinations are actually possible. @jnthn `lag_last_accept_pos` was introduced in https://github.com/MoarVM/MoarVM/commit/a6abd3c6654413d2230470dbaa82b7b3a2b05762, do you have a suggestion?

src/profiler/telemeh.c
MasterDuke17: With `gcc (Ubuntu 9.2.1-9ubuntu2) 9.2.1 20191008` I otherwise get the warning:
```
src/profiler/telemeh.c: In function ‘MVM_telemetry_init’:
src/profiler/telemeh.c:317:74: warning: cast between incompatible function types from ‘void * (*)(void *)’ to ‘void (*)(void *)’ [-Wcast-function-type]
  317 |     threadCreateError = uv_thread_create(&backgroundSerializationThread, (uv_thread_cb)backgroundSerialization, (void *)outfile);
      |                                                                          ^
```

src/strings/utf8.c
jnthn: I think it hinges on the use of `-1` in `lag_codepoint` being used as an indication of whether there'll be something valid in `lag_last_accept_pos`. It may be that the compiler can't track that.

MasterDuke17: Ah, so you think initializing to 0 to silence the warning is acceptable?

PR #1228, opened 2020-01-13
Implement `--git-reference` Configure.pl option

patrickbkr:
This option makes the configure process create / update git repositories for all of its submodules in the given folders and use those as reference repos.
This is similar, but not exactly the same as what the rakudo `--git-reference` option has been. The rakudo option did not create or update the referenced repository, but just used it. The implementation of this PR creates and updates the reference repository if it's not found. So as a user one doesn't need to know how the submodules are called or what their urls are.
I plan to create corresponding PRs for nqp and rakudo shortly.

Draft PR to see what CI will do.

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -4,6 +4,7 @@ use strict;
 use warnings;
 
 use Config;
+use Cwd;
 use Getopt::Long;
 use Pod::Usage;
 use File::Spec;
@@ -48,7 +49,7 @@ GetOptions(\%args, qw(
     build=s host=s big-endian jit! enable-jit
     prefix=s bindir=s libdir=s mastdir=s
     relocatable make-install asan ubsan
-    valgrind telemeh show-autovect
+    valgrind telemeh show-autovect git-cache-dir=s
     show-autovect-failed:s),
 
     'no-optimize|nooptimize' => sub { $args{optimize} = 0 },
@@ -77,18 +78,9 @@ for my $target (@target_dirs) {
     $config{$target} = $args{$target} if $args{$target};
 }
 
-if (-d '.git') {
-    print dots("Updating submodules");
-    my $msg = qx{git submodule sync --quiet && git submodule --quiet update --init 2>&1};
-    if ($? >> 8 == 0) { print "OK\n" }
-    else {
-        if ($msg =~ /[']([^']+)[']\s+already exists and is not an empty/) {
-            $folder_to_delete = "\n\nERROR: Cannot update submodule because directory exists and is not empty.\n" .
-            ">>> Please delete the following folder and try again:\n$1\n\n";
-        }
-        softfail("git error: $msg")
-    }
-}
+# Download / Update submodules
+my $code = system($^X, 'tools/update-submodules.pl', Cwd::cwd(), @args);
+exit 1 if $code >> 8 != 0;
 
 # fiddle with flags
 $args{optimize}     = 3 if not defined $args{optimize} or $args{optimize} eq "";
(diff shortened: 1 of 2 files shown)


patrickbkr:
CI is happy. I tested this on Linux and Windows.

niner:
What's the use case for this?

patrickbkr:
The `--git-reference` option is used by rakudobrew to speed up redownloading the repos. I'm not sure if there have ever been other users.
The way it worked before (it broke during the big build system refactor) only saved on the loading time of the rakudo, nqp and moar repositories, not all the submodules. Rakudobrew cloned Moar, NQP and Rakudo into a reference directory and then passed that folder to the rakudo build, the option was passed down to NQP (which then cloned Moar with the `--reference` git option). But sumodules were never sped up. Downloading moars submodules is currently >50MB (maybe even >100MB, I don't remember). So this (together with the other PRs for nqp and rakudo) makes all git repository checkouts (including submodules) use a reference directory and thus save a lot of time / energy.
Also usage of the `--git-reference` option is a lot simpler now. It's not necessary to clone the reference repository manually beforehand anymore. One can just point `--git-reference` to an empty directory and everything will work out automatically. I'm currently even thinking about making this controllable by a environment variable. Then one wouldn't even have to think about passing this option. All checkouts (of nqp, moarvm and their submodules) would automatically speed up dramatically.

niner:
Ok, that makes a lot of sense and could indeed be useful in other cases as well. I'm not sure I'm comfortable with the writing aspect though. A reference is something you use for looking up information. It's not something I'd expect to get changed by the use. A use for the reference can e.g. be to avoid network access. OTOH help with setting up the reference can be useful. Can you split these two operations?

patrickbkr:
I think it's important that each project is responsible for its own reference repos (so the Configure.pl of rakudo should not try to set up the reference repos for the submodules of moarvm or nqp). The way the PR currently works is that setup happens on first use. The only information flowing from one project to the other is the `--git-reference` parameter which is passed to the subprojects Configure.pl *during the configuration step* each project sets up its own references repos.


Using a reference can not avoid network access entirely. It will always access the remote repository to query what needs to be updated, but then avoids downloading parts that can be found in the reference repository. So it usually makes sense to not only initially set up a reference repository and then only reference it, but update (fetch) the reference repository before each use, so the objects that are downloaded also end up in the reference. So when splitting the functionality up, we'd need three operations:
1. Setup
2. Update
3. Use

The update and use operations are easy to do. But having a separate setup functionality is tricky.

If such a functionality sets up *all* reference repos that are needed (so that calling `Configure.pl --setup-references=/path/to/ref/dir/` in a rakudo repository will also automatically set up the reference repos for nqp and moarvm) means the setup functionality would need to clone the nqp and moarvm repos to call their setup functionality. Questions
- Where should the moarvm and nqp repos be cloned to?
- Should they be deleted after having setup the reference repos?
- Should we use those clones as reference directly and just accept that we have non-bare reference repos for nqp and moar?
- What if one doesn't want to build for moarvm, but JVM, how do we prevent setting up moarvm in that case?

Alternatively one could just *not* try to have a setup functionality that sets up references for *all* involved repos, but only the ones for the own project. So one would need to manually clone rakudo, nqp and moarvm and call `Configure.pl --setup-references=/path/to/ref/dir/` in each of them separately and probably delete the nqp and moarvm repositories again afterwards.

None of the above ideas for a separate setup step seem attractive to me. I'm open for suggestions!

patrickbkr:
@niner: What about renaming the option to `--git-cache-dir`? That would be a bit more telling of how the option works. Also that would differentiate the option from the previously existing `--git-reference` option.

PR #1231, opened 2020-01-30
[IP6NS Grant] Implement support for bind before connect with sync sockets

Kaiepi:
Needed for upcoming support for bind before connect and UDP sockets with `IO::Socket::INET`.

Passes `make test` and `make spectest`.


src/jit/core_templates.expr
```diff
@@ -20,10 +20,6 @@
 
 (template: set (copy $1))
 
-(template: trunc_i8 (ucast $1 int_sz 1))
```
jnthn: These changes don't seem to belong in this PR?

Kaiepi: These changes are here because I ran the script that sorts the templates so the new listen op would be in the right place.

PR #1233, opened 2020-02-02
Use new functions introduced in ltm v1.2.0...

MasterDuke17:
to instead of our own implementations written before they were available
upstream.


lizmat:
So what's the reason for adding all those:

         mp_clear(ic); \
         MVM_free(ic); \

??  Will that affect performance?

MasterDuke17:
Those are to clean/free up allocated resources before throwing an exception. In the switch to libtommath v1.2.0 I added a lot of error checking and throwing of exceptions, but they would leak.

MasterDuke17:
And yes, there will be a slight performance loss, but most of those are in failure paths anyway, and the cost of not doing them is leaking memory.

lizmat:
Ok, as long as they're in failure paths  :-)  And not leaking memory is also good.

lizmat:
Merged because we're relatively early in the release cycle.

PR #1238, opened 2020-02-12
Make nqp::rand_I respect nqp::srand again

MasterDuke17:
The upgrade to libtommath v1.2.0 brought along a new version of mp_rand
that uses sources of randomness that aren't seeded with srand. So copy
their mp_rand, but modify to use tinymt64 as the source of randomness.

NQP builds ok and passes `make m-test` and Rakudo builds ok and passes `make m-test m-spectest`. Fixes https://github.com/rakudo/rakudo/issues/3469, should also fix https://github.com/rakudo/rakudo/issues/3470.


lizmat:
Thanks!

PR #1239, opened 2020-02-15
Revert "Update libuv to version 1.34.2"

Kaiepi:
This version bump breaks `Kernel.cpu-cores` on OpenBSD. See
https://github.com/libuv/libuv/pull/2685

This reverts commit f47f531100183c05e99123be2b2c5390525e794e.


lizmat:
Is that the only thing it breaks?  Of so, maybe we can special case nqp::cpucores on that OS?  I thought this was a pretty important libuv  update?

Kaiepi:
`nqp::cpucores` seems to be the only thing this bump breaks, but `ThreadPoolScheduler` depends on it as well, which is a bigger issue because that always returns `0` with this. Making a special case for OpenBSD wouldn't be too big a big deal to write, but I'm wary about making a change like that that people might forget about before the next libuv bump.

I don't know if this version in particular is special for MoarVM compared to the ones before it, but I bisected the problem to https://github.com/libuv/libuv/commit/2ab3dc1567d67924b634be6f478d25f4528ce574, which is a change from v1.34.1. v1.34.0 works OK for me and includes a good chunk of the changes this bump brought. Would using that over this be a good compromise?

lizmat:
https://github.com/rakudo/rakudo/commit/1a7f16bc42 will make sure that the `ThreadPoolScheduler` will not break on `nqp::cpucores` returning garbage.  This feels like a sensible thing to do anyway, regardless of this issue.  But it should make this PR unnecessary.

Kaiepi:
I tested out your changes and... unfortunately they don't; promise and supply tests in roast are still flapping with them. This doesn't appear to happen on v1.34.0.

If it's important to have the changes from v1.34.1 and v1.34.2, I think it'd be alright to have the special case for OpenBSD in `MVM_platform_cpu_count` if an issue were opened about it as a reminder to remove it the next time someone bumps libuv.

Kaiepi:
This is no longer necessary with https://github.com/MoarVM/MoarVM/pull/1258/commits/0743c5df34f46e9396d64857a5592e409d7f744d.

PR #1243, opened 2020-02-22
Doing an mp_clear after a failed mp_init is UB

MasterDuke17:
I checked with jaeckel on #libtom and we shouldn't `mp_clear()` a variable if its `mp_init*()` failed.


lizmat:
Is this important enough to cherry pick into a 2020.02 release?

MasterDuke17:
Not sure. They're pretty much all in error branches, and ones unlikely to be hit. mp_init is pretty simple, about the only thing that it contains that could go wrong is a `calloc`. And if a calloc fails, I doubt the rest of MoavVM is going to be very happy.

jnthn:
I'd say not worth the cherry-pick in my opinion.

PR #1252, opened 2020-03-01
Add nextdispatcherfor/takenextdispatcher ops

vrurg:
Provide support for Raku chained dispatchers. Their purpose is to pass
information to the downstream dispatcher about what upstream dispatcher
must take over the control next when downstream exhausts.

For example, when one wraps a candidate in a multi, the `WrapDispatcher`
must know about the instance of `MultiDispatcher` to switch back to it
when all wrappers are done and the candidate calls one of
`{next|call}{same|with}`.

Technically, both ops are clones for `setdispatcherfor`/`takedispatcher` except that they operate with `next_dispatcher` and `next_dispatcher_for` members of `ThreadContext`; and their purpose differs too.


vrurg:
For whom will review this: I most worry are JIT-related pieces. Though they also cloned from the originals and only stricture member names are changed.

In either case, tested localy with a rather big project of mine. Runs smoothly.

MasterDuke17:
> For whom will review this: my biggest worry are JIT-related pieces. Though they also cloned from the originals and only stricture member names are changed.

FYI, I believe you also need to add an entry in src/jit/graph.c for it to be fully jitted.

vrurg:
@MasterDuke17 I found `takedispatcher` there but not `setdispatcherfor`. Anyway, I think somebody with better understanding of JIT than mine should have a look at it.

PR #1254, opened 2020-03-03
Adds a GitHub action that uses an Alpine container to test-compile

JJ:
This is destined to avoid #1246 in the future by test-compiling in an Alpine container every single push to master.


jnthn:
@JJ Does anything have to be done to enable this other than merging it? Is it normal that I don't see any result of it in Checks in the PR (like would happen if adding or changing a `.travis.yml` in a branch)?

JJ:
1. No
2. Well, you'll see the results in the "Actions" menu once you merge it. It can't be checked really.

niner (changes requested):
Can you please squash the commits into a single one with an appropriate commit message? Then it'll be ready for merge.

JJ:
No problem in doing that, but there's also a pull-down menu in the merge button where you can do that yourself, it says "Squash and merge". It's fairly recent, but does the job.

PR #1269, opened 2020-04-11
Add missing parentheses

dogbert17:
The original test used a boolean result in a bitwise operation
which is clearly incorrect.


lizmat:
Good catch!

PR #1272, opened 2020-04-17
Use smaller buffer for native int to str coercion

MasterDuke17:
There can only be up to 20 chars, and the actual worker functions only
use 20 char arrays also.

### DIFF
--- a/src/core/coerce.c
+++ b/src/core/coerce.c
@@ -184,7 +184,6 @@ static char * u64toa_naive_worker(uint64_t value, char* buffer) {
         *buffer++ = *--p;
     } while (p != temp);
 
-    *buffer = '\0';
     return buffer;
 }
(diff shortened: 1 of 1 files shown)


niner:
I think there's an off-by-one here as the buffer gets ended with a null byte: `*buffer = '\0';` which I think is actually unnecessary.

MasterDuke17:
> I think there's an off-by-one here as the buffer gets ended with a null byte: `*buffer = '\0';` which I think is actually unnecessary.

That obviously pre-exists my change, but I'm seeing what happens if it's removed right now.

MasterDuke17:
NQP and Rakudo and all tests are ok with removing it. Should I add that to this PR or create a new one?

niner:
Just add it please

MasterDuke17:
> Just add it please

Done.

PR #1273, opened 2020-04-17
Fix SEGV in MVM_spesh_plugin_guard_list_mark

dogbert17:
MVM_spesh_plugin_guard_list_mark contains several tests checking that
the worklist parameter is not NULL. Unfortunately one call, to
MVM_spesh_plugin_guard_list_mark, had slipped through the cracks.
This could lead to a SEGV when running the heap profiler since it
can call MVM_gc_root_add_tc_roots_to_worklist with a NULL worklist
which in turn was used when calling MVM_spesh_plugin_guard_list_mark.
Fixed by adding a check which ensures that worklist is not NULL
before calling MVM_spesh_plugin_guard_list_mark.

### DIFF
--- a/src/gc/roots.c
+++ b/src/gc/roots.c
@@ -215,7 +215,8 @@ void MVM_gc_root_add_tc_roots_to_worklist(MVMThreadContext *tc, MVMGCWorklist *w
         else
             MVM_spesh_graph_describe(tc, tc->spesh_active_graph, snapshot);
     }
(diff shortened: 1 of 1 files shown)


niner:
The first sentence of the commit message is wrong. You meant MVM_gc_root_add_tc_roots_to_worklist not MVM_spesh_plugin_guard_list_mark.

PR #1277, opened 2020-04-27
Make the first page for some FSA bins bigger

MasterDuke17:
This means fewer calls to the slow FSA alloc path (which calls
`add_page()` and then mallocs). These specific numbers were based on just
calling `raku -e ''` and seeing how many `add_page()` calls there were for
each bin. Any modifications to the number or type of FSA allocations
could mean these defaults might need to be changed. Additionally, we
might decide to increase them assuming that most programs will actually
do more allocations than done by the empty program.

I did some testing with callgrind and time. I ran `MVM_SPESH_BLOCKING=1 valgrind --tool=callgrind raku -e ''` 100 times and stored the instruction count for both master and this branch. Master averaged 702,547,960 instructions and this branch averaged 702,285,088. I also ran each version 100 times with /usr/bin/time (but without using `MVM_SPESH_BLOCKING=1`), master averaged 0.945s elapsed and this branch averaged 0.947s.

### DIFF
--- a/src/core/fixedsizealloc.c
+++ b/src/core/fixedsizealloc.c
@@ -87,6 +87,34 @@ static void setup_bin(MVMFixedSizeAlloc *al, MVMuint32 bin) {
     /* Work out page size we want. */
     MVMuint32 page_size = MVM_FSA_PAGE_ITEMS * ((bin + 1) << MVM_FSA_BIN_BITS) + MVM_FSA_REDZONE_BYTES * 2 * MVM_FSA_PAGE_ITEMS;
 
+    /* The below numbers were determined by logging add_page() calls for `raku -e ''`.
+     * This makes the first page added the minumum size needed for any raku program,
+     * to reduce the number of mallocs and page adds when starting up. */
+    switch(bin) {
+        case  5:
+            page_size *= 315;
+            break;
+        case  6:
+        case 15:
+            page_size *= 38;
+            break;
+        case  1:
+        case  2:
+        case  3:
+        case  4:
+        case  7:
+            page_size *= 8;
+            break;
+        case  0:
+        case  8:
+        case 12:
+        case 17:
+            page_size *= 2;
+            break;
+        default:
+            break;
+    }
+
     /* We'll just allocate a single page to start off with. */
     al->size_classes[bin].num_pages = 1;
     al->size_classes[bin].pages     = MVM_malloc(sizeof(void *) * al->size_classes[bin].num_pages);


niner:
To be honest, I'm not convinced by this PR. It adds a couple of magic numbers tuned for the current state of one of MoarVM's users (rakudo). While MoarVM's goal is specifically to be the best runtime for Raku, it already has another user with NQP. While that alone wouldn't be enough of an argument for preventing a merge, there doesn't seem to be much argument for. There's no measurable performance benefit other than 0.04 % fewer instructions. That just feels like not worth the added maintenance cost.

MasterDuke17:
I think I'd really argue for a larger change. These numbers are just tuned to the empty program, but why should we optimize for that? If we know that any program at all is going to use more memory than we initially allocate, why not allocate more up front? Then we get bigger chunks of contiguous memory. Or should we just make all pages bigger? I.e., increase the multiplier here https://github.com/MoarVM/MoarVM/pull/1277/files#diff-c20c5286479c59cae5058b9716286f1bL88? Should I see if there are scenarios with a larger measurable impact?

lizmat:
Perhaps a benchmark reading a file line by line and saying it?

MasterDuke17:
I just benchmarked with a more dramatic change. I changed `MVM_FSA_PAGE_ITEMS` from 128 to 1024 (but that's it, I didn't include the change to first page sizes made in this PR). The code I tested was `my $c = Channel.new; for ^1_000_000 { $c.send: $_; }`, run with 100 times with callgrind and 100 times with /usr/bin/time. At HEAD of master, it took an average of 3,556,038,563 instructions and 0.4776s. With only the change to `MVM_FSA_PAGE_ITEMS` applied, it took an average of 3,532,358,600 instructions and 0.4654s.

MasterDuke17:
FWIW, adding a print in `add_page` gives these results for the above example:
master
```
      1 adding a page of size 15360 for bin 14
      1 adding a page of size 18432 for bin 17
      1 adding a page of size 21504 for bin 20
      1 adding a page of size 46080 for bin 44
      2 adding a page of size 1024 for bin 0
      3 adding a page of size 24576 for bin 23
      3 adding a page of size 49152 for bin 47
      4 adding a page of size 12288 for bin 11
      4 adding a page of size 8192 for bin 7
      5 adding a page of size 5120 for bin 4
      6 adding a page of size 4096 for bin 3
     38 adding a page of size 16384 for bin 15
     40 adding a page of size 7168 for bin 6
     41 adding a page of size 9216 for bin 8
    259 adding a page of size 3072 for bin 2
    390 adding a page of size 6144 for bin 5
   7815 adding a page of size 2048 for bin 1
```
`MVM_FSA_PAGE_ITEMS` == 1024
```
      4 adding a page of size 131072 for bin 15
      5 adding a page of size 57344 for bin 6
      5 adding a page of size 73728 for bin 8
     32 adding a page of size 24576 for bin 2
     48 adding a page of size 49152 for bin 5
    976 adding a page of size 16384 for bin 1
```

PR #1278, opened 2020-04-27
Fix some compiler warnings

MasterDuke17:
By adding some casts and using correct sized format string directives.

### DIFF
--- a/src/6model/reprs/P6opaque.c
+++ b/src/6model/reprs/P6opaque.c
@@ -647,9 +647,10 @@ static MVMuint16 * allocate_unbox_slots() {
 }
 static void compose(MVMThreadContext *tc, MVMSTable *st, MVMObject *info_hash) {
     MVMint64   mro_pos, mro_count, num_parents, total_attrs, num_attrs,
-               cur_slot, cur_type, cur_alloc_addr, cur_obj_attr,
+               cur_slot, cur_type, cur_obj_attr,
                cur_init_slot, cur_mark_slot, cur_cleanup_slot,
                unboxed_type, i;
+    MVMuint64  cur_alloc_addr;
     MVMObject *info;
     MVMP6opaqueREPRData *repr_data;
 
--- a/src/profiler/configuration.c
+++ b/src/profiler/configuration.c
@@ -290,7 +290,7 @@ static void validate_op(MVMThreadContext *tc, validatorstate *state) {
                     break;
 
                 default:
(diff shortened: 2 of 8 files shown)


src/6model/reprs/P6opaque.c
```diff
@@ -840,7 +840,7 @@ static void compose(MVMThreadContext *tc, MVMSTable *st, MVMObject *info_hash) {
             /* C structure needs careful alignment. If cur_alloc_addr is not
              * aligned to align bytes (cur_alloc_addr % align), make sure it is
              * before we add the next element. */
-            align_to(&cur_alloc_addr, align);
+            align_to((MVMuint64 *)&cur_alloc_addr, align);
```
niner, line 843: Why not turn cur_alloc_addr into a proper MVMuint64 instead of casting here? I don't see a place where it would ever have a negative value.

src/profiler/configuration.c
```diff
@@ -290,7 +290,7 @@ static void validate_op(MVMThreadContext *tc, validatorstate *state) {
                     break;
 
                 default:
-                    MVM_exception_throw_adhoc(tc, "STRUCT_SELECT string length %llu (index %d) NYI or something", string_length, string_idx);
+                    MVM_exception_throw_adhoc(tc, "STRUCT_SELECT string length %lu (index %d) NYI or something", string_length, string_idx);
```
niner, line 293: I think both `%llu` and `%lu` are wrong as their meaning is platform dependent. MVMuint64 is an alias for uint64_t and according to `man inttypes.h` the PRIu64 macro is the correct choice: `"STRUCT_SELECT string length " PRIu64 " (index %d) NYI or something"`

src/spesh/arg_guard.c
```diff
@@ -309,7 +309,7 @@ static void add_nodes_for_callsite(MVMThreadContext *tc, MVMSpeshArgGuard *tree,
         ? add_nodes_for_typed_argument(tc, tree, candidates, cc.cs, cc.typed_idxs,
                 MVM_VECTOR_ELEMS(cc.typed_idxs), next_type_index(cc.cs, -1), 0,
                 end_node)
-        : end_node;
+        : (MVMuint32)end_node;
```
niner, line 312: Same here, I'd think making end_node an MVMuint32 in the first place should work out fine. `add_result_node` already returns an MVMuint32

MasterDuke17: Yeah, those are all better. Force-pushed an update. That's what I get for making late night PRs.

PR #1281, opened 2020-05-03
Spesh and JIT get(cur)?hllsym

MasterDuke17:
Because one of the arguments to the `MVM_hll_sym_get()` call is not an
argument of the op, it can't be JITted normally. So instead re-write it
in spesh to a new op that can, and then JIT the new op. An unsafe template
for `getcurhllsym` already existed, but wasn't being called because there
was no lego JIT implementation. So instead I just adapted that template
to the new op.

Speeds up the benchmark in https://github.com/rakudo/rakudo/issues/3648 by ~0.3s.


niner:
The implementation looks perfectly fine to me and is certainly merge-worthy as-is.

However, I sense an opportunity for further optimization. Looking at the implementation of `MVM_hll_sym_get`, that's a pretty heavy weight operation but it actually just a glorified hash access. Actually it's 2 hashes: first it gets the symbol hash for the given `hll`, then it gets the requested symbol from this symbol hash. If the HLL doesn't have a symbol hash yet it creates one and returns VMNull for the value. Once the symbol hash is created, it never changes again. Worst case it gets moved into gen2 by the GC.

This check is necessary only for the very first call to MVM_hll_sym_get or bindhllsym or bindcurhllsym for a given HLL. When we spesh code, we do so because it has run often enough and in anticipation of the code running often again in the future. So we'll want to optimize away a check that will always succeed anyway. While optimizing away a branch is good, there's much more to be gained here: we may be able to get rid of the lock that is taken on every call.

Under perfect circumstances we can not only put the HLL name into a spesh slot, and not even just the symbol hash, but the HLL symbol's hash entry itself. These entries are not GC-allocated, so their address doesn't change. Binding to an existing key will just overwrite the existing hash entry's `value` pointer. Overwriting that pointer is an atomic operation by itself. The possible road block is that writing to that entry will also do a write barrier, i.e. if the hash is in gen2 (which it will be in most cases) and the value is not, the hash must be put onto the GC's gen2 root list if it's not already there. And this operation is not thread safe. But AFAICT this could still be ok, as I think worst case, the hash will land in multiple thread's root lists and I haven't seen protection against such a situation anywhere else.

Anyway, all of this is moot if `getcurhllsym` isn't actually used all that much. Which may well be, since I don't find it in anything I'd presume hot code. Makes me wonder where that speedup is coming from?

MasterDuke17:
getcurhllsym is called by DYNAMIC, so running say() in a loop will have a call every iteration for $*OUT. At least, IIRC that’s what I was testing with.

lizmat:
Not directly, which nqp::op does it then internally?

MasterDuke17:
> Not directly, which nqp::op does it then internally?

https://github.com/rakudo/rakudo/blob/master/src/core.c/io_operators.pm6#L12-L18

MasterDuke17:
Just force-pushed a new implementation that also does `gethllsym`. For some rough numbers, `my $o; $o := $*OUT for ^1_000_000; say $o` goes from ~1.1s to ~0.95s and 10.16b instructions to 9.13b instructions. I looked at a couple spesh logs and saw both `gethllsym` and `getcurhllsym` being specialized into `sp_gethashentryvalue`.

niner:
The commit message and/or the comment in the code should clearly state the assumptions that the code makes:
* that MVMHashEntrys are not GC allocated and therefore never move
* that the MVMHashEntry pointers stay constant even when the hash grows
* that there cannot be concurrency issues wrt an existing hash keys (i.e. updating a hash entry keeps the MVMHashEntry struct and just overwrites the value pointer atomically)
* that an HLL's symbol hash will never be replaced once it's created
* that the code with the baked in pointer cannot have a shorter life time than the HLL, i.e. it can't happen that the MVMHashEntry gets freed while the code is still being used.

I believe all of these assumptions to be sound, but they really need to be stated explicitly.

PR #1284, opened 2020-05-06
Remove JIT template for prof_enter

MasterDuke17:
The JIT probably won't (should never?) see `prof_enter`, but if it does,
mark the call to `MVM_profile_log_enter` as coming from the JIT.

### DIFF
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -2810,13 +2810,6 @@
       (carg (^spesh_slot_value $1) ptr)
       (carg $2 int)) ptr_sz))
 
-(template: prof_enter
-  (callv (^func MVM_profile_log_enter)
-    (arglist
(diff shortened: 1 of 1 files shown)


jnthn:
Hm, it's a bug if the JIT ever sees this op?

MasterDuke17:
> Hm, it's a bug if the JIT ever sees this op?

Should I change to call `MVM_panic` instead? Or just remove the template entirely?

niner:
I'd say remove the template. While there's some bug that lets the JIT see this op, it's will only cause the JIT to bail, which arguably is preferable to the program crashing. Someone interested in fixing this can easily search a spesh log for a bail on prof_enter to find a case.

MasterDuke17:
Changed the commit to remove the template (and updated the PR title accordingly).

PR #1286, opened 2020-05-14
Lego JIT implementation of sp_getlexstatic_o

MasterDuke17:
Manually inline MVM_disp_inline_cache_get_spesh because it's small.

NQP and Rakudo both build ok, and a spesh log now only shows template bails for `sp_getlexstatic_o`. There are some failing tests, but the same ones fail without this also.

### DIFF
--- a/src/spesh/disp.c
+++ b/src/spesh/disp.c
@@ -906,11 +906,18 @@ static MVMSpeshIns * translate_dispatch_program(MVMThreadContext *tc, MVMSpeshGr
                 /* If it's not C, we can pre-select a spesh candidate. This will
                  * be taken care of by the optimizer; for now we poke -1 into the
                  * value to mean that we didn't pre-select a candidate. We also
-                 * add deopt all and logged annotations. */
+                 * add deopt all and logged annotations, plus the predeopt one
+                 * if we inserted no guards. */
                 if (!c) {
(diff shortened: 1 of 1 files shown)


src/jit/x64/emit.dasc
```diff
@@ -2084,6 +2086,23 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
...
         break;
     }
+    case MVM_OP_sp_getlexstatic_o: {
+        MVMint16 dst       = ins->operands[0].reg.orig;
+        MVMint16 name      = ins->operands[1].reg.orig;
+        MVMint16 spesh_idx = ins->operands[2].lit_i16;
+        MVMuint32 slot     = ins->operands[3].lit_ui32;
+        | mov ARG1, TC;
+        | get_spesh_slot ARG2, spesh_idx;
+        | lea ARG2, STATICFRAME:ARG2->body.inline_cache;
```
niner, line 2096: Doesn't this overwrite the spesh slot fetched into ARG2?

niner: Ah, yes it does and that's ok, because you're just doing what I suggested :D

MasterDuke17: Ha, yep.

PR #1297, opened 2020-05-29
Remove Perl 5.10 directive from Travi config

MasterDuke17:
The download and subsequent using of it with perlbrew fails and Travis
just uses the system Perl 5.(22 with 16.04, or 26 with 18.04).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,6 @@ language: "perl"
 os:
   - linux
   - osx
-perl:
-  - "5.10"
 # Do a full clone to be sure we get the tags and can properly retrieve version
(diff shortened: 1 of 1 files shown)


MasterDuke17:
Huh, for some reason it (successfully at least) downloads and installs Perl 5.14.

patrickbkr:
I plan to replace travis with AzureCI soon. So don't put too much time into this.

MasterDuke17:
> I plan to replace travis with AzureCI soon. So don't put too much time into this.

Great, I wasn't sure if you were going to do it for MoarVM.

PR #1302, opened 2020-06-03
Update libuv to version 1.38.0

dogbert17:
Clean stresstest under Linux (Mint)


jnthn:
May as well get this in ahead of the next Blin run.

PR #1304, opened 2020-06-04
nativecall: Don't obliterate lib_name in exception text

softmoth:
Fix calling sprintf with NULL (`Cannot locate native library '(null)'`) in exception message.


niner:
Wow....that one's so obvious once you pointed it out. Many thanks for fixing my bugs!

PR #1307, opened 2020-06-08
Prefer lexical names list

nwc10:
The three patches are related, but they all need a little different review.

The first eliminates 5 hash iterations, replacing them with array traversal. It would be a no-brainer, but it affects the debug server, and I don't know how to test that.

I believe that wrapping access to lexical_names makes sense, even if the next patch isn't wanted, but it makes more sense with it

The third patch, eliminating the lookup hash for small lists, should help performance slightly, but

1. I don't have a good way to benchmark this (it doesn't seem to noticeably change setting compilation in any direction)
1. I don't have any feel for what a good threshold is. Are there typically many small frames with just 3 named lexicals? How often do we even do named lexical lookup?

### DIFF
--- a/src/core/bytecode.c
+++ b/src/core/bytecode.c
@@ -653,15 +653,24 @@ void MVM_bytecode_finish_frame(MVMThreadContext *tc, MVMCompUnit *cu,
         for (j = 0; j < sf->body.num_lexicals; j++) {
             MVMString *name = get_heap_string(tc, cu, NULL, pos, 6 * j + 2);
             MVMLexicalRegistry *entry = MVM_calloc(1, sizeof(MVMLexicalRegistry));
-            MVM_HASH_BIND_FREE(tc, sf->body.lexical_names, name, entry, {
-                if (sf->body.num_locals) {
-                    MVM_free_null(sf->body.local_types);
-                }
-                MVM_free_null(sf->body.lexical_types);
-                MVM_free_null(sf->body.lexical_names_list);
-                MVM_free(entry);
-            });
 
+            /* If we do not have "many" lexicals, don't bother making a lookup
+             * hash. Instead, MVM_get_lexical_by_name will simply do a linear
+             * search of the list. This should be faster for short lists.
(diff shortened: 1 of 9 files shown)


lizmat:
> I don't have any feel for what a good threshold is. Are there typically many small frames with just 3 named lexicals? How often do we even do named lexical lookup?

Shouldn't that be decided by the threshold of the number of elements at which a linear lookup becomes slower than a hash lookup?  Regardless of the number of small frames?

src/core/bytecode.c
```diff
@@ -641,7 +641,17 @@ void MVM_bytecode_finish_frame(MVMThreadContext *tc, MVMCompUnit *cu,
...
 
             sf->body.lexical_types[j] = read_int16(pos, 6 * j);
-            MVM_HASH_BIND(tc, sf->body.lexical_names, name, entry)
+
+            /* If we do not have "many" lexicals, don't bother making a lookup
+             * hash. Instead, MVM_get_lexical_by_name will simply do a linear
+             * search of the list. This should be faster for short lists.
+             * The choice of 5 entries is guesswork, and ought to be refined by
+             * benchmarking. This location is the only place where we need this
+             * "magic" constant, hence I don't see the ned to #define it
```
MasterDuke17, line 650: "ned"

nwc10: @lizmat Not sure if I'm answering the wrong question here...

There is a list and (currently) a lookup hash maintained for each and every static frame. The threshold for "when not to bother with a hash" is going to be per-frame. It's easiest to have it as a constant, and the same constant for all frames.

You're right that it should be "becomes slower than a hash lookup" - I have no good idea when that is, or even how to measure it well. 5 was a guess. The memory overhead of the hash structure is a lot more than 5 structs, and there's a bunch more CPU work before you even get to compare the wanted string key with the entry in the hash. Plus doing the hash lookup generates (at least) 1 more pointer indirection, 2 if your hash isn't perfectly spread out and you have two entries in the same linked list.

PR #1311, opened 2020-06-12
Add missing concreteness check

dogbert17:
The missing check led to a noticable performance regression.
[Tux]++ for spotting.


MasterDuke17:
@timo pointed out this should probably be `if (is->invocation_handler && IS_CONCRETE(is->invocation_handler))` or `if (!MVM_is_null(tc, is->invocation_handler) && IS_CONCRETE(is->invocation_handler))`

PR #1317, opened 2020-06-21
Pre-declare loop vars.

0racle:
Fixes builds on older versions of gcc.

### DIFF
--- a/src/6model/reprs/MVMStaticFrame.c
+++ b/src/6model/reprs/MVMStaticFrame.c
@@ -68,7 +68,8 @@ static void copy_to(MVMThreadContext *tc, MVMSTable *st, void *src, MVMObject *d
         memcpy(lexical_names_list, src_body->lexical_names_list,
                sizeof(MVMLexicalRegistry *) * src_body->num_lexicals);
 
-        for (MVMuint32 j = 0; j < num_lexicals; j++) {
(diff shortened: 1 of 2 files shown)


0racle:
This is similar to commit fc09255. Just prior to that commit being merged, another commit was merged which added more of these loops var declarations.

Can we have these detected in the CI pipeline to prevent breaking builds on older systems?

ugexe:
fwiw the Rakudo Circle CI has been complaining/failing over this for awhile

niner:
Is there really no other way to convince gcc that this is correct code? We've just got rid of the restrictions imposed on us by MSVC. It would be a shame if we now needed to code like it's 1995 because of gcc versions that really fully support C99 and even parts of C11.

patrickbkr:
I have a PR for a different approach here: [#1318](https://github.com/MoarVM/MoarVM/pull/1318)

0racle:
I have confirmed that Moar also builds successfully for me if I manually specify `CFLAGS='-std=gnu99'`, so I'm happy for #1318 to be merged in lieu of in this one. It also means we could revert fc09255

patrickbkr:
#1318 has been merged. So I'll close this one.
@0racle Thanks for the report and work you put into this!

PR #1321, opened 2020-06-27
Don't add placeholders such as "<control>" to the Unicode names lookup hash.

nwc10:
@samcv , @jnthn  or *, could you check that I'm not doing anything stupid here?

~~I don't want to just push it, in case there was something I missed.~~

*Woah* I think I've done something stupid in the "fix" here (it's not in the right place). But is the reasoning sound? That we don't want these duplicates in the hash.

Rest of the commit message is the reasoning:

```
Including these make no sense - the hash is for a 1->1 mapping of name to
codepoint, but for each of these names we have multiple values.

The way our hashes are currently implemented (with "bind"), these duplicate
keys are not ignored - we add hundreds of duplicate entries to the same`
bucket chain (which will never successfully split to shorter chains).
This will cause performance degradation for other name lookups on the same
bucket chain - algorithmic pain, not merely memory bloat.

We *are* effectively executing a DOS attack against ourselves. Oops. :-)
```

For completeness, I should note that there are some duplicates in `unicode_property_value_keypairs`, such as `{"18-number",1258291201},{"18-number",1258291201}` but
 

1. there don't seem to be more than 2 entries with the same key duplicated
1. the values are indentical
1. I wasn't sure if it's safe to de-duplicate them in `ucd2.pl`, or if the array index they are at *also* matters.

so I don't think that it's a performance hit, or much memory bloat.

And I didn't want to try to experiment, as the way `UCD-download.p6` is written, it's awkward to do anything other than the most recent release of Unicode. `LATEST` is now on 13. We're on 12.1

### DIFF
--- a/src/strings/unicode_ops.c
+++ b/src/strings/unicode_ops.c
@@ -748,13 +748,6 @@ MVMString * MVM_unicode_get_name(MVMThreadContext *tc, MVMint64 codepoint) {
         MVMint32 codepoint_row = MVM_codepoint_to_row_index(tc, codepoint);
         if (codepoint_row != -1) {
             name = codepoint_names[codepoint_row];
-            if (!name) {
-                while (codepoint_row && !codepoint_names[codepoint_row])
-                    codepoint_row--;
-                name = codepoint_names[codepoint_row];
-                if (name && name[0] != '<')
-                    name = NULL;
-            }
         }
         if (!name) {
             /* U+FDD0..U+FDEF and the last two codepoints of each block
--- a/tools/ucd2c.pl
+++ b/tools/ucd2c.pl
@@ -658,24 +658,21 @@ sub add_extent {
 }
 # Used in emit_codepoints_and_planes to push the codepoints name onto bitfield_index_lines
 sub ecap_push_name_line {
-    my ($name_lines, $name, $point, $bitfield_index_lines, $bytes, $index, $annotate_anyway) = @_;
+    my ($name_lines, $name, $point, $bitfield_index_lines, $index) = @_;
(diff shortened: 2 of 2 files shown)


nwc10:
OK, so having completely goofed and put the new code in the wrong place, here it is in the "right" place.

*Except*

I think that this still isn't the correct place. If I follow the logic correctly (the size of the file does not make my editor happy), the *only* use of the array  `codepoint_names` is in `generate_codepoints_by_name`. In which case, really we shouldn't even put the names into the C file if they start with `<`

And (oh my, is it complicated) - the lack of names might mean that `ucd2c.pl` could actually turn more of these sections into spans, if these sections have no properties. (I suspect that they do, but that code is really hard to follow.)

lizmat:
Question: does ucd2c.pl need to be run during the MoarVM build process?  If not, I would suggest making it a Raku script  :-)

nwc10:
OK, I conclude that it's quite sane to remove the various placeholders such as `<control>` from the lookup, given the current behaviour:

```
$ for i in `seq 0 9`; do ./nqp -e 'nqp::say(nqp::codepointfromname("<control>"))'; done
31
158
31
31
159
5
159
31
159
31
```

No way can anyone be relying on that behaviour.

(I think that the unpredictable value you get is a side effect of bucket insertion order randomisation)

I would also suggest that `ucd2c.pl` would benefit from being rewritten to something sane, but as it is about as sane as one of Damian's regexs, it's not a SMOP. (The whole thing, it and the C code it generates/interfaces with, is 6 to 7 years old, and rather too clever for my liking)

PR #1326, opened 2020-07-14
Fix build with gcc 4.8

ffontaine:
Build with gcc 4.8 is broken since version 2020.06 and commit c70587551ab093368e041a65f4a4a8b49d78ce73:

```
src/debug/debugserver.c: In function 'request_context_lexicals':
src/debug/debugserver.c:1456:9: error: 'for' loop initial declarations are only allowed in C99 mode
         for (MVMuint32 j = 0; j < num_lexicals; j++) {
         ^
```
Fixes:
 - http://autobuild.buildroot.org/results/179/1799769c531f9f26ff191b1adcd5e5b17dc3b244/build-end.log

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/6model/reprs/MVMStaticFrame.c
+++ b/src/6model/reprs/MVMStaticFrame.c
@@ -68,7 +68,8 @@ static void copy_to(MVMThreadContext *tc, MVMSTable *st, void *src, MVMObject *d
         memcpy(lexical_names_list, src_body->lexical_names_list,
                sizeof(MVMLexicalRegistry *) * src_body->num_lexicals);
 
-        for (MVMuint32 j = 0; j < num_lexicals; j++) {
+        MVMuint32 j;
+        for (j = 0; j < num_lexicals; j++) {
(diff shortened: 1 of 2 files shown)


patrickbkr:
Thanks for your work on this. But I think this has already been fixed with [23dfde978ce327](https://github.com/MoarVM/MoarVM/commit/23dfde978ce32760001c7b4fb0c1bfa2f1adcb7c).
Have you tried this only with the 2020.06 release (which I think did not include the above commit) or also with latest master?

niner:
This should actually have been fixed with https://github.com/MoarVM/MoarVM/commit/23dfde978ce32760001c7b4fb0c1bfa2f1adcb7c
Can you confirm that the compiler is run with `-std=gnu99`? Can you please investigation why that isn't enough?

ffontaine:
Yes, commit 23dfde978ce32760001c7b4fb0c1bfa2f1adcb7c fixes the issue, thanks for your help, I'll close this PR.

PR #1327, opened 2020-07-16
Improve serialization of C types

Kaiepi:
This adds support for serializing C arrays and fixes a potential integer overflow in C string deserialization. The purpose of this is to make it possible for abstract UNIX socket addresses to be serialized in my solution for https://github.com/Raku/problem-solving/issues/111.

### DIFF
--- a/src/6model/serialization.c
+++ b/src/6model/serialization.c
@@ -1739,8 +1739,8 @@ MVMString * MVM_serialization_read_str(MVMThreadContext *tc, MVMSerializationRea
 
 /* Reading function for arrays. */
 void * MVM_serialization_read_array(MVMThreadContext *tc, MVMSerializationReader *reader, size_t *size) {
-    size_t  array_size = MVM_serialization_read_int(tc, reader);
-    void   *array      = NULL;
+    MVMint64  array_size = MVM_serialization_read_int(tc, reader);
+    void     *array      = NULL;
     if (array_size) {
         const MVMuint8 *read_at = (MVMuint8 *)*(reader->cur_read_buffer) + *(reader->cur_read_offset);
         assert_can_read(tc, reader, array_size);
@@ -1755,8 +1755,8 @@ void * MVM_serialization_read_array(MVMThreadContext *tc, MVMSerializationReader
 
 /* Reading function for null-terminated char array strings. */
 char * MVM_serialization_read_cstr(MVMThreadContext *tc, MVMSerializationReader *reader, size_t *len) {
-    size_t  string_len = MVM_serialization_read_int(tc, reader);
-    char   *string     = NULL;
+    MVMint64  string_len = MVM_serialization_read_int(tc, reader);
+    char     *string     = NULL;
     if (string_len) {
         if (string_len < SIZE_MAX) {
             const MVMuint8 *read_at = (MVMuint8 *)*(reader->cur_read_buffer) + *(reader->cur_read_offset);
@@ -1768,7 +1768,7 @@ char * MVM_serialization_read_cstr(MVMThreadContext *tc, MVMSerializationReader 
         }
         else
             fail_deserialize(tc, NULL, reader,
-                "Cannot read a C string with length %zu.",
+                "Cannot read a C string with length %"PRIi64".",
                 string_len);
     }
     if (len)
--- a/src/6model/serialization.c
+++ b/src/6model/serialization.c
@@ -326,6 +326,17 @@ static void expand_storage_if_needed(MVMThreadContext *tc, MVMSerializationWrite
     }
 }
 
+/* Writing function for arrays. */
+void MVM_serialization_write_array(MVMThreadContext *tc,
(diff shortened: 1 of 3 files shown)


src/6model/serialization.c
```diff
@@ -337,21 +337,9 @@ void MVM_serialization_write_array(MVMThreadContext *tc,
...
-        MVM_serialization_write_int(tc, writer, len);
-        expand_storage_if_needed(tc, writer, len);
-        memcpy(*(writer->cur_write_buffer) + *(writer->cur_write_offset), string, len);
-        *(writer->cur_write_offset) += len;
-    } else {
-        MVM_serialization_write_int(tc, writer, 0);
-    }
+/* Writing function for null-terminated char array strings. */
+void MVM_serialization_write_cstr(MVMThreadContext *tc, MVMSerializationWriter *writer, const char *string) {
+    MVM_serialization_write_array(tc, writer, string, string && strlen(string));
```
niner, line 342: Looks like a Perlism slipped through. I'm pretty sure this will pass 1 instead of the string's length as `size` and if it doesn't the compiler actually violates the C standard.

src/6model/serialization.c
```diff
@@ -1755,23 +1743,27 @@ void * MVM_serialization_read_array(MVMThreadContext *tc, MVMSerializationReader
...
-        if (strbuf == 0)
-            fail_deserialize(tc, NULL, reader, "Cannot read a c string: malloc failed.");
-        memcpy(strbuf, read_at, len);
-        strbuf[len] = 0;
-        *(reader->cur_read_offset) += len;
-    } else if (len < 0) {
-        fail_deserialize(tc, NULL, reader, "Cannot read a c string with negative length %"PRIi64".", len);
+/* Reading function for null-terminated char array strings. */
+char * MVM_serialization_read_cstr(MVMThreadContext *tc, MVMSerializationReader *reader, size_t *len) {
+    size_t  string_len = MVM_serialization_read_int(tc, reader);
```
niner, line 1748: `MVM_serialization_read_int` reads a 64 bit integer (stored in a variable length field), thus putting the result into an `MVMint64` was correct. `size_t` on the other hand is platform dependent and will only be 32 bits on a 32 bit platform. Same is true for SIZE_MAX. Thus the `string_len < SIZE_MAX` check will only reject a single, precise length: `2**32 - 1` or `2**64 - 1`. I'd leave `string_len` as MVMint64 and cast it to size_t after we've checked that it actually fits into one.

niner (changes requested):

Kaiepi:
Updated.

niner (approved):

Kaiepi:
Can this be merged?

src/6model/serialization.c
```diff
@@ -1738,23 +1736,53 @@ MVMString * MVM_serialization_read_str(MVMThreadContext *tc, MVMSerializationRea
...
-    } else if (len < 0) {
-        fail_deserialize(tc, NULL, reader, "Cannot read a c string with negative length %"PRIi64".", len);
+/* Reading function for pointers. */
+void * MVM_serialization_read_ptr(MVMThreadContext *tc, MVMSerializationReader *reader, size_t *size) {
+    MVMint64  ptr_size;
+    void     *ptr;
+
+    if (!(ptr_size = MVM_serialization_read_int(tc, reader)))
+        ptr = NULL;
+    else if (ptr_size < 0 || ptr_size > SIZE_MAX)
```
MasterDuke17, line 1746: This causes a compiler warning:
```
src/6model/serialization.c: In function ‘MVM_serialization_read_ptr’:
src/6model/serialization.c:1746:39: warning: comparison of integer expressions of different signedness: ‘MVMint64’ {aka ‘long int’} and ‘long unsigned int’ [-Wsign-compare]
 1746 |     else if (ptr_size < 0 || ptr_size > SIZE_MAX)
      |                                       ^
```

src/6model/serialization.c
```diff
@@ -1738,23 +1736,53 @@ MVMString * MVM_serialization_read_str(MVMThreadContext *tc, MVMSerializationRea
...
+}
+
+/* Reading function for null-terminated char array strings. */
+char * MVM_serialization_read_cstr(MVMThreadContext *tc, MVMSerializationReader *reader, size_t *len) {
+    MVMint64  cstr_len;
+    char     *cstr;
+
+    if (!(cstr_len = MVM_serialization_read_int(tc, reader)))
+        cstr = NULL;
+    else if (cstr_len < 0 || cstr_len >= SIZE_MAX)
```
MasterDuke17, line 1770: And so does this:
```
src/6model/serialization.c: In function ‘MVM_serialization_read_cstr’:
src/6model/serialization.c:1770:39: warning: comparison of integer expressions of different signedness: ‘MVMint64’ {aka ‘long int’} and ‘long unsigned int’ [-Wsign-compare]
 1770 |     else if (cstr_len < 0 || cstr_len >= SIZE_MAX)
      |                                       ^~
```

PR #1329, opened 2020-07-22
Fix case insensitive string matching with an empty string

patrickbkr:
Fixes `'' ~~ m:i/''/` in Rakudo. I did a spectest run with this.

This PR is related to respective tests in roast: <https://github.com/Raku/roast/pull/655> Does it matter which of these PRs is merged first?

Fixes rakudo/rakudo#3815.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -1246,10 +1246,10 @@ static MVMint64 string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *Hay
         if (H_offset < 0)
             H_offset = 0; /* XXX I think this is the right behavior here */
     }
-    /* If the offset is greater or equal to the number of Haystack graphemes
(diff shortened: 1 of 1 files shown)


src/strings/ops.c
```diff
@@ -1249,7 +1249,7 @@ static MVMint64 string_equal_at_ignore_case(MVMThreadContext *tc, MVMString *Hay
     /* If the offset is greater or equal to the number of Haystack graphemes
```
MasterDuke17, line 1249: The comment should be corrected also.

patrickbkr: Thanks for the hint!

patrickbkr:
I think this is ready for merge. I'm just unsure if the roast PR should be merged first.

jnthn:
@patrickbkr I'd get a MoarVM/NQP version bump done and then merge the roast tests, that way folks won't see them failing.

PR #1334, opened 2020-08-06
YYY a better hash

nwc10:
Not for merging. A pull request to track the known issues.

And not yet even a well formatted pull request

1. data race between setting the flag bit `MVM_CF_HAS_OBJECT_ID` (in one thread) and the write barriers setting flag bits (in a different thread). This is not new, but this branch exposes it. It can be replicated on master: https://colabti.org/irclogger/irclogger_log/moarvm?date=2020-08-04
2. strict aliasing bug introduced in MVMHash.c: https://colabti.org/irclogger/irclogger_log/moarvm?date=2020-07-30#l124
3. commit `XXX has XXXes to remove. Please could @jnthn check my assumptions?` needs those `XXX`s checking
4. all the other commits should be reviewed too (for code style, actual errors typos, unclear comments, anything)
5. test failures reported by coke: https://gist.github.com/coke/cc744eade897aaf23c3d4114b93495ec

Attempting to solve the data race by simply eliminating `MVM_CF_HAS_OBJECT_ID` seems to be a 10% performance regression (HEAD on this branch).

Data race can be provoked on master with:

```
--- a/src/gc/objectid.c
+++ b/src/gc/objectid.c
@@ -5,6 +5,9 @@
 MVMuint64 MVM_gc_object_id(MVMThreadContext *tc, MVMObject *obj) {
     MVMuint64 id;

+    assert(!(obj->header.flags & MVM_CF_STABLE));
+    assert(!(obj->header.flags & MVM_CF_FRAME));
+
     /* If it's already in the old generation, just use memory address, as
      * gen2 objects never move. */
     if (obj->header.flags & MVM_CF_SECOND_GEN) {
@@ -15,12 +18,20 @@ MVMuint64 MVM_gc_object_id(MVMThreadContext *tc, MVMObject *obj) {
     else {
         MVMObjectId *entry;
         uv_mutex_lock(&tc->instance->mutex_object_ids);
+        HASH_FIND(hash_handle, tc->instance->object_ids, (void *)&obj,
+                  sizeof(MVMObject *), entry);
         if (obj->header.flags & MVM_CF_HAS_OBJECT_ID) {
             /* Has one, so just look up by address in the hash ID hash. */
-            HASH_FIND(hash_handle, tc->instance->object_ids, (void *)&obj,
-                sizeof(MVMObject *), entry);
+            if (!entry) {
+                MVM_oops(tc, "object_id lookup fail for %p", obj);
+            }
         }
         else {
+            if (entry) {
+                MVM_oops(tc, "object_id insert dup, %p is already %p",
+                         obj, entry->gen2_addr);
+            }
+
             /* Hasn't got one; allocate it a place in gen2 and make an entry
              * in the persistent object ID hash. */
             entry            = MVM_calloc(1, sizeof(MVMObjectId));

```

(it will hit the oops for "insert dup", not the asserts)

I fear that to address the data race we're going to need to use more atomic ops, and to do **that** we're going to need to merge `flags` `size` and maybe `owner` in `struct MVMCollectable` into one integer sized appropriately for what libatomic can offer us.

I think also that the data race means that sometimes `MVM_CF_REF_FROM_GEN2` is not set (when `MVM_gc_object_id` wins and `MVM_gc_write_barrier_hit_by` loses), so it's a real and dangerous bug. But this paragraph has not yet had code review.


src/core/fixkey_hash_table.c
```diff
@@ -0,0 +1,302 @@
...
+    MVMuint32 true_size = hashtable->official_size + hashtable->max_items - 1;
+    if (hashtable->official_size + MVM_HASH_MAX_PROBE_DISTANCE < true_size) {
+        true_size = hashtable->official_size + MVM_HASH_MAX_PROBE_DISTANCE;
+    }
+    return true_size;
+}
+
+/* Frees the entire contents of the hash, leaving you just the hashtable itself,
+   which you allocated (heap, stack, inside another struct, wherever) */
+void MVM_fixkey_hash_demolish(MVMThreadContext *tc, MVMFixKeyHashTable *hashtable) {
```
MasterDuke17, line 17: Nitpick, but most (all?) of the other similar functions are `*_destroy`.

nwc10:
Superseded by #1337

src/core/fixkey_hash_table.c
nwc10: Oops, I missed this comment. My intent was deliberate - they don't **free** memory. Whereas my understanding of `destroy` was that that meant "and frees the memory".

jnthn: They do tend to, indeed (though there may be an exception somewhere). That's reasonable, anyway. Though I note `uv_mutex_destroy` does work like you described, so we're not in a state of complete `destroy` consistency in the codebase (even if libuv's naming isn't really our deciding :)).

PR #1337, opened 2020-08-13
A better hash

nwc10:
This fixes all the known bugs detailed in #1334 

There is still this specific commit that needs checking and then editing

```
commit 90eb6992c03755f533b331abe810b67974068dfc
Author: Nicholas Clark <nick@ccl4.org>
Date:   Wed Jun 3 18:09:32 2020 +0000

    XXX has XXXes to remove. Please could @jnthn check my assumptions?

    Convert MVMHash and HashAttrStoreHashAttrStore to MVMStrHashTable.

    The MVMStrHashTable struct can be stored directly in the object body - we
    no longer need another layer of indirection.

    Add MVM_str_hash_count to return the count of items in the hash.
    Add MVM_str_hash_end to return an iterator for which MVM_str_hash_at_end
    is immediately true.

    For debugging iterators, given that the hashtable structure is now public,
    and can be embedded in other structures, we can't use it as the debugging
    ID. [The GC likes to move things! :-)]

    Hence use a 64 bit random number as the debugging ID. This necessitates moving
    the headers with the relevant function prototypes earlier.
```

Apart from that, I **think** that it is good to merge, but previous review by me, @dogbert17 and @coke found subtle bugs, so there might still be some hiding.

Edit:

The branch `A-better-hash` is byte-for-byte identical to `AAAA-A-better-hash`.

It has

1. One commit pushed after @jnthn's last commit, which adds two oopses (that would otherwise be panics)
2. The commit with XXX rewritten to remove the XXX
3. One other commit message edited to fix some text
4. Three of the code review commits re-ordered to be immediately after the commit that they referred to.

Specifically, the section from my `git log --oneline` file that I used to plan the re-order; the moved commits are indented:

```
480f09cf1 Eliminate uthash. It served us well, and earned a happy retirement.
f8b7defb7 Hash randomisation for MVMStrHashTable.
33edf845d MVM_str_hash_fsck.
    24c86e6ff Fix a typo. In five places.
7c65ae56b Hash iterators now run in "reverse", from the largest index down to 0.
e954e6f76 Re-implement MVMStrHashTable as a Robin Hood Hash.
7bb6a9276 The JIT template for MVM_OP_sp_boolify_iter_hash will no longer work.
889d61d04 Revert "Spesh and JIT get(cur)?hllsym"
4a1c16fe1 Re-implement MVMFixKeyHashTable as a Robin Hood Hash.
ba629fa60 Re-implement MVMPtrHashTable as a Robin Hood Hash.
fd12897b7 Re-implement MVMIndexHashTable as a Robin Hood Hash.
542847255 Refactor MVM_uni_hash_insert in terms of MVM_uni_hash_lvalue_fetch.
53e1fe9c6 Re-implement MVMUniHashTable as a Robin Hood Hash.
9507a098f Convert loaded_compunits from a MVMStrHashTable to a MVMFixKeyHashTable.
778eb8762 Add MVM_string_hash_code, which encapsulates access to the string's hash code.
232260ace Add container_registry's keys in MVM_gc_root_add_instance_roots_to_worklist.
3c8dae717 Remove all of the UT hash functional macros (and most of uthash.h)
c5c67562a Migrate the other internal Unicode hashes to MVMUniHashTable.
3a53f7466 Add MVMUniHashTable, and use it to re-implement codepoints_by_name.
4ff5785a1 Delete more now-unused MVM_HASH macros and inline functions.
a5d58f8a0 Convert `seen` in MVMHeapSnapshotState to MVMPtrHashTable.
141779cb5 Add MVMPtrHashTable, for hashes mapping a pointer to a pointer-sized integer.
5de3c4ddb Eliminate the last two uses of the macro MVM_HASH_KEY
    db62b91e1 Printing the pointer address of the MVMString isn't adding anything.
f287d8927 Convert repr_hash to MVMIndexHashTable, repr_list to repr_names & repr_vtables
3f8a8e5c0 Add MVMIndexHashTable, for hash lookups into an array of MVMString*.
b7665665a Re-order the representations registry into ID order.
c38beb9e8 Replace MVM_str_hash_bind_nt with MVM_str_hash_insert_nt.
17a8a99d3 Replace MVM_str_hash_unbind with MVM_str_hash_delete
263ea58d3 Use MVM_str_hash_lvalue_fetch instead of a fetch/bind pair.
    bafd06faf Comments clarifying why each oops call added makes valid assumptions.
90eb6992c XXX has XXXes to remove. Please could @jnthn check my assumptions?
cff7a0cda Converge the implementations of MVMHash and HashAttrStore
2bf3a4e2b Move code from interp.c to MVMiter.h to avoid breaking encapsulation.
```

I have not rebased it. Commits are only reordered, or have their commit message changed. No commit contents have changed.

@jnthn - I believe that `A-better-hash` is ready to merge.


nwc10:
> as long as a zeroed body is sufficient

I'm not sure if I understand all the details implied by this question (I'm not confident enough of all the context), so I might be answering the wrong question, or answering badly.

If one zeros out the `struct MVMStrHashTable`, then it is safe to call `MVM_str_hash_build` on that memory again.

As implemented, zeroing out the memory (or starting with zeroed memory) and then calling **other** hash functions (without/before) calling `MVM_str_hash_build` first will, um, go horribly wrong, because there does not seem to be a sanity test on `hashtable->entry_size` in `hash_allocate_common`. It will call `MVM_malloc` with a zero size, which **will** panic if the C library returns NULL, but won't if it returns a pointer. (I think that which happens is implementation dependent)

nwc10:
> A paranoid assert just in case `MVM_str_hash_end` doesn't do its job right?

Yes.

But on reflection I don't think that this add value any more. It did during development (when things were messy) so I should remove this one.

jnthn:
> As implemented, zeroing out the memory (or starting with zeroed memory) and then calling other hash functions (without/before) calling MVM_str_hash_build first will, um, go horribly wrong

Presumably `MVM_str_hash_entry_size` an exception to this statement, given you're calling it immediately before `MVM_str_hash_build`?

jnthn:
These two mentions of `uthash` (found with `git grep` could be cleaned up:

```
LICENSE:- uthash.h   MIT
src/strings/ops.c: * implementation in uthash.h. */
```

src/6model/reprs.c
```diff
@@ -271,58 +275,54 @@ void MVM_repr_initialize_registry(MVMThreadContext *tc) {
...
-    if (entry == NULL) {
+    MVMuint32 idx = MVM_index_hash_fetch_nt(tc, &tc->instance->repr_hash, tc->instance->repr_names, name);
+    if (idx == MVM_INDEX_HASH_NOT_FOUND) {
         char *c_name = MVM_string_ascii_encode_any(tc, name);
         char *waste[] = { c_name, NULL };
         uv_mutex_unlock(&tc->instance->mutex_repr_registry);
-        MVM_exception_throw_adhoc_free(tc, waste, "Lookup by name of unknown REPR: %s",
-            c_name);
+        MVM_exception_throw_adhoc_free(tc, waste, "Lookup by name %p of unknown REPR: %s",
+                                       name, c_name);
```
jnthn, line 309: Not quite sure printing the pointer address of the `MVMString` adds anything to this error output?

src/core/str_hash_table.c
```diff
@@ -0,0 +1,494 @@
...
+        ++hashtable->cur_items;
+    }
+    return new_entry;
+}
+
+/* UNCONDITIONALLY creates a new hash entry with the given key and value.
+ * Doesn't check if the key already exists. Use with care.
+ * (well that's the official line. As you can see, the XXX suggests we currently
+ * don't exploit the documented freedom.) */
+void *MVM_str_hash_insert_nt(MVMThreadContext *tc,
```
jnthn, line 248: I don't see the `XXX` this refers to.

jnthn: What does `nt` mean here? At first I thought "null terminated" (we use it for that in some of the string handling stuff) and got worried, but it doesn't seem to mean that here...

src/6model/reprs/MVMIter.c
```diff
@@ -105,12 +105,29 @@ static void shift(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *da
             }
             return;
         case MVM_ITER_MODE_HASH:
+            ; /* Sigh, C99 won't let me put a declaration here. */
+            MVMStrHashTable *hashtable = &(((MVMHash *)target)->body.hashtable);
```
jnthn, line 109: My usual trick is to use a compound statement (wrap the "body" of the `case` in curlies).

src/core/str_hash_table.c
nwc10: `nt` - *no throw*. I don't **think** that I was the first person to add this to the codebase. But I haven't checked. Better name would be fine, but I don't have a great suggestion.

src/6model/reprs.c
nwc10: Agree. That should go.

src/core/str_hash_table.c
jnthn: We use `nocheck` in quite a few places for "doesn't check stuff and just goes ahead and does the operation". May be applicable here.

nwc10:
> > As implemented, zeroing out the memory (or starting with zeroed memory) and then calling other hash functions (without/before) calling MVM_str_hash_build first will, um, go horribly wrong
> 
> Presumably `MVM_str_hash_entry_size` an exception to this statement, given you're calling it immediately before `MVM_str_hash_build`?

Yes. Missed that. It's even documented that it's safe:

`/* If this returns 0, then you have not yet called MVM_str_hash_build */`

But I think that I should add an `oops` to the `initial_alloc` code so that if size isn't set, it goes `oops` instead of multiplying by zero, attempting to allocate zero, and then likely panicing.

src/core/fixkey_hash_table.c
```diff
@@ -0,0 +1,302 @@
...
+
+MVM_STATIC_INLINE void hash_allocate_common(MVMFixKeyHashTable *hashtable) {
+    hashtable->max_items = hashtable->official_size * FIXKEY_LOAD_FACTOR;
+    size_t actual_items = hash_true_size(hashtable);
+    hashtable->entries = MVM_malloc(sizeof(MVMString ***) * actual_items);
+    hashtable->metadata = MVM_calloc(1 + actual_items + 1, 1);
+    /* A sentinel. This marks an occupied slot, at its ideal position. */
+    *hashtable->metadata = 1;
+    ++hashtable->metadata;
+    /* A sentinel at the other end. Again, occupited, ideal position. */
```
jnthn, line 48: occupited

src/core/fixkey_hash_table.h
```diff
@@ -0,0 +1,58 @@
+/* A Better Hash.
+
+A C implementation of https://github.com/martinus/robin-hood-hashing
+by Martin Ankerl <http://martin.ankerl.com>
+
+Better than what we have. Not better than his. His is hard to beat.
```
jnthn, line 6: or "what we had" :-)

src/core/str_hash_table.c
nwc10: Oh bother. Yes, the `XXX` got refactored out. It used to be a comment about roughly "OK, we said that we didn't check for duplicates, but we just did, and now what should we do?". It's now that "oops", which (arguably) still contradicts the official documentation of "UNCONDITIONALLY". "oops and exit" is not the documented interface (of "la la la, I can't hear you")

src/6model/reprs.c
nwc10: Agree. That should be neuralised.

src/6model/reprs/MVMIter.c
nwc10: IIRC I didn't really want to change the indenting of a bunch of lines that I wasn't modifying.

And I still suspect that the C99 spec/grammar is wrong here - if you're going to allow decarations *almost* anywhere, make declarations into statements, and hence permit them everywhere. But that ship sailed (in 1999)

src/debug/debugserver.c
```diff
@@ -1929,36 +1935,18 @@ static MVMint32 request_object_metadata(MVMThreadContext *dtc, cmp_ctx_t *ctx, r
...
-                cmp_write_int(ctx, 0);
-                cmp_write_str(ctx, "mvmhash_ineff_expands", 21);
-                cmp_write_int(ctx, 0);
-            }
+            MVMStrHashTable *hashtable = &(((MVMHashBody *)OBJECT_BODY(target))->hashtable);
+
+            /* FIXME. What stats should we generate? Some are O(1),
+             * some are O(n) (like mean probe length, and its SD) */
+            cmp_write_str(ctx, "mvmhash_num_items", 17);
+            cmp_write_int(ctx, hashtable->cur_items);
```
jnthn, line 1949: @timo Maybe you have some comment here

src/moar.h
```diff
@@ -91,10 +92,13 @@ typedef double   MVMnum64;
 #  define MVM_USED_BY_JIT
 #endif
 
-/* stuff for uthash */
-#define uthash_fatal(msg) MVM_exception_throw_adhoc(tc, "internal hash error: " msg)
+/* Hashes */
+#define HASH_DEBUG_ITER 1
```
jnthn, line 96: Should this be on when we merge?

src/6model/reprs/MVMIter.c
jnthn: I just put the `{` after the `:` and the closing one at the same level as the `case`. :-)

jnthn: And yeah, agree the spec is weird here.

src/moar.h
nwc10: Er, yes, bother, no. As in, no it should not be on when we merge.

Although I **did** wonder about whether the "throw an error when both iterating and modifying the same hash" should be enabled by default. As in go from what we have:

```
typedef struct {
    MVMuint32 pos;
#if HASH_DEBUG_ITER
    MVMuint32 serial;
    MVMuint64 owner;
#endif
}  MVMStrHashIterator;
```
to this instead

```
typedef struct {
    MVMuint32 pos;
    MVMuint32 serial;
#if HASH_DEBUG_ITER
    MVMuint64 owner;
#endif
}  MVMStrHashIterator;
```

and default to something like this error we have disabled by default in `uthash.h`:

```
#if MVM_HASH_THROW_ON_ITER_AFTER_ADD_KEY
    /* Warn if the user has been *caught* inserting keys during an iteration. */
    if (ht->bucket_rand_last != ht->bucket_rand)
        MVM_exception_throw_adhoc(tc, "Warning: trying to insert into the hash while iterating has undefined functionality\n");
#endif
```

except that we can also do it on naughty deletes (ie not at the iterator's current position)

jnthn: I guess that'd be a language-visible change, so we'd want to proceed with caution. Perhaps something to see if we can get away with after we know that shipping the new hash implementation hasn't had unexpected fallout.

jnthn:
Merged branch `A-better-hash` as 51d399cb54, effectively merging this PR.

PR #1339, opened 2020-08-25
Update libuv to v1.39.0

dogbert17:
Spectest clean under Linux Mint (Ulyana) 64-bit


jnthn (approved):
Thanks, good to get this in well before the next release to allow more testing time.

PR #1340, opened 2020-08-27
make moar --dump use memmem to find "MOARVM\r\n" string

timo:
this is the magic marker that starts a moarvm bytecode file.

precomp folders have a bit of information prepended to the
moar bytecode files, and having to manually cut it off can
be tedious. Since moar --dump is a user-facing and debugging
feature, it should be fine to have it do a skipping or
searching step first.

making this a pull request because i want an opinion on whether it's fine to turn the MVM_memmem function into a define.

### DIFF
--- a/src/moar.c
+++ b/src/moar.c
@@ -1,8 +1,10 @@
+#include "platform/memmem.h"
 #include "moar.h"
 #include "platform/io.h"
 #include <platform/threads.h>
 #include "platform/random.h"
 #include "platform/time.h"
+#include "platform/mmap.h"
 #if defined(_MSC_VER)
 #define snprintf _snprintf
 #endif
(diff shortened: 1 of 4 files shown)


src/moar.c
```diff
@@ -478,7 +480,42 @@ void MVM_vm_run_bytecode(MVMInstance *instance, MVMuint8 *bytes, MVMuint32 size)
...
+    /* Map the bytecode file into memory. */
+    if ((fd = uv_fs_open(NULL, &req, filename, O_RDONLY, 0, NULL)) < 0) {
+        MVM_exception_throw_adhoc(tc, "While trying to open '%s': %s", filename, uv_strerror(req.result));
+    }
+
+    block = MVM_platform_map_file(fd, &handle, (size_t)size, 0);
+
+    if (block == NULL) {
+        /* FIXME: check errno or GetLastError() */
+        MVM_exception_throw_adhoc(tc, "Could not map file '%s' into memory: %s", filename, "FIXME");
```
jnthn, line 504: Maybe fix this? :-)

src/platform/memmem.h
```diff
@@ -15,9 +15,7 @@
 #include <string.h>
 #endif
 
-void * MVM_memmem(const void *haystack, size_t haystacklen, const void *needle, size_t needlelen) {
-    return memmem(haystack, haystacklen, needle, needlelen);
-}
+#define MVM_memmem memmem
```
jnthn, line 18: Hmm, wonder why it wasn't this way before...

jnthn:
This is very useful. :+1:

PR #1341, opened 2020-08-28
Fix MVM_file_isexecutable() when being root

patrickbkr:
Different from the readability and writeability tests, executability is not
affected by the user being root. As long as the `x` flag isn't set, not
even root can execute a file.

Fixes Raku/nqp#652

I'm unsure how one could write a test for this fix. It requires root access to test the behaviour. Any ideas?

### DIFF
--- a/src/io/fileops.c
+++ b/src/io/fileops.c
@@ -290,7 +290,18 @@ static int are_we_group_member(MVMThreadContext *tc, gid_t group) {
     }
 FILE_IS(readable, R)
 FILE_IS(writable, W)
-FILE_IS(executable, X)
+    MVMint64 MVM_file_isexecutable (MVMThreadContext *tc, MVMString *filename, MVMint32 use_lstat) {
+        if (!MVM_file_exists(tc, filename, use_lstat))
+            return 0;
+        else {
+            uv_stat_t statbuf = file_info(tc, filename, use_lstat);
+            MVMint64 r = (statbuf.st_mode & S_IXOTH)
+                      || (statbuf.st_uid == geteuid() && (statbuf.st_mode & S_IXUSR))
+                      || (are_we_group_member(tc, statbuf.st_gid) && (statbuf.st_mode & S_IXGRP))
+                      || (geteuid() == 0 && (statbuf.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH)));
+            return r ? 1 : 0;
+        }
+    }
 #endif
 
 /* Get a MoarVM file handle representing one of the standard streams */
--- a/src/io/fileops.c
+++ b/src/io/fileops.c
@@ -13,6 +13,19 @@
 #define DEFAULT_MODE _S_IWRITE /* work around sucky libuv defaults */
 #endif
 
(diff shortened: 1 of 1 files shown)


src/io/fileops.c
```diff
@@ -290,7 +290,17 @@ static int are_we_group_member(MVMThreadContext *tc, gid_t group) {
     }
 FILE_IS(readable, R)
 FILE_IS(writable, W)
-FILE_IS(executable, X)
+    MVMint64 MVM_file_isexecutable (MVMThreadContext *tc, MVMString *filename, MVMint32 use_lstat) {
+        if (!MVM_file_exists(tc, filename, use_lstat))
+            return 0;
+        else {
+            uv_stat_t statbuf = file_info(tc, filename, use_lstat);
+            MVMint64 r = (statbuf.st_mode & S_IXOTH)
```
jnthn, line 298: Hm, can we not determine whether the file exists using the result in `uv_stat_t`? That'd seem to avoid two system calls, plus it's a race between the exists check and asking for the file info so it can't be relied upon anyway.

patrickbkr:
I have written a small test script that tests the following scenarios against the expected results:

- Raku via  `.IO.r/w/x`
- Raku by actually trying via `.slurp`, `.spurt`, `run`
- Python via `os.access()`
- Python by actually trying via `.read()`, `.write()`, `subprocess.call()`

All four giving consistent results in all of the 36 permission combinations I was able to come up with (user/group/other, user/root, can/can't access, read/write/execute).

The exception being: If the program we are trying to execute is a script (in contrast to a native executable), the `r` permission is needed in addition to the `x` permission to execute it. Neither Raku, nor Python (basically POSIX `access()`) catch that.

I'm not sure if and where these tests could be integrated into a test suite. They do rely on being able to become the root user.

Any opinions?

src/io/fileops.c
patrickbkr: I have refactored the file respectively.

src/io/fileops.c
```diff
@@ -202,48 +215,47 @@ MVMint64 MVM_file_exists(MVMThreadContext *tc, MVMString *f, MVMint32 use_lstat)
...
-                    char *pext = getenv("PATHEXT");
-                    int plen   = strlen(pext);
-                    int i;
-                    for (i = 0; i < plen; i++) {
-                        if (0 == stricmp(ext, pext++)) {
-                             r = 1;
-                             break;
-                        }
+        /* true if fileext is in PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC */
+        MVMString *dot = MVM_string_ascii_decode_nt(tc, tc->instance->VMString, ".");
```
jnthn, line 238: As a minor optimization, this could be one of the strings that we create at startup and cache on the `MVMInstance`.

jnthn (approved):

jnthn:
> I'm not sure if and where these tests could be integrated into a test suite. They do rely on being able to become the root user.

Tricky; do any of our CI environments allow for that?

jnthn:
Thanks for the improvements; merged. If wanting to do the minor perf improvement, feel free to submit it as a follow-up PR.

PR #1344, opened 2020-09-07
Convert MVMSpeshCandidate to a REPR

MasterDuke17:
This will allow them to be safely removed if desired (e.g., there were too many deopts for a candidate).


src/6model/reprs/MVMSpeshCandidate.c
```diff
@@ -308,6 +314,7 @@ void MVM_spesh_candidate_add(MVMThreadContext *tc, MVMSpeshPlanned *p) {
             spesh->body.spesh_candidates);
     }
     new_candidate_list[spesh->body.num_spesh_candidates] = candidate;
```
zhuomingliang, line 320:
```suggestion
```
MVM_ASSIGN_REF does `=`, so this line can be removed.

MasterDuke17: Thanks, removed it.

src/6model/reprs/MVMSpeshCandidate.c
```diff
@@ -0,0 +1,366 @@
...
+
+/* Initialize the representation. */
+const MVMREPROps * MVMSpeshCandidate_initialize(MVMThreadContext *tc) {
+    return &SpeshCandidate_this_repr;
+}
+
+static void describe_refs(MVMThreadContext *tc, MVMHeapSnapshotState *ss, MVMSTable *st, void *data) {
+    /* TODO */
+}
+
```
jnthn, line 28: If not planning to fill this out, it's better not to implement it at all, so the heap snapshot can fall back on the default GC marking approach.

src/6model/reprs/MVMSpeshCandidate.c
```diff
@@ -0,0 +1,366 @@
...
+        memcpy(new_candidate_list, spesh->body.spesh_candidates, orig_size);
+        MVM_fixed_size_free_at_safepoint(tc, tc->instance->fsa, orig_size,
+            spesh->body.spesh_candidates);
+    }
+    MVM_ASSIGN_REF(tc, &(spesh->common.header), new_candidate_list[spesh->body.num_spesh_candidates], candidate);
+    spesh->body.spesh_candidates = new_candidate_list;
+
+    /* May now be referencing nursery objects, so barrier just in case. */
+    //if (spesh->common.header.flags2 & MVM_CF_SECOND_GEN)
+        //MVM_gc_write_barrier_hit(tc, (MVMCollectable *)spesh);
```
jnthn, line 325: I think this can safely go away now; we know that we just allocated the `MVMSpeshCandidate` object in the nursery.

src/6model/reprs/MVMSpeshCandidate.c
```diff
@@ -0,0 +1,366 @@
...
+ * need to ignore these ones from here on. */
+void MVM_spesh_candidate_discard_existing(MVMThreadContext *tc, MVMStaticFrame *sf) {
+    MVMStaticFrameSpesh *spesh = sf->body.spesh;
+    if (spesh) {
+        MVMuint32 num_candidates = spesh->body.num_spesh_candidates;
+        MVMuint32 i;
+        for (i = 0; i < num_candidates; i++)
+            spesh->body.spesh_candidates[i]->body.discarded = 1;
+        MVM_spesh_arg_guard_discard(tc, sf);
+    }
```
jnthn, line 365: We'll be able to do this rather more properly (actually discard them) once we get that mechanism working for deopt (for now though, I guess this stays as it is).

src/6model/reprs/MVMStaticFrameSpesh.c
```diff
@@ -95,25 +77,25 @@ static MVMuint64 unmanaged_size(MVMThreadContext *tc, MVMSTable *st, void *data)
...
 
         /* XXX probably don't need to measure the bytecode size here,
          * as it's probably just a pointer to the same bytecode we have in
          * the static frame anyway. */
 
         /* Dive into the jit code */
-        if (cand->jitcode) {
-            MVMJitCode *code = cand->jitcode;
+        if (cand->body.jitcode) {
+            MVMJitCode *code = cand->body.jitcode;
```
jnthn, line 98: This lot is wrong now, I think; the spesh candidate is its own GC-able object, and the unmanaged memory size is accounted for there. So we should not account for it in this object also.

src/6model/reprs/MVMStaticFrameSpesh.c
```diff
@@ -138,13 +120,13 @@ static void describe_refs(MVMThreadContext *tc, MVMHeapSnapshotState *ss, MVMSTa
...
-                    (MVMCollectable *)body->spesh_candidates[i]->spesh_slots[j],
+                    (MVMCollectable *)body->spesh_candidates[i]->body.spesh_slots[j],
                     "Spesh slot entry");
-            for (j = 0; j < body->spesh_candidates[i]->num_inlines; j++)
+            for (j = 0; j < body->spesh_candidates[i]->body.num_inlines; j++)
                 MVM_profile_heap_add_collectable_rel_const_cstr(tc, ss,
-                    (MVMCollectable *)body->spesh_candidates[i]->inlines[j].sf,
+                    (MVMCollectable *)body->spesh_candidates[i]->body.inlines[j].sf,
                     "Spesh inlined static frame");
         }
```
jnthn, line 131: This describe refs logic wants to move into the `SpeshCandidate` REPR also.

src/spesh/deopt.c
```diff
@@ -162,19 +162,19 @@ static void uninline(MVMThreadContext *tc, MVMFrame *f, MVMSpeshCandidate *cand,
...
-        *materialized = MVM_calloc(MVM_VECTOR_ELEMS(cand->deopt_pea.materialize_info), sizeof(MVMObject *));
+        *materialized = MVM_calloc(MVM_VECTOR_ELEMS(cand->body.deopt_pea.materialize_info), sizeof(MVMObject *));
     if (!(*materialized)[info_idx]) {
-        MVMSpeshPEAMaterializeInfo *mi = &(cand->deopt_pea.materialize_info[info_idx]);
-        MVMSTable *st = (MVMSTable *)cand->spesh_slots[mi->stable_sslot];
+        MVMSpeshPEAMaterializeInfo *mi = &(cand->body.deopt_pea.materialize_info[info_idx]);
+        MVMSTable *st = (MVMSTable *)cand->body.spesh_slots[mi->stable_sslot];
         MVMP6opaqueREPRData *repr_data = (MVMP6opaqueREPRData *)st->REPR_data;
         MVMROOT(tc, f, {
             MVMObject *obj = MVM_gc_allocate_object(tc, st);
```
jnthn, line 180: `cand` can become outdated here (and probably we need to check in the callers about this also)

src/spesh/inline.c
```diff
@@ -270,14 +270,18 @@ MVMSpeshGraph * MVM_spesh_inline_try_get_graph(MVMThreadContext *tc, MVMSpeshGra
 
     /* Build graph from the already-specialized bytecode and check if we can
      * inline the graph. */
-    ig = MVM_spesh_graph_create_from_cand(tc, target_sf, cand, 0, &deopt_usage_ins);
+    MVMROOT(tc, cand, {
+        ig = MVM_spesh_graph_create_from_cand(tc, target_sf, cand, 0, &deopt_usage_ins);
+    });
```
jnthn, line 275: I think this doesn't allocate.

src/spesh/inline.c
```diff
@@ -270,14 +270,18 @@ MVMSpeshGraph * MVM_spesh_inline_try_get_graph(MVMThreadContext *tc, MVMSpeshGra
...
         /* We can inline it. Do facts discovery, which also sets usage counts.
          * We also need to bump counts for any inline's code_ref_reg to make
          * sure it stays available for deopt. */
         MVMuint32 i;
-        MVM_spesh_facts_discover(tc, ig, NULL, 1);
-        add_deopt_usages(tc, ig, cand->deopt_usage_info, deopt_usage_ins);
+        MVMROOT(tc, cand, {
+            MVM_spesh_facts_discover(tc, ig, NULL, 1);
+        });
+        add_deopt_usages(tc, ig, cand->body.deopt_usage_info, deopt_usage_ins);
```
jnthn, line 284: Pretty sure this doesn't allocate either.

MasterDuke17: @niner thinks it does https://github.com/MoarVM/MoarVM/blob/master/tools/check-roots.py#L230

niner: I generated that list with `cflow --number --brief --reverse  src/*.c src/*/*.c src/*/*/*.c  > call_graph` and then extracted the sub graph resulting in a run_gc call from that huge file.

The path it found is MVM_spesh_facts_discover -> add_bb_facts -> wvalfrom_facts -> MVM_sc_get_object -> MVM_serialization_demand_object -> stub_object -> read_object_table_entry -> MVM_sc_get_stable -> MVM_serialization_demand_stable -> stub_stable -> MVM_gc_allocate_stable

But, looking at wvalfrom_facts, the call to MVM_sc_get_object is protected by a MVM_sc_is_object_immediately_available condition. So at least this path is incorrect.

I guess we would need some custom annotations to get a better list. But then, adding unnecessary MVMROOTs hurts less than missing some.

src/spesh/inline.c
MasterDuke17: Not sure why I added it, removing it doesn't cause check-roots to complain.

src/spesh/inline.c
MasterDuke17: Or a list of false-positive-causing functions in check-roots.py? Anyway, removing this one since we've confirmed it's not needed.

src/spesh/deopt.c
MasterDuke17: You (just) mean it needs to be added to the MVMROOT?

jnthn: Yes, here, but also check if there's a caller of this function that holds a pointer to `cand`, and that would thus require an `MVMROOT` in the caller also.

src/6model/reprs/MVMSpeshCandidate.c
```diff
@@ -93,13 +103,40 @@ static void deserialize_stable_size(MVMThreadContext *tc, MVMSTable *st, MVMSeri
...
+
+    size += sizeof(MVMint32) * body->num_deopts;
+
+    size += sizeof(MVMSpeshInline) * body->num_inlines;
+
+    size += sizeof(MVMuint16) * (body->num_locals + body->num_lexicals);
+
+    /* XXX probably don't need to measure the bytecode size here,
+     * as it's probably just a pointer to the same bytecode we have in
+     * the static frame anyway. */
```
jnthn, line 122: We do need to measure it here, I think, as its the specialized bytecode; the static frame is not copied per specialization. So this comment can go.

jnthn (approved):
Everything I requested to be addressed has been, so I'm content with this (aside from one comment that I think can be deleted).

MasterDuke17:
@jnthn what are your thoughts re squashing all/some of the commits before merging vs leaving them in?

niner:
I'd be all for squashing commits. Even for the review I quickly switched to the consolidated diff, as there's no point reading code and (in best case) spotting bugs, just to see that the bugs were fixed in a later commit anyway. In general the commit history should reflect the shortest way from the previous to the new code in atomic steps. I.e. each commit should have merit on its own,  be as small as possible, be buildable and - except for extreme cases - also runnable (to facilitate bisection).

MasterDuke17:
> In general the commit history should reflect the shortest way from the previous to the new code in atomic steps. I.e. each commit should have merit on its own, be as small as possible, be buildable and - except for extreme cases - also runnable (to facilitate bisection).

In theory all my commits are exactly like this...In practice...

Yes, I'm in favor of squashing the commits in this PR. It is a little annoying to lose some of the messages associated with some of the individual changes. They don't make as much sense as comments in the code, and it's harder to point out exactly which smaller, non-atomic change is being referred to in a large commit message. It would be nice if there was some way to attach a message to parts of the diff of a commit.

PR #1347, opened 2020-09-17
copy_to should call MVM_gc_write_barrier with the *new* key's address.

nwc10:
This is a regression introduced since the last release by commit 9d0f1b326d8d:
    Replace MVM_str_hash_bind_nt with MVM_str_hash_insert_nt.

(I think. Please can someone double check this)


nwc10:
After some sleep, I realise that this isn't as bad as I thought it is.

Specifically, I'm confusing values with addresses. The GC's *root macros take the *address* of the C variable, and manipulate that. Whereas the write barrier code is using the value from the C variable and "taking action" based on that. It becomes clearer when you look at what it is doing - inspecting and maybe updating the flags on the `MVMCollectable` that the value points to. And the hash code here is *copying* a pointer from source hash to destination hash - the same value - the address of the same `MVMCollectable`. So it can't matter.

(This week)

It's a style issue.

(Until/unless someone changes the hash code to copy keys. Which seems unlikely)

Still, please could someone else check.

jnthn (approved):
I think this is in the "abundance of caution"; the way write barriers work today, it won't matter much.

PR #1349, opened 2020-09-21
oops if MVM_fixed_size_alloc() is called for a size of 0 bytes.

nwc10:
Previously this wasn't trapped. With FSA_SIZE_DEBUG enabled (not the default)
everything works (it can allocate "0" bytes, as it adds its debugging record
keeping to the size requested). With default settings, we corrupt the malloc
heap, which doesn't get reported at the time, and may or may not get reported
in any useful way later.


niner (approved):

nwc10:
rebased.

jnthn (approved):

PR #1352, opened 2020-09-23
Simplify and optimize isprime_I

MasterDuke17:
Remove the `rounds` option. Rakudo only ever used a single (too large)
value and the op does not appear to be used in the ecosystem. Instead,
hardcode the value passed to the libtommath function to 40, chosen based
on https://security.stackexchange.com/questions/4544/how-many-iterations-of-rabin-miller-should-be-used-to-generate-cryptographic-saf
Also, libtommath now returns false for 1, so remove that check in our
code.

Additionally, add a branch for !MVM_BIGINT_IS_BIG, and handle that case
with the FJ32_256 algorithm from Forišek and Jančina, "Fast Primality
Testing for Integers That Fit into a Machine Word", 2015.

After making the single necessary modification to Rakudo,
`say (^100_000).grep(*.is-prime).tail; say now - INIT now` goes from
~12.8s to ~0.036s. The speedup isn't as dramatic for big integers,
`say (10_000_000_000..10_000_100_000).grep(*.is-prime).tail; say now - INIT now`
goes from ~26s to ~11s.


AlexDaniel (commented):
This is really great!

I find this comment rather interesting: https://github.com/rakudo/rakudo/pull/3923#issuecomment-697326385 Maybe we can plan a switch to GMP for the future?

MasterDuke17:
> Maybe we can plan a switch to GMP for the future?

I'm not aware of any technical or licensing impossibility, but it would likely be a lot of grunt work.

MasterDuke17:
15:02 | jnthn | MasterDuke:  We ended up with libtommath partly out of license concerns (although  LGPL is probably OK; I always forget exactly what it enforces), but in a  bigger part because of libtommath being far easier to build across  platforms.
15:03 | jnthn | MasterDuke: Of note, stock gmp doesn't build on the MSVC compiler.
15:03 | jnthn | There is these days a fork that does: http://mpir.org/

AlexDaniel:
In other words there wasn't any strong justification, more like an arranged marriage and we just ended up with libtommath.

Just to get a feel of both, I wrote these two programs:

```c
#include <gmp.h>
#include <stdio.h>

mpz_t number1;
mpz_t number2;


int main() {
    mpz_init_set_ui(number1, 0);
    mpz_init_set_ui(number2, 1);

    for (size_t i = 0; i < 10000000; i++) {
        mpz_add(number1, number1, number2);
    }

    printf("heh %lu\n", mpz_get_ui(number1));
    printf("heh %lu\n", mpz_get_ui(number2));

    return 0;
}
```

```c
#include <tommath.h>
#include <stdio.h>

mp_int number1;
mp_int number2;

int main() {
    if (mp_init_ul(&number1, 0) != MP_OKAY)
        printf("whoopsie!");

    if (mp_init_ul(&number2, 1) != MP_OKAY)
        printf("whoopsie!");

    for (size_t i = 0; i < 10000000; i++) {
        if (mp_add(&number1, &number2, &number1) != MP_OKAY)
            printf("whoopsie!");
    }

    printf("heh %u\n", mp_get_u32(&number1));
    printf("heh %u\n", mp_get_u32(&number2));

    return 0;
}
```

Basically, just init two bigints to 0 and 1, and then keep adding the second one to the first one 10_000_000 times. Nothing exciting, but here are the timings:

GMP
```
heh 100000000
heh 1

real	0m0.984s
user	0m0.981s
sys	0m0.001s
```

LibTomMath
```
heh 100000000
heh 1

real	0m1.331s
user	0m1.330s
sys	0m0.001s
```

The error checking with libtommath is required with `-Werror`, but I also tried without it and saw no difference in timings.

I do not know what to think of this, microbenchmarks are kinda nasty. But if basic bigint arithmetic is 1.35x as fast with GMP, then the grunt work may as well be justified.

Also, I think GMP supports rational numbers? Mmmm.

lizmat:
> In other words there wasn't any strong justification

Again, this remark is not needed and unnecessarily abrasive (again).  If you read the conversation a bit further on IRC, you'd see that @jnthn was working on Windows at the time, using the MSVC compiler.  Not being supported on the compiler that you're using, would be an **excellent** reason for not selecting that library.

jnthn:
> In other words there wasn't any strong justification, more like an arranged marriage and we just ended up with libtommath.

@AlexDaniel Not building on a major platform / toolchain - especially when it was, at the time, the primary development toolchain of the MoarVM founders - seems strong enough as a reason to pick an alternative that did satisfy that constraint. It's not like it committed MoarVM to always use it, if the situation changed (as it seems to now have done so).

As @lizmat points out, your remark was also unnecessarily abrasive and unconstructive. I've had quite enough of behaviors like:

* Painting anybody who made a decision differently than you imagine you would have under the same constraints as doing things without thought
* Using terms like "naysayer" to describe people who present valid concerns about something you'd like to see happen
* Using loaded or offensive language, both on IRC and in issues
* Disrespectfully dismissing points put forward by others with responses like "this is not about having great debating skills"

Which have sucked much of the joy out of contributing to Raku for a number of people by this point. For now, you continue to get away with this behavior in a Raku context, and ultimately it shall be for the Raku Steering Council to decide if it wishes to continue to entertain it.

So far as MoarVM goes, it's *my* call to make. Please refrain from such behavior; a further instance in the context of MoarVM development shall be grounds for banning you from this project.

AlexDaniel:
@jnthn I am for honest and direct language that doesn't mislead readers and in fact calls for action to change the situation. The idea of picking a specific library for the project just because you used MSVC at the time isn't a good justification, and there is no need to double down. Raku project is riddled with decisions that were seemingly random at the time but are now extremely hard to undo/redo, here we are just scratching the surface. How about starting with “yeah, maybe it would've been better if we did X, we can perhaps still do it” instead of providing a long description with multiple reasons that have no relevance today. Especially if you want to insist that everything should be *your* call, except that you're making an extremely small amount of said calls.

Granted, in [full log](https://colabti.org/irclogger/irclogger_log/moarvm?date=2020-09-23#l63) you did say that maybe it should be done, but you still seem to be implying that MSVC is a hard requirement (as if… better performance is less important?):


> jnthn | **Given that exists**, we maybe can think about a move to gmp, though of course it needs a vict^Wvolunteer...

Meanwhile, MSVC support doesn't even [seem to be](https://github.com/JuliaLang/julia/issues/12056#issuecomment-487768933) planned in Julia.


> Which have sucked much of the joy out of contributing to Raku for a number of people by this point.

Raku project had 20 years of happy mindless progress in no particular direction. Unsurprisingly, it didn't get anywhere we all want it to be. Accepting the current situation is not a pleasant act. We can't always be happy, it is life. IMO the biggest question for the council right now is what exactly Raku should be and *how can we get there*, but you can of course spend the energy on trying to exclude me because you don't like the things I say and how I say them directly and openly. By the way, I clearly expressed my frustration with the current state of things in [my nomination](https://github.com/Raku/Raku-Steering-Council/blob/main/nominations/2020/AlexDaniel.md). 


image

lizmat:
@AlexDaniel

> I am for honest and direct language that doesn't mislead readers

Here you are implying that @jnthn is misleading readers.  Is that what you intended to convey?

> calls for action to change the situation

This is **OPEN SOURCE**.  If you have an itch, you scratch it **YOURSELF**.  And that is **NOT** telling other people how things should be.

> The idea of picking a specific library for the project just because you used MSVC at the time isn't a good justification

Within the given constraints at the time, it was a **PERFECTLY VALID** decision to make.  And as such justified.  This is independent of the fact that there may be better solutions available *now*.

> Raku project is riddled with decisions that were seemingly random at the time

The reasons for some decisions may have been lost in time.  But just stating that *random decisions" were made at the time, is doing a disservice to **ALL** of the people who have put in a significant amount of work to get a system going.  Possibly, some of these decisions in the past, would have been made different today.  But to say that they have been random, again, is showing that you do **NOT** know what you're talking about.

> How about starting with “yeah, maybe it would've been better if we did X, we can perhaps still do it” instead of providing a long description with multiple reasons that have no relevance today.

How about you start not being abrasive and constructive for a change?  How about you start actually providing PR's, instead of just opening issues.  How about you taking [@taboege's example](https://github.com/MoarVM/MoarVM/pull/1353)?  Now **THAT** is being constructive.

> Especially if you want to insist that everything should be your call, except that you're making an extremely small amount of said calls.

I'm pretty sure that @jnthn **WILL** make a call after this.

> Meanwhile, MSVC support doesn't even seem to be planned in Julia.

So what does *that* have to do with anything?

> Raku project had 20 years of happy mindless progress in no particular direction

**OH BOY**.  Thank you for smearing the professional standing of all the people involved in the Raku project over the past 20 years.

> Unsurprisingly, it didn't get anywhere we all want it to be.

And who is the "we" you speak of?  Is that your "majestic plural"?

> Accepting the current situation is not a pleasant act

And **WHO** is saying that nothing should change?  Or that nothing is changing?  Have you looked at the work in the `RakuAST` and `newdisp` branches?  How, e.g. the former should allow us to get rid of string EVAL in the core for things like `sprintf` (already done), `assuming` (planned), `regex interpolation` (planned) and module selection (planned).

> The biggest question for the council right now is what exactly Raku should be and how can we get there

An interesting question.  Perhaps a recent [/r/rakulang thread](https://www.reddit.com/r/rakulang/comments/ixo408/how_would_you_describe_raku_in_one_sentence/) shows that different people find different things important.

> you can of course spend the energy on trying to exclude me because you don't like the things I say and how I say them directly and openly

Saying things directly and openly is **NOT** a guarantee to being constructive.  If anything, in the Raku community, as hopefully in *any* community, people discussing issues are expected to do so in a constructive way.  Generally, you are **NOT** being constructive, you're just busy shouting **FIRE**!!

> By the way, I clearly expressed my frustration with the current state of things in my nomination.

I think the result of the election shows that only 1/3 of the voters actually wants you on the Steering Council.  As opposed to @jnthn (about 90%) and me (about 75%).  Doesn't that tell you something?

Since you were against the Raku Steering Council from the beginning, always claiming that the problems solving repo would be the place to raise issues and have discussions, I would suggest you consider *not* accepting the Raku Steering council seat, and to thereby offering @vrurg to take a seat on the Steering Council.  I think that would the Raku Steering Council to function in a way that would be productive, while not taking away your voice from discussions prior to being handled by the Raku Steering Council.

AlexDaniel:
Relax. I just made a comment saying the decision wasn't justified. Big deal. Now that there's a PR moving things in that direction I don't see a lot of need to discuss this more (honestly, wouldn't've left a previous comment if I knew there was a PR, I missed it).

lizmat:
> I just made a comment saying the decision wasn't justified

Which clearly indicates that you do not consider what such a comment means.

> Big deal

Yes, this **IS** a big deal.  You've become noise.  Do you want to be considered "noise"?   I would hope that you want to be taken seriously.  If so, then please, do not just say things that just come up in your mind.  Think about what they would mean in a wider world.  How people might interpret it.

>  I don't see a lot of need to discuss this more 

Indeed.  The original subject is now well underway of becoming an improvement to Rakudo on MoarVM.  But things might still go sideways: there is no full implementation yet, nor do we have a certainty that it will all build on all of the architectures Rakudo supports.

> (honestly, I wouldn't've left the previous comment if I knew there was a PR, I missed it).

But you **DID** miss it.  And you **DID** assume nothing was happening.  And you **DID** just shout **FIRE!!**.  Even managed to meme a GIF for it and include it.  And you're not apologizing for this.  You're just saying that **YOU** are not interested in pursuing the discussion.

This is yet another reason why you, in your current state of just being able to shout **FIRE!!**, are **NOT** a constructive member of the Raku Community, let alone of the Raku Steering Council.

Clean up your act!  It is clear that you are an intelligent person.  Use that intelligence in a constructive manner.  It will help you in life.

jnthn:
@AlexDaniel Yesterday I asked that you stop attacking other contributors in your interactions here. You immediately responded by doing more of the same. The reply from @lizmat points out the many problems with your response, however this:

> Raku project is riddled with decisions that were seemingly random 

And this:

> Raku project had 20 years of happy mindless progress in no particular direction. 

Are a clear continuation of an explicit behavior I called out as unacceptable, namely:

> Painting anybody who made a decision differently than you imagine you would have under the same constraints as doing things without thought

The development of the Raku language took place very much in public - more so than many other languages - and the history of that process remains accessible to this day. It shows the progression from initial ideas and design principles, through numerous changes based on feedback from implementation and early adopters, to reach the language as it exists today. You state that:

> I am for honest and direct language that doesn't mislead readers 

However, it's hard for me to see how "random" and "mindless" can be considered honest conclusions to draw based on the readily available project history.

Further, it feels to me that "direct language" really means not being concerned about the impact of ones words on other people.  However, people are at the heart of a volunteer run open source project. MoarVM has seen commits from ~120 individuals, and there are surely others who have not produced a commit, but have provided valuable input in other ways. I'm grateful to all of them.

From the start of MoarVM, I've aimed for a professional and friendly community around it. I'm quite sure I've made my own missteps in trying to achieve that, but generally I'd say that those working on MoarVM have done pretty well at providing feedback on each others work without anyone feeling put down or attacked as a result. I will strive for that to continue to be the case.

Which brings me to a sad first in this project: the first time in its 7 year history where I believe the best way forward is to exclude somebody from further participation. On at least one thing I do agree with @AlexDaniel: we can't always be happy. Without happiness, the ban on participation proposed yesterday is thus implemented.

jnthn (approved):
Looks good to me. Guess it needs NQP and Rakudo patches that match, along with version bumps. Wonder if the NQP one needs to account for other backends too...

MasterDuke17:
> Guess it needs NQP and Rakudo patches that match, along with version bumps. Wonder if the NQP one needs to account for other backends too...

Ok, I'll make PRs for NQP and Rakudo, and then update them with a version bump once this is merged.

Yeah, I've looked at the other backends and I think the changes are trivial.

Issue #321, opened 2015-12-28
A request to put 3rd party libraries into "hidden" locations to prevent conflicts


mojca:
I'm trying to make a package for MoarVM (for MacPorts). The huge advantage of your official tarball is that it comes with "batteries included", with all the required 3rd party libraries that might be needed. The problem is that it also installs all the files of included libraries to "standard" locations, so that the files installed by MoarVM conflict with already installed libraries. If I use the built-in `dyncall` for example, I'm later not able to install the `dyncall` package from the package manager (or if I have it already installed, it would not let me install MoarVM).

Many software packages would put the libraries somewhere else. As an example, wxWidgets would install

```
${prefix}/include/wx-3.0/wx/dynload.h
```

while MoarVM would install

```
${prefix}/include/dyncall/dynload.h
```

which causes a problem/conflict.

I didn't try to create a complete list, but installation locations of these libraries seem a tiny bit problematic: `dyncall`, `libatomic_ops`, `libtommath`, `libuv`, `msinttypes`, `tinymt`, `sha1`.

coke:
See also https://trac.macports.org/ticket/50153

PR #1353, opened 2020-09-24
Use GMP instead of libtommath

taboege:
Bundle, build and use GMP instead of libtommath for speed.

### DIFF
--- a/build/setup.pm
+++ b/build/setup.pm
@@ -26,6 +26,13 @@ our %TP_TOM = (
     src  => [ '3rdparty/libtommath' ],
 );
 
+our %TP_GMP = (
+    name => 'gmp',
+    path => '3rdparty/gmp',
+    rule  => 'cd 3rdparty/gmp && sh .bootstrap && CC=\'$(CC)\' CFLAGS=\'$(CFLAGS)\' ./configure @crossconf@ && $(MAKE) && cp .libs/libgmp.a . && cd ..',
+    clean => 'cd 3rdparty/gmp && $(MAKE) clean && $(RM) libgmp.a',
+);
+
 our %TP_MT = (
     name => 'tinymt',
     path => '3rdparty/tinymt',
@@ -76,6 +83,7 @@ our %TP_UV = (
 our %THIRDPARTY = (
     lao => { %TP_LAO },
     tom => { %TP_TOM },
+    gmp => { %TP_GMP },
     sha => { %TP_SHA },
     mt  => { %TP_MT },
     dc  => { %TP_DC },
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "3rdparty/libatomicops"]
 	path = 3rdparty/libatomicops
 	url = https://github.com/MoarVM/libatomic_ops
+[submodule "3rdparty/gmp"]
+	path = 3rdparty/gmp
+	url = https://github.com/ShiftMediaProject/gmp
--- a/Configure.pl
+++ b/Configure.pl
@@ -44,7 +44,7 @@ GetOptions(\%args, qw(
     debug:s optimize:s instrument! coverage
     os=s shell=s toolchain=s compiler=s
     ar=s cc=s ld=s make=s has-sha has-libuv
-    static has-libtommath has-libatomic_ops
+    static has-libtommath has-gmp has-libatomic_ops
     has-dyncall has-libffi pkgconfig=s
     build=s host=s big-endian jit! enable-jit
     prefix=s bindir=s libdir=s mastdir=s
@@ -92,8 +92,8 @@ if ( $args{relocatable} && ($^O eq 'aix' || $^O eq 'openbsd') ) {
     ".\n    Leave off the --relocatable flag to do a non-relocatable build.");
 }
 
-for (qw(coverage instrument static big-endian has-libtommath has-sha has-libuv
-        has-libatomic_ops asan ubsan tsan valgrind dtrace show-vec)) {
+for (qw(coverage instrument static big-endian has-libtommath has-gmp has-sha
+        has-libuv has-libatomic_ops asan ubsan tsan valgrind dtrace show-vec)) {
     $args{$_} = 0 unless defined $args{$_};
 }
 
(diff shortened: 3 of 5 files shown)


Altai-man:
Please, don't merge until next week.

taboege:
This is still at the beginning. At the moment it pulls in and builds GMP successfully on my computer. I use the repository at ShiftMediaProject/gmp linked by @MasterDuke17 but don't do anything MSVC-specific yet.

The first thing I want to ask at this point is: looking at #321, I build and link a static library version of GMP to moarvm and throw it away afterwards. No headers or shared library of GMP are installed, making moar supposedly fatter. Is that the right direction or should a shared library be installed in location that makes it clear that this is MoarVM's GMP library?

Also @MasterDuke17 : do you have a small patch that partially replaces libtommath with GMP so that I can try out my changes as I go?

MasterDuke17:
> Also @MasterDuke17 : do you have a small patch that partially replaces libtommath with GMP so that I can try out my changes as I go?

Not yet, I was waiting for a way to build it to test (bit of a chicken-and-egg problem here). But I can check out your branch and give it a whirl. Probably won't be until later today, but I can try to get you something to test with.

build/setup.pm
```diff
@@ -26,6 +26,13 @@ our %TP_TOM = (
     src  => [ '3rdparty/libtommath' ],
 );
 
+our %TP_GMP = (
+    name => 'gmp',
+    path => '3rdparty/gmp',
+    rule  => 'cd 3rdparty/gmp && sh .bootstrap && CC=\'$(CC)\' CFLAGS=\'$(CFLAGS)\' ./configure @crossconf@ && $(MAKE) && cp .libs/libgmp.a . && cd ..',
```
MasterDuke17, line 32: Do we want the `--enable-fat` option (or a way to pass it through from MoarVM's Configure.pl)? AIUI, that builds binaries that choose the right versions of functions at runtime depending on the type of cpu. For someone compiling from source for themselves it wouldn't make a difference, but it seems like it would be good when packagers are compiling for distribution.

taboege: Added a pass-through option in da20fd3.

nwc10:
I tried to build this branch on the sparc64 system [gcc202](https://cfarm.tetaneutral.net/machines/list/). It fails like this:

```
 ccache gcc -c -DHAVE_CONFIG_H -I. -I.. -D__GMP_WITHIN_GMP -I.. -DOPERATION_add_n -std=gnu99 -Wextra -Wall -Wno-unused-parameter -Wno-unused-function -Wno-missing-braces -Werror=pointer-arith -Og -DNDEBUG -g3 -D_REENTRANT -D_FILE_OFFSET_BITS=64 -fPIC -DDEBUG_HELPERS -DWORDS_BIGENDIAN -DMVM_HEAPSNAPSHOT_FORMAT=2 -DMVM_TRACING=0 -DMVM_CGOTO=1 -DMVM_RDTSCP=0 -Wa,--noexecstack tmp-add_n.s -fPIC -DPIC -o .libs/add_n.o
tmp-add_n.s: Assembler messages:
tmp-add_n.s:106: Error: Architecture mismatch on "addxccc %o4,%o5,%l7".
tmp-add_n.s:106: (Requires v9d|v9e|v9v|v9m|m8; requested architecture is v9.)
tmp-add_n.s:115: Error: Architecture mismatch on "addxccc %g1,%g5,%l7".
tmp-add_n.s:115: (Requires v9d|v9e|v9v|v9m|m8; requested architecture is v9.)
tmp-add_n.s:125: Error: Architecture mismatch on "addxccc %o4,%o5,%g3".
tmp-add_n.s:125: (Requires v9d|v9e|v9v|v9m|m8; requested architecture is v9.)
tmp-add_n.s:128: Error: Architecture mismatch on "addxccc %g1,%g5,%g5".
tmp-add_n.s:128: (Requires v9d|v9e|v9v|v9m|m8; requested architecture is v9.)
tmp-add_n.s:138: Error: Architecture mismatch on "addxccc %o4,%o5,%g3".
tmp-add_n.s:138: (Requires v9d|v9e|v9v|v9m|m8; requested architecture is v9.)
tmp-add_n.s:142: Error: Architecture mismatch on "addxc %g0,%g0,%i0".
tmp-add_n.s:142: (Requires v9d|v9e|v9v|v9m|m8; requested architecture is v9.)
make[3]: *** [Makefile:772: add_n.lo] Error 1
make[3]: Leaving directory '/home/nick/Perl/MoarVM/3rdparty/gmp/mpn'
make[2]: *** [Makefile:996: all-recursive] Error 1
make[2]: Leaving directory '/home/nick/Perl/MoarVM/3rdparty/gmp'
make[1]: *** [Makefile:786: all] Error 2
make[1]: Leaving directory '/home/nick/Perl/MoarVM/3rdparty/gmp'
make: *** [Makefile:743: 3rdparty/gmp/libgmp.a] Error 2
```

Looking carefully at what it is complaining about, it seems to be saying that you can't use this assembler instruction on the current CPU, which appears to be "sparc v9 whatever", but it is on every *specific* sparc v9. Oh, and v8 too. So, sigh, it seems to be a sort of assembler bug-ish.

*But then*

About two steps later I realised that *gmp is a dependency of gcc* and I'm building this with gcc, right? So it has to have worked at some point.

*To cut to the chase*

I cloned https://github.com/ShiftMediaProject/gmp directly and built it with defaults. It builds. *There* I see these compiler lines like this (sorry, different targets. Lots of scrollback):

```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I.. -D__GMP_WITHIN_GMP -I.. -O2 -pedantic -m64 -mptr64 -mcpu=niagara4 -Wa,-Av9d -c obprintf.c -o obprintf.o >/dev/null 2>&1
```

Note there it has a bunch of `-m` flags to specify exactly the CPU.

So I think what is going wrong here is that, as-is, we're *overriding* the CPU flags for its make stage. Instead, *adding* might seem better, except that - we're building a static library here, right? (with Position Independent Code). So we never expose it? In which case, I think that the best answer is simply not to override its `CFLAGS`

MasterDuke17:
Looks like you're missing a `-$(CMD)@gmpclean@ $(NOOUT) $(NOERR)` in build/Makefile.in

PR #1360, opened 2020-10-13
Hash bits in metadata

nwc10:
Store some hash bits in the low bits of the hashtable metadata byte, with the probe distance in the high bits, and the split between the two updated dynamically.

This works (where tested, with ASAN and with assertions)

Currently it's not much faster. I had to resort to cachegrind to see any difference - it seems to dispatch more instructions, but result in about 0.1% fewer writes misses, and 0.1% fewer last-level cache misses.

This is disappointing. I hope that the numbers can be improved with some tuning. There is a *lot* that can be tuned

1. initial choice for the split probe distance:hash bits
1. how many times to move the split (to increase the probe distance) before growing the hash
1. for each hash grow, whether to "reset" the probe distance fully, "back off" (go somewhat smaller) or just keep it the same
1. whether to pick different values if the hash is given an expected size when first built
1. load factor (this of course could also be used to tune the code on master)

and all these can be varied as a function of hash size (eg it might work best to start with a notional "8" entry hash with a high load factor, then reduce the load factor for medium sized hashes, and then raise it again for large hashes, on the assumption that they really stress the CPU cache)

**Update** - it *is* a win for last-level cache lookups. The trade is more CPU work for less data misses.
I've subsequently added a few more optimisations (again CPU work for fewer data misses) that I believe help further.

- ~~test on "awkward" platforms (eg sparc64 - I think that index hashes might currently bust alignment assumptions there)~~ - they did. Fixed in the rebase
- ~~a "total memory used" API function as requested by \@timo~~ - done. cachegrind thinks that this causes fractionally more last level cache lookup fails
- ~~test out using the FSA instead of regular malloc - at least the first two sizes allocated fit into the normal buckets~~ -- done. cachegrind likes this
- ~~allocate just a control structure to start - thanks \@lizmat~~ done. cachegrind likes this too
- ~~architecture comment explaining how the heck this works~~ done. feedback wanted

What the github UI doesn't seem to make obvious is that commit d53a4a12bc26c21cfb783e5a53ecf1a76d8f0783 on this branch is origin/hash-single-allocation and that branch *is* ready for review and merging - #1350

I think that this branch is ready for *review*, but I'd prefer not to merge it until after the next release.

### DIFF
--- a/src/core/str_hash_table.h
+++ b/src/core/str_hash_table.h
@@ -178,6 +178,237 @@ be worth it.
  * 3) fix up the GC invariants
  */
 
+/* How does storing hash bits in the metadata work? And how come there doesn't
+ * seem to be *any* explicit code for testing it in the fetch loop?
+ *
+ * Consider the 6 keys 'R', 'a', 'k', 'u' 'd' 'o'.
+ * We'll assume that the hash value for each is their ASCII code.
+ * Assuming a hash with 16 buckets, we use the top 4 bits of the hash value
+ * to determine the ideal bucket for the key - the bucket that the key would go
+ * in, if inserted into an empty hash. So our keys look like this:
+ *
+ *           hash   bucket   extra
+ * R     01010010        5       2
+ * a     01100001        6       1
+ * k     01101011        6      11
+ * u     01110101        7       5
+ * d     01100100        6       8
+ * o     01101111        6      15
+ *      /   ||   \
+ *     bucket extra
+ *
+ * If we insert them in that order into the hash, then they would be laid out
+ * like this:
+ *
+ *     +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+ *     | A | B | C | D | E | F | G | H | I | J | K | L | M | N | O | P | ...
+ *     +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+ * key                       R   a   k   d   o   u
+ * probe distance            1   1   2   3   4   4
+ *
+ *
+ * Insert order doesn't matter - this order would be equally valid and would be
+ * what we get by inserting the 6 keys in reverse order:
+ *
+ *     +---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+---+
+ *     | A | B | C | D | E | F | G | H | I | J | K | L | M | N | O | P | ...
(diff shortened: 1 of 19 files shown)


lizmat:
Cool work!

I think it's important to have a very well performing *empty* hash (as that is the most common created hash in method calls to keep `*%_` in, and small (like 8 keys) hash, as that sub / method calls usually do not have more than 8 named parameters.

nwc10:
> Cool work!
> 
> I think it's important to have a very well performing _empty_ hash (as that is the most common created hash in method calls to keep `*%_` in, and small (like 8 keys) hash, as that sub / method calls usually do not have more than 8 named parameters.

Good point. I had thought "it's OK - small initial alloc is... " and then realised that it's the control structure, actually 15 pointers and 16 bytes of metadata. Which is OK - only the control structure is actually going to be hot in the CPU cache.

But I think that it's relatively easy to start an empty hash as *just* being a control structure. "flag" this as both `cur_items` 0 and `max_items` 0, and not worry that `official_size_log2` can't actually represent 0 entries (or, heck, that there isn't even the bonus metadata byte for the sentinel, which would round up to a pointers-worth of memory)

This should work fairly cleanly because

- `cur_items` tested for 0 is (or is supposed to be) a fast-path return "nope" test for all the RVALUE API functions
- `max_items` set to 0 is a the trigger value for "must call `maybe_grow_hash`" for all the LVALUE API functions

meaning that the special casing for this needs only to be in the allocation and free functions.

(And inevitably no plan survives contact with the enemy - there will turn out to be code paths that don't yet have these checks, which will need to be found with asserts and then rectified)

src/core/str_hash_table.h
```diff
@@ -176,19 +178,267 @@ Not all the optimisations described above are in place yet. Starting with
...
+ *                                  >= 0x2b (so keep going)
+ *
+ *                                       ^
+ *                                      == 0x3b (so might match)
+ *                                      ('k' eq 'k', so found; return)
+ *
+ * Notice that we didn't even have to do the string comparison for 'k' != 'a'
+ * and for 'k' != 'd'. They fall out naturally from the existing byte
+ * comparisons, and how the metadata is stored.
+ *
```
jnthn, line 293: Goodness, that's a nice trick. Thanks for writing it up.

jnthn (approved):

jnthn:
Thanks for documenting and explaining the various trade-offs; this was an interesting read.

PR #1363, opened 2020-10-22
`char` can be unsigned. Rewrite the "is it ASCII?" test to handle thi…

nwc10:
…s too.

Without this gcc correctly warns that the "comparison is always false" on
platforms where plain char is unsigned, and the code (as was) was buggy.

On x86_64 (where plain char is signed) this commit makes no difference to
the generated code.


lizmat:
For later: appears to be directly committed with https://github.com/MoarVM/MoarVM/commit/a989f7b553

PR #1368, opened 2020-10-25
Use MVM_{malloc,realloc,calloc,free} consistency.

nwc10:
All these calls changed to the MVM_* wrappers are on memory blocks where the
other calls are MVM_* wrappers.

There are still a few "bare" `malloc`s/`free`s, but these seem to be
self-contained, or related to memory allocated by `strdup`/`strndup`, which
is using the platform `malloc` internally.


niner (approved):

nwc10:
Rebased to master and pushed live.

PR #1371, opened 2020-10-26
`swap_bytes` in utf16.c needs to also swap bytes on big endian platforms

nwc10:
The way it had been written worked perfectly on little endian platforms, and
was actually a no-op on big endian platforms, because it *read* from memory
in fixed (little endian) order, but wrote back using bitwise operations,
meaning that the read acted as a second byte swap on big endian platforms.
This was somewhat LTA.


nwc10:
Pushed as fast-forward.

PR #1373, opened 2020-10-30
Fix interning of parametrics

niner:
Interning of parametric types was only implemented for the most simple cases. This ought to make it work reliably in all cases.


dogbert17:
FWIW, passes

    make stresstest

running under ASAN and with the following changes:

    dogbert@dogbert-VirtualBox:~/repos/rakudo/nqp/MoarVM$ git diff
    diff --git a/src/core/fixedsizealloc.c b/src/core/fixedsizealloc.c
    index 504e5ed0e..ce1aeb194 100644
    --- a/src/core/fixedsizealloc.c
    +++ b/src/core/fixedsizealloc.c
    @@ -10,7 +10,7 @@
      * behavior. */
     
     /* Turn this on to switch to a mode where we debug by size. */
    -#define FSA_SIZE_DEBUG 0
    +#define FSA_SIZE_DEBUG 1
     #if FSA_SIZE_DEBUG
     typedef struct {
         MVMuint64 alloc_size;
    diff --git a/src/gc/collect.h b/src/gc/collect.h
    index 153805096..2a1644638 100644
    --- a/src/gc/collect.h
    +++ b/src/gc/collect.h
    @@ -1,6 +1,6 @@
     /* The maximum size of the nursery area. Note that since it's semi-space
      * copying, we could actually have double this amount allocated per thread. */
    -#define MVM_NURSERY_SIZE 4194304
    +#define MVM_NURSERY_SIZE 8192
     
     /* The nursery size threads other than the main thread start out with. If
      * they fill it and trigger a GC run, then it is doubled. If they are
    diff --git a/src/gc/debug.h b/src/gc/debug.h
    index 91d833abd..20735652d 100644
    --- a/src/gc/debug.h
    +++ b/src/gc/debug.h
    @@ -4,7 +4,7 @@
      * 2 = Checks on every object register access (slow)
      * 3 = Collects garbage on every allocation
      */
    -#define MVM_GC_DEBUG 0
    +#define MVM_GC_DEBUG 1
     
     #if MVM_GC_DEBUG
     #define MVM_ASSERT_NOT_FROMSPACE(tc, c) do { \

PR #1374, opened 2020-10-31
Eliminate static race condition

nwc10:
The code in master:src/6model/serialization.c uses a C `static` variable to pass the running thread's `MVMThreadContext` pointer into the C callback function passed to `qsort`. This is a subtle race condition - if two threads are serialising (ie compiling) at the same time, they might see the wrong value.

We should instead use some form of thread local storage to pass this value into the callback. C11 and later provide a `_Thread_local` attribute which makes all of this (almost) transparent. Before this gcc (and its impersonators) provided `__thread` which works in the same way. It seems also that all Win32 compilers that UV supports provide an attribute which we can use. If there is nothing better (likely vendor compilers on OSes we don't yet reach) we fall back to UV's API, which seems to be a thin wrapper around pthreads.

I had hoped that the `_Thread_local` approach would be significantly faster than the classic pthreads approach, and about as fast as the static global, but it seems that this is only true when compiling regular unshared code. MoarVM is set up to build a shared library by default, and the implication of this is that there still needs to be a call through the shared library redirection tables to get the correct thread locals for this thread for this process (to something like `__tls_get_addr(PLT)`). It looks like TLS is *slightly* faster - for ARM, with it the callback compiles to 11 instructions, with UV (wrapping pthreads) its 12. The called function *might* also be simpler (I haven't checked), and there is no need to call `uv_key_create` to create a key each runtime - the setup is now done at C compilation time.

(And it ought to be 10 instructions, because the compiler seems to miss the opportunity to exploit the tailcall. We left it an open goal and still it missed :\-()

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -490,12 +490,14 @@ print "OK\n\n";
 if ($config{crossconf}) {
     build::auto::detect_cross(\%config, \%defaults);
     build::probe::static_inline_cross(\%config, \%defaults);
+    build::probe::thread_local_cross(\%config, \%defaults);
     build::probe::unaligned_access_cross(\%config, \%defaults);
     build::probe::ptr_size_cross(\%config, \%defaults);
 }
 else {
     build::auto::detect_native(\%config, \%defaults);
     build::probe::static_inline_native(\%config, \%defaults);
+    build::probe::thread_local_native(\%config, \%defaults);
     build::probe::unaligned_access(\%config, \%defaults);
     build::probe::ptr_size_native(\%config, \%defaults);
 }
--- a/build/config.h.in
+++ b/build/config.h.in
@@ -46,6 +46,11 @@
 /* How this compiler does static inline functions. */
 #define MVM_STATIC_INLINE @static_inline@
 
+#if @has_thread_local@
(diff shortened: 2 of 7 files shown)


nwc10:
Oh great. :\-(

Thanks - I thought I'd found and fixed that. Must have managed to end up with the version where it was wrong.

nwc10:
That's interesting. That implies that the fallback code works on Win32 too. That wasn't the plan.

jnthn (approved):
Makes sense; merging it now means we've most of the release window to spot any unexpected platform issues, but since it falls back to libuv anyway, they seem unlikely.

There is [some code here](https://github.com/MoarVM/MoarVM/blob/master/src/core/nativecall.c#L1153) that could go away in favor of this mechanism also, I think, which might make native callbacks rather faster.

nwc10:
> There is [some code here](https://github.com/MoarVM/MoarVM/blob/master/src/core/nativecall.c#L1153) that could go away in favor of this mechanism also, I think, which might make native callbacks rather faster.

See #1391

I don't know how to benchmark this. But it's obviously simpler.

PR #1377, opened 2020-11-05
we need to import pthread on windows to get a type

timo:
we have an `#include <pthread.h>` in the setname_np probe already, so if we want to use it, we should also include it there.

this blew up moarvm build on MinGW


patrickbkr:
Can this be merged? We are a good way away from the next release, so it's a good moment to get this in.

PR #1378, opened 2020-11-07
MVM_box_{int,num,str,uint} and MVM_unbox_str duplicate functionality in reprconv.c

nwc10:
MVM_box_{int,num,str,uint} and MVM_unbox_str duplicate functionality in reprconv.c and can be replaced with calls to MVM_repr_box_{int,num,str,uint} and MVM_repr_get_str.

PR #1379, opened 2020-11-09
Eliminate the intcache; add "fastcreate" style logic to MVM_repr_box_int

nwc10:
With this branch this code now works:

    $ ./rakudo-m -e 'my int $a = 1; my int $b = 1; $a does role { method Str() { "foo" } }; $b.Str.say'
    1

It eliminates the intcache, which returned singleton MVMObjects when boxing ints in the range (-1, 15), which caused the above code to output `foo`.

To mitigate some of the slowdown this causes, it adds fast-path object allocation and initialisation for `P6int` and `P6opaque` in `MVM_repr_box_int`, similar to how various spesh ops and the JIT implement it. I think that the total slowdown is about 1%, but it would be useful it someone else double checks this. On the upside, there's less code, and two fewer spesh ops.

It *doesn't* fix this bug, which is somewhere in NQP or Rakudo, given that `rakudo-j` gives the same result:

    $ ./rakudo-m -e 'my $a = 1; my $b = 1; $a does role { method Str() { "foo" } }; $b.Str.say'
    foo

This follows on from the branch MVMROOT-in-box-foo in #1378


jnthn:
> $ ./rakudo-m -e 'my int $a = 1; my int $b = 1; $a does role { method Str() { "foo" } }; $b.Str.say'
> 1

Really it's the Raku spec that wants refining, in order to to provide a means to refuse mixing in to an immutable `Int` in the first place. It's already a DIHWIDT today; one should really be using `but` in this situation. I don't think losing the small integer cache is the right trade-off for a behavior that we should probably find a way to forbid.

lizmat:
Basically forbid `does` on objects that have a `.WHICH` that is a `ValueObjAt`?

niner:
@jnthn the int cache is also a blocker for in-process precompilation.
Though I may have an idea that could maybe with luck get around this and an issue with parametrics (which share the singleton behaviour though are far less prevalent).

jnthn:
> the int cache is also a blocker for in-process precompilation.

Hm, ideally they'd be owned by the core VM serialization context (same as the various BOOT type objects).

jnthn:
> Basically forbid does on objects that have a .WHICH that is a ValueObjAt?

Perhaps those two should always happen together, but that might be the tail wagging the dog...

MasterDuke17:
> I don't think losing the small integer cache is the right trade-off for a behavior that we should probably find a way to forbid.

Aside from the behavior change/bugfix, "I think that the total slowdown is about 1%, but it would be useful it someone else double checks this. On the upside, there's less code, and two fewer spesh ops.", **if** it turns out the slowdown is small enough (for whatever value of enough we decide), is it worth it to simplify the code overall?

jnthn:
> I think that the total slowdown is about 1%, but it would be useful it someone else double checks this

I'm curious what benchmark was used (and can't find it in the commit messages or the PR message; did I miss it?) A microbenchmark would need writing very carefully (and the specializer output analyzing), because it'd be very easy to write something like `for ^1_000_000 { my int $i = 10; my $box = $i; }`, and then find that the specializer turns it into an empty loop, and thus only the pre-specialization impact is measured.

Even a benchmark that doesn't fall into that trap could still potentially give quite an under-estimate, however, because a GC run in a single-threaded benchmark with a small heap is going to be rather cheaper than one in a larger application with a bigger heap and potentially threads to sync up during the stop the world phase.

MasterDuke17:
> I'm curious what benchmark was used (and can't find it in the commit messages or the PR message; did I miss it?)

FYI, I did some non-rigorous testing (by looking at the timing of a couple runs for stage parse of compiling CORE.c) of @niner's patch to just disable the int cache. I saw ~0.5s slowdown (out of ~35s) and ~20mb increase in maxrss according to /usr/bin/time.

nwc10:
On the assumption that the desired language behaviour was to have the `does` construction work without sideways leakage, I was working on the assumption that the intcache is incompatible with correctness. Hence I really wanted just a ballpark for "how bad?". It's the user + system CPU times for the entire regression test suite (on the same machine, with only this branch as the change)

Whilst your points about carefully benchmarking loops is totally valid (and I'd not considered given the reasoning in the previous paragraph), the branch also changes pretty much everything else that boxes ints, meaning that all values outside of (-1, 15) should be slightly faster, and it also removes a small amount of code in other places (adds up to about 200 lines). So I was equally interested in the macro benchmark.

Whatever the *language* level decision is (to ensure correctness of `$b` in the examples) I think that most of the refactoring here still makes sense, as there was a *lot* of copy-paste repetition in the code that used the intcache - it is all a lot cleaner if it's inside `MVM_repr_box_int`. ie it's not wasted.

We don't have a good macro benchmark do we - "the how much does everything *else* change" question?

nwc10:
obsoleted by #1448

PR #1380, opened 2020-11-13
Implement serialize/deserialize of CStr REPR

jnthn:
As reported in https://github.com/rakudo/rakudo/issues/2209 the lack
of this can sometimes end up blocking precompilation.

### DIFF
--- a/src/6model/reprs/CStr.c
+++ b/src/6model/reprs/CStr.c
@@ -71,6 +71,17 @@ static void deserialize_stable_size(MVMThreadContext *tc, MVMSTable *st, MVMSeri
     st->size = sizeof(MVMCStr);
 }
 
+static void deserialize(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *data, MVMSerializationReader *reader) {
+    MVMString *orig = MVM_serialization_read_str(tc, reader);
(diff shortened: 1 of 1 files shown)


niner (approved):

src/6model/reprs/CStr.c
```diff
@@ -71,6 +71,17 @@ static void deserialize_stable_size(MVMThreadContext *tc, MVMSTable *st, MVMSeri
     st->size = sizeof(MVMCStr);
 }
 
+static void deserialize(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *data, MVMSerializationReader *reader) {
+    MVMString *orig = MVM_serialization_read_str(tc, reader);
+    MVMCStrBody *body = (MVMCStrBody *)data;
+    MVM_ASSIGN_REF(tc, &(root->header), body->orig, orig);
```
niner, line 77: Isn't using MVM_ASSIGN_REF here actually overly cautious? The deserialized string will be allocated in gen2 directly, so a write barrier should never been needed.
Not saying that we should make it less cautious. Just checking my understanding.

jnthn: Your understanding is correct.

PR #1384, opened 2020-11-18
Minimally exact bigint/bigint => num conversion, including rounding.

nwc10:
Given that Rakudo parses decimal constants such as 0.01 as `Rat`s (ie represents them internally as (1 / 100), and persists with this exact representation as long as possible), this effectively makes the operator that converts `Rat`s to `Num`s Rakudo's decimal => binary floating point conversion.

Decimal to binary floating point conversion is messy, inherently lossy (in the general case), subject to errors if not done carefully and correctly (subtle errors such as drift - failure to round trip), and potentially slow, and can require considering a lot more digits than seems intuitive for correct rounding in the extreme corner cases.

(It's something approaching 750 decimal digits for the absolute worst case value, to determine the correct rounding direction where the true value is very close to the boundary between the nearest two values representable in IEEE binary64, but I can no longer find the citation with the guilty numbers.)

The previous code in `MVM_bigint_div_num` calculated the integer division with 64 extra bits of precision, then implicitly truncated when converting to a double. This handled most cases correctly, but one of the spectests failed on 32 bit systems. (arm, ppc32, x86 - so not a compiler or CPU bug).

As described in the revised commit message, there is no right answer. Going beyond 64 bits doesn't improve the situation.

Hence this commit calculates the exact extra precision needed to ensure that the integer division generates the needed mantissa directly, and then performs the necessary rounding before directly writing out a complete floating point representation.

For decimal conversions, it is bit-for-bit consistent with David M. Gay's `strtod` (which seems to be taken as the gold standard in conversions), not just for "normal" numbers, but also values that overflow, underflow (into subnormals, zero or negative zero), and handles all the rounding correctly for every test value I throw at it. (ie the last bit of the mantissa is "correct", where "correct" is consistent with `strtod`).

Worst case this code will increase the inputs to the integer division by 2\*\*54 (ie 17 decimal digits), whereas the previous code would always use 2\*\*64 (ie 20 decimal digits, but always). The rest of the operations are logical operations and bitshifts on integers. It should be slightly faster, but more importantly I believe it to be canonically correct.

### DIFF
--- a/src/math/bigintops.c
+++ b/src/math/bigintops.c
@@ -3,12 +3,12 @@
 #include "tinymt64.h"
 #include <math.h>
 
-#ifndef MANTISSA_BITS_IN_DOUBLE
 #define MANTISSA_BITS_IN_DOUBLE 53
-#endif
-#ifndef MAX_BIGINT_BITS_IN_DOUBLE
-#define MAX_BIGINT_BITS_IN_DOUBLE 1023
-#endif
+#define EXPONENT_SHIFT 52
+#define EXPONENT_MIN -1022
+#define EXPONENT_MAX 1023
+#define EXPONENT_BIAS 1023
+#define EXPONENT_ZERO 0
 
 #ifndef MAX
     #define MAX(x,y) ((x)>(y)?(x):(y))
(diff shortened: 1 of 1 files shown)


nwc10:
I think that this means that `mp_get_double` in libtomath is inherently buggy for some corner cases, and really 2 of our 3 other calls to it should use this approach. As the 3rd is the value passed to `pow` for the exponent, we can probably leave it as is. `$x ** $y` for `$y > 9007199254740992` is going to overflow for `$x > 1.00000000000008`, and  values between 1.00000000000008 and 1 aren't really something we can represent very well anyway, so a little bit of error in `$y` isn't going to change anything.

MasterDuke17:
> I think that this means that `mp_get_double` in libtomath is inherently buggy for some corner cases, and really 2 of our 3 other calls to it should use this approach.

In my convert-to-gmp branch (https://github.com/MoarVM/MoarVM/compare/master...MasterDuke17:taboege-configure-gmp) I just `s/mp_get_double/mpz_get_d/g`. From the gmp docs:

"Convert op to a double, truncating if necessary (i.e. rounding towards zero).

If the exponent from the conversion is too big, the result is system dependent. An infinity is returned where available. A hardware overflow trap may or may not occur."

Is this going to be any better?

nwc10:
It consistently returns with `return negative ? -MVM_NUM_NEGINF : MVM_NUM_POSINF;` in these cases, and i think that MoarVM always defines that as a constant, not a calculation.

So, I guess that this is "better" because it's consistent. I assume that gmp has to be portable to platforms which aren't using IEEE arithmetic, whereas MoarVM is somewhat self-selecting here - we can't actually bootstrap on any platform which isn't IEEE because our stage 0 files are assume it. (And slightly more so - we are assuming that the IEEE bit and byte endiannesss etc is exactly the same as the platform's integer endianness, which is probably true these days. And I only know this because i cut my teeth on a platform where this was *not* true.)

I'm not certain about the "hardware overflow trap" - I didn't check how `MVM_NUM_NEGINF` and `MVM_NUM_POSINF` are actually defined, but we don't seem to hit it currently. Even if we did, as the code is building IEEE doubles from raw bits, we could just return `0xFFF00000000000000000` : `0x7FF00000000000000000` and be done \:-)

(I didn't use to know how to figure those out.)

jnthn (approved):

PR #1385, opened 2020-11-19
Add a test configuration for MinGW on Windows

patrickbkr:
This uses the MinGW that is provided by the AzureCI `windows-2019` image.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -29,7 +29,7 @@ stages:
            RAKUDO_OPTIONS: ''
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: ''
-         Win_MVM_relocatable:
+         Win_MVM_reloc:
            IMAGE_NAME: 'windows-2019'
            RELOCATABLE: 'yes'
            RAKUDO_OPTIONS: '--relocatable'
@@ -62,44 +62,59 @@ stages:
 
          MVM_gcc_njit:
            IMAGE_NAME: 'ubuntu-20.04'
+           RAKUDO_OPTIONS: ''
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--no-jit --cc=gcc'
          MVM_gcc:
            IMAGE_NAME: 'ubuntu-20.04'
+           RAKUDO_OPTIONS: ''
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--cc=gcc'
          MVM_clang_njit:
            IMAGE_NAME: 'ubuntu-20.04'
+           RAKUDO_OPTIONS: ''
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--no-jit --cc=clang'
          MVM_clang:
            IMAGE_NAME: 'ubuntu-20.04'
+           RAKUDO_OPTIONS: ''
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--cc=clang'
          MVM_gcc_njit_libffi:
            IMAGE_NAME: 'ubuntu-20.04'
+           RAKUDO_OPTIONS: ''
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--has-libffi --no-jit --cc=gcc'
          MVM_gcc_libffi:
            IMAGE_NAME: 'ubuntu-20.04'
+           RAKUDO_OPTIONS: ''
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--has-libffi --cc=gcc'
          MVM_clang_njit_libffi:
            IMAGE_NAME: 'ubuntu-20.04'
+           RAKUDO_OPTIONS: ''
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--has-libffi --no-jit --cc=clang'
          MVM_clang_libffi:
            IMAGE_NAME: 'ubuntu-20.04'
+           RAKUDO_OPTIONS: ''
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--has-libffi --cc=clang'
+         MVM_gcc_mingw:
+           IMAGE_NAME: 'windows-2019'
+           RAKUDO_OPTIONS: ''
+           NQP_OPTIONS: '--backends=moar'
+           MOAR_OPTIONS: '--os=mingw32'
 #         MVM_coverage:
 #           IMAGE_NAME: 'ubuntu-20.04'
+#           RAKUDO_OPTIONS: ''
 #           NQP_OPTIONS: '--backends=moar'
 #           MOAR_OPTIONS: '--compiler=clang --coverage --optimize=0 --debug=3 --cc=clang'
 #           COVERAGE: 'yes'
 
          MVM_Ub_18_gcc:
            IMAGE_NAME: 'ubuntu-18.04'
+           RAKUDO_OPTIONS: ''
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--cc=gcc'
          MVM_Ub_18_clang:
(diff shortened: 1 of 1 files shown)


patrickbkr:
There are three failing tests in NQP.

```
Test Summary Report
-------------------
t\hll/06-sprintf.t                   (Wstat: 0 Tests: 294 Failed: 3)
  Failed tests:  40, 119-120
Files=145, Tests=13408, 34 wallclock secs ( 1.05 usr +  0.20 sys =  1.25 CPU)
Result: FAIL
```

Since [rakudo#4009](https://github.com/rakudo/rakudo/issues/4009) is marked as a blocker and depends on this, this is a blocker as well, I guess.

My gut feeling tells me this is probably a problem in MoarVM and not in NQP.

@Altai-man: I'm unsure who is best to ping. Any idea?

Altai-man:
My opinion is:
- Given we know it actually builds, the issue is no longer a blocker for this release. We can delay merging this until tests are fixed.
- Fixing those tests is a blocker for next release respectively.

nwc10:
> There are three failing tests in NQP.
> 
> ```
> Test Summary Report
> -------------------
> t\hll/06-sprintf.t                   (Wstat: 0 Tests: 294 Failed: 3)
>   Failed tests:  40, 119-120
> Files=145, Tests=13408, 34 wallclock secs ( 1.05 usr +  0.20 sys =  1.25 CPU)
> Result: FAIL
> ```

Do we get any more diagnostics than that? (If there's a link to longer output, I failed to spot it).

I don't have any access to MinGW (or even Windows) but it's easier to help remote debug when it (the test) gives some more clues.

Altai-man:
ping @patrickbkr ?

patrickbkr:
The above snippet was copied from the failed Azure build of this PR. The logs seem to be gone already. There wasn't any more info in them though.
@nwc10 I do have access to a windows system. I can try to run it later.

patrickbkr:
@nwc10 <https://gist.github.com/patrickbkr/e36b87dd4522f74856ea9eb8c6d43046>

patrickbkr:
For the record: the failures are caused by slight deviations in very big and very small numbers.

nwc10:
I'm aware that this is too late for yesterday's release, but I have now had a little bit of time to look (but not enough to write any patch)

This failure:

```ok 39 - %d on bigint
not ok 40 - %d on negative bigint
#   Actual value: <-13190017482579450>
```

is this code:

`is(nqp::sprintf('<%d>', [$large-negative-int]), '<-13190017482579452>', '%d on negative bigint');`

where we have

`my $large-negative-int := nqp::fromnum_I(nqp::neg_n(nqp::pow_n(2.42, 42.0)), $knowhow);`

and that simply is not a good way to set up an integer constant. The example of Irix dates this, but I know that it would evaluate 2.0\**31.0 as 214748364**7**.0. One cannot rely on total accuracy on floating point operations outside of (IIRC) `+`, `-`, `*`, `/` modulo and `sqrt`. (I don't mean rounding here - i mean that you might not get the numerically correct 53rd bit in the mantissa, before rounding even comes into play)

One can't rely on floating point exponentiation to set up integers correctly. So the test needs rewriting to express the integer in integer operations. I'm not sure what the test is trying to test, versus what it needs to avoid, so I don't know if the test should simply write a literal `13190017482579450` and then negate that, or if it should get to that value with some *integer* (or even floating point) *arithmetic* on smaller values. I believe that even a loop of 42 multiplies should technically work, but it wouldn't be clear as to why it was being done that way.

nwc10:
The other two failures:

```
not ok 119 - %e handles small numbers
#   Actual value: <3.141e-20>
# Expected value: <3.142e-20>
not ok 120 - %e handles small negative numbers
#   Actual value: <-3.141e-20>
# Expected value: <-3.142e-20>
```

are going to have the same cause, given:

```
is(nqp::sprintf('<%7.3e>', [3.1415e-20]), '<3.142e-20>', '%e handles small numbers');
is(nqp::sprintf('<%7.3e>', [-3.1415e-20]), '<-3.142e-20>', '%e handles small negative numbers');
```

I *think* that the code that implements this is in `src/HLL/sprintf.nqp`
I am *completely* unfamiliar with this, but at a guess, it might be `stringify-to-precision2` which is the problem here, as I can see that it uses `nqp::pow_n(10.0, nqp::sub_n($precision, nqp::add_n($exp, 1.0))` and it already looks like `nqp::pow_n` is generating different results on this platform.

I'm not comfortable with that code, but I'm also not competent to replace it. I still think that we should be implementing sprintf as NQP, **but** I'm now far more aware of all the pitfalls of implementing floating point conversions in **any** language. Likely that code should **not** be using `nqp::pow_n` to calculate integer multiples of 10, but I'm not sure if it's good **enough** to just make that change, or if the whole approach should be done differently. (eg convert the value to a decimal string at maximal precision, and then round the decimal digits in the string. I'm not familiar with how competent C sprintf implementations do this, but taking their approach and transcribing C to NQP might be the way to go)

patrickbkr:
Might this be a case where a colaboration of @nwc10 (guts expert) and @lizmat (our expert in sprintf and nqp) will prove fruitful to solve this or provide a workaround?

patrickbkr:
By following nwc10s advice I fixed the first failure (`%d on negative bigint`).

The second error I'm in the process of debugging.

`sprintf` is called with "3.1415e-20".

`stringify-to-precision2` is not actually involved in the failing test (verified with a `say`). It's only using `stringify-to-precision`.

Some prints show, that `stringify-to-precision` is called with the following float: "3.1414999999999935". I think it is legitimate for `stringify-to-precision` to turn that into "3.141e-20".

The actual function where that number is calculated is `scientific`. 
[src/HLL/sprintf.nqp:L369](https://github.com/Raku/nqp/blob/master/src/HLL/sprintf.nqp#L369)

    $float := nqp::div_n($float, nqp::pow_n(10.0, $exp));

turns `3.1415e-20` into  `3.1414999999999935`. But that might also  be legitimate.

So I'm unsure whether the observed behavior is even wrong.

@nwc10 Help!

nwc10:
So, what we figured out yesterday was that `nqp::pow_n(10, 20)` returns `1.0000000000000022e-20` on MinGW - https://colabti.org/irclogger/irclogger_log/moarvm?date=2021-04-12#l168

I'm not sure if IEEE actually makes any guarantees about how wrong `pow` is allowed to be, but I think that this result is off in the 5 least significant bits, which is a lot less than awesome.

The bad news is that I think that this means that the only viable solution is to rewrite the relevant code in `sprintf.nqp` to avoid using the C library pow. There doesn't seem to be a quick workaround.

The good news is that as MoarVM's `coerce_ns` (the OP beneath the `~` operator) effectively already returns a number formatted as `%17g` (and the JVM as `%17G` - sigh) this might be easier than it first sounds - mostly **string** parsing and limted decimal maths using strings.

It won't be before the next release.

lizmat:
Are you aware that there is a Raku based `sprintf` implementation in the RakuAST branch?

nwc10:
I am now. Thanks for the timely "heads up".

My plans so far are about 14 lines written on a single A4 sheet, so it's no loss if that turns out be more useful. Investigations already show that (current) Rakudo builds on nqp's sprintf, **but** nqp itself doesn't use sprintf much, so if Rakudo stops needing nqp's, then a "downsizing" might be possible. (The best code is no code.)

lizmat:
Yup, that would be my plan.

nwc10:
@patrickbkr - does #1472 build on MinGW?

It fixes **this** problem, which is getting in the way of fixing the problem on this ticket:

    nqp -e 'nqp::say(1e23)'
    9.9999999999999992e+22
    nqp -e 'nqp::say(1e23 - 9.9999999999999992e+22)
    0

patrickbkr:
@nwc10 The branch does build and run on MinGW. The three test failures are unchanged.

nwc10:
> The three test failures are unchanged.

Thanks for testing. I didn't expect it to fix the existing MinGW bugs, but it's very good to know that it didn't introduce new regressions, and does fix `1e23` everywhere.

PR #1386, opened 2020-11-20
P6opaque's `spesh` must handle P6bigint values inline, as some are too big

nwc10:
I think that there might actually be two PRs in one here.

### The first is what this branch actually fixes.

With `MVM_SPESH_NODELAY=1` t/spec/S02-types/declare.rakudo.moar and t/spec/S02-types/int-uint.rakudo.moar fail with an exception `Unhandled exception: Cannot unbox 64 bit wide bigint into native integer`

Minimum test case I can golf it down to is

```
use v6;

use Test;
use lib $?FILE.IO.parent.add: 'rakudo/t/spec/packages/Test-Helpers';
use Test::Util;

# L<S09/Sized types/Sized low-level types are named most generally by appending the number of bits to a generic low-level type name>

my $type = 'int64';

my ($minval,$maxval) = ::($type).Range.int-bounds;

try { EVAL "my $type \$var = {$minval-1}" };

use nqp;
nqp::say("Here");

EVAL("my $type \$var = 0; \$var++; \$var");
EVAL("my $type \$var = 1; \$var--; \$var");
```

(yes, really, all the modules are needed too except the nqp)

Under gdb that looks like this:

```
(gdb) r
Starting program: /home/nick/Sandpit/moar-g/bin/moar --libpath=/home/nick/Perl/rakudo --libpath=/home/nick/Perl/rakudo/blib --libpath=/home/nick/Sandpit/moar-g/share/nqp/lib /home/nick/Perl/rakudo/rakudo.moarvm -Ilib ../boom.pl
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff6e69700 (LWP 1068854)]
Here
Unhandled exception: Cannot unbox 64 bit wide bigint into native integer
[Thread 0x7ffff6e69700 (LWP 1068854) exited]
[Inferior 1 (process 1068850) exited with code 01]
(gdb) b MVM_exception_throw_adhoc
Breakpoint 1 at 0x7ffff7817c7e: file src/core/exceptions.c, line 878.
(gdb) r
Starting program: /home/nick/Sandpit/moar-g/bin/moar --libpath=/home/nick/Perl/rakudo --libpath=/home/nick/Perl/rakudo/blib --libpath=/home/nick/Sandpit/moar-g/share/nqp/lib /home/nick/Perl/rakudo/rakudo.moarvm -Ilib ../boom.pl
[...]

PR #1390, opened 2020-11-24
Exceptions in spesh should oops

nwc10:
Calls to `MVM_exception_throw_adhoc` and `MVM_oops` now report if they are in the spesh worker thread or the event loop thread, and exceptions thrown in these threads are treated as oopses, and exit the process immediately.

As suggested in https://github.com/MoarVM/MoarVM/pull/1386#issuecomment-731832866,

Additionally oops with a NULL `tc` is treated reported and then triggers a coredump (instead of causing a SEGV and causing a coredump anyway)

### DIFF
--- a/src/core/exceptions.c
+++ b/src/core/exceptions.c
@@ -648,10 +648,10 @@ static void panic_unhandled_cat(MVMThreadContext *tc, MVMuint32 cat) {
     }
     else {
         if (tc->nested_interpreter)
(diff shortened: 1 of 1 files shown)


jnthn (approved):
The CI error seems to be "##[error]Artifact Test.T.WinMVMrelocatable already exists for build 241.", and that is after it's passed all the tests, so I think it's a CI quirk, not a real issue.

PR #1399, opened 2020-12-07
Fix some simple things the clang static analyzer found

MasterDuke17:
It was easy to run `make clean; scan-build make install` and it produces an html report. There are 112 more things it found, but verifying them is a bit trickier.

### DIFF
--- a/src/instrument/line_coverage.c
+++ b/src/instrument/line_coverage.c
@@ -26,7 +26,6 @@ static void instrument_graph_with_breakpoints(MVMThreadContext *tc, MVMSpeshGrap
             filename_string_index = bbba->filename_string_heap_index;
             MVM_free(bbba);
         } else {
-            line_number = -1;
             bb = bb->linear_next;
             continue;
         }
@@ -147,7 +146,6 @@ static void instrument_graph(MVMThreadContext *tc, MVMSpeshGraph *g) {
             filename_string_index = bbba->filename_string_heap_index;
             MVM_free(bbba);
         } else {
-            line_number = -1;
             bb = bb->linear_next;
             continue;
         }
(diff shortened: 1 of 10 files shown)


src/instrument/line_coverage.c
```diff
@@ -308,12 +306,11 @@ void MVM_line_coverage_report(MVMThreadContext *tc, MVMString *filename, MVMuint
     if (tc->instance->coverage_control == 2 || (!tc->instance->coverage_control && cache[cache_slot] == 0)) {
         char *encoded_filename;
         char composed_line[256];
-        size_t length;
 
         cache[cache_slot] = 1;
 
         encoded_filename = MVM_string_utf8_encode_C_string(tc, filename);
-        if ((length = snprintf(composed_line, 255, "HIT  %s  %d\n", encoded_filename, line_number)) > 0) {
+        if (snprintf(composed_line, 255, "HIT  %s  %d\n", encoded_filename, line_number) > 0) {
```
nwc10, line 313: Whilst this change looks valid, wouldn't be even better to replace `snprintf`-then-`fputs` with a single `fprintf` ?

(I'm only feeling half awake - am I missing something?)

MasterDuke17: Sure, that would make sense, but I'm not sure why we limit the line to 255 characters. Ah, https://github.com/MoarVM/MoarVM/commit/80e5fe5ad9cbb643d1579515e8e66aa462c691ee changes it from fprintf to snprintf+fputs. @timo I don't understand what that commit message means? And why didn't you use `%.255s` with the fprintf instead? Is it not portable?

jnthn: I got curious and went to look at the glibc source and at least there it [seems to acquire a lock](https://github.com/lattera/glibc/blob/master/stdio-common/vfprintf.c#L1318). The [docs for flockfile](https://man7.org/linux/man-pages/man3/flockfile.3.html) indicate that, with few exceptions, the stdio functions applying locking. So in theory a single `printf` would do it.

src/spesh/optimize.c
```diff
@@ -1149,8 +1149,6 @@ static void optimize_string_equality(MVMThreadContext *tc, MVMSpeshGraph *g, MVM
     MVMSpeshFacts *b_facts = MVM_spesh_get_facts(tc, g, ins->operands[2]);
     MVMuint8 was_eq = 0;
 
-    return;
-
```
jnthn, line 1153: This may well have been done to disable an optimization that is broken, so this is a bit risky

MasterDuke17: The `return` was in the original commit (9d8000c) that introduced `optimize_string_equality`, so I'm not sure. But a spectest does pass with the `return` removed.

PR #1401, opened 2020-12-13
Modern raku script extensions

lizmat:
Change the extensions of helper scripts from `.p6` to `.raku` and adjust comments and pod accordingly.


niner:
Missing at least a change in https://github.com/MoarVM/MoarVM/blob/modern-raku-script-extensions/src/core/oplist#L4 as well

niner (approved):

tools/update_ops.raku
```diff
@@ -252,16 +252,16 @@ sub MAIN($file = "src/core/oplist") {
...
 
-    # Generate a p6 Ops file into the tools directory
+    # Generate a Raku Ops file into the tools directory
     my $pf = open("tools/lib/MAST/Ops.pm", :w);
-    $pf.say("# This file is generated from $file by tools/update_ops.p6.");
+    $pf.say("# This file is generated from $file by tools/update_ops.raku.");
     $pf.say("");
-    $pf.say(%op_constants<P6>);
+    $pf.say(%op_constants<Raku>);
     $pf.close;
```
jnthn, line 266: This change (and one below) would seem to imply a change needed in whatever it is downstream that uses the tool - which isn't in this repository.

jnthn: Oh, sign, I misread generating code vs generated code.

PR #1402, opened 2020-12-15
Switch from tommath to gmp

MasterDuke17:
This makes all the build and code changes necessary to use GMP for our big integer library instead of LibTomMath. Uses the ShiftMediaProject fork of GMP because it adds support for building with MSVC (though this PR does not (yet) do anything to configure it for our CI or vice versa).


lizmat:
Could you re-iterate the advantages of this change?  Would be nice copy for the RWN when this is merged :-)

MasterDuke17:
> Could you re-iterate the advantages of this change? Would be nice copy for the RWN when this is merged :-)

Yeah, I'll add some more explanation and benchmarks before this is to be merged. Just getting into a PR now to see what CI has to say.

jnthn:
Does this include the option of using a system gmp as well as building one? (Both matter, the former is probably a requirement for packaging.)

MasterDuke17:
> Does this include the option of using a system gmp as well as building one? (Both matter, the former is probably a requirement for packaging.)

Theoretically yes, but I haven't officially tested it yet (I unofficially did when at first I couldn't get it to use the .a we built). I'll give that a try now.

MasterDuke17:
> Theoretically yes, but I haven't officially tested it yet (I unofficially did when at first I couldn't get it to use the .a we built). I'll give that a try now.

```
$ make realclean
<...>
$ perl Configure.pl --prefix=/home/dan/Source/perl6/install/ --has-gmp --debug=3 --telemeh --dtrace && make -j12 install
<...>
$ find . -name 'libgmp.a'
$ ldd libmoar.so 
        linux-vdso.so.1 (0x00007ffd07f3f000)
        libgmp.so.10 => /usr/lib/libgmp.so.10 (0x00007fdcde3a9000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007fdcde263000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007fdcde241000)
        librt.so.1 => /usr/lib/librt.so.1 (0x00007fdcde236000)
        libdl.so.2 => /usr/lib/libdl.so.2 (0x00007fdcde230000)
        libzstd.so.1 => /usr/lib/libzstd.so.1 (0x00007fdcde15e000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007fdcddf93000)
        /usr/lib64/ld-linux-x86-64.so.2 (0x00007fdcdedbd000)
```
MoarMV builds ok, NQP builds ok and passes `make m-test`, Rakudo builds ok and passes `make m-test`. I'm now getting two tests fail in a `make m-spectest`, but that happens with `--has-gmp` and without (I think it's something with a correction to MVM_bigint_div_num for the rebase to master I recently force pushed).

```
not ok 105 - FatRat + FatRat works (-0.25, 0.00000000000000000000000000000000000000000000000000000000000000000000000000000864)
# Failed test 'FatRat + FatRat works (-0.25, 0.00000000000000000000000000000000000000000000000000000000000000000000000000000864)'
# at t/spec/S32-num/fatrat.t line 124
#     expected approximately: -0.25
#                        got: -0.24999999999999999999999999999999999999999999999999999999999999999999999999999136
# maximum relative tolerance: 1e-06
# actual relative difference: 3
not ok 106 - FatRat + FatRat works (-0.25, 0.00000000000000000000000000000000000000000000000000000000000000000000000000000864)
# Failed test 'FatRat + FatRat works (-0.25, 0.00000000000000000000000000000000000000000000000000000000000000000000000000000864)'
# at t/spec/S32-num/fatrat.t line 125
#     expected approximately: -0.25
#                        got: -0.24999999999999999999999999999999999999999999999999999999999999999999999999999136
# maximum relative tolerance: 1e-06
# actual relative difference: 3
```

MasterDuke17:
> I'm now getting two tests fail in a `make m-spectest`, but that happens with `--has-gmp` and without (I think it's something with a correction to MVM_bigint_div_num for the rebase to master I recently force pushed).

Now fixed with a force push.

2colours:
@MasterDuke17 sorry, what is the current state of this PR? It seems like quite a lot of work and it would really make a difference for performance.

MasterDuke17:
> @MasterDuke17 sorry, what is the current state of this PR? It seems like quite a lot of work and it would really make a difference for performance.

It needs to be rebased obviously, but that actually shouldn't be too bad, it mostly touches code that hasn't been changed all that much recently. Otherwise, it passes all spectests on linux, except for one where gmp succeeds in a calculation that tommath fails. On windows it fails at least one other spectest, something to do with unsigned natives iirc. I haven't figured out what the problem is yet, and getting back to a working config on windows is a daunting task. I believe the only other thing that would need to be figured out is exponentiation, where gmp will `abort()` if the value is too big (tommath just returns an error code), and I'm not sure if it's possible to catch/handle that, we might have to carry a local patch to do something other than `abort()`.

However, I'm wishing I had done the work up front to make this change a build-time option, especially since one of the largest incentives was how much faster gmp's big-integer-to-string function was compared to tommath, but it looks like tommath might be getting a much faster one soon (https://github.com/libtom/libtommath/pull/499).

PR #1408, opened 2020-12-23
Change `procspawnasync` to explicitly take the program name

patrickbkr:
As the value passed as program name and arg0 need to be different in some
cases (e.g. a space in the program name on Windows) we need to retrieve
both separately. Rakudo can then dictate what each value should be.
Part of a fix for rakudo/rakudo#4117

Needs to be merged together with a respective Rakudo PR.

### DIFF
--- a/src/io/procops.c
+++ b/src/io/procops.c
@@ -988,17 +988,22 @@ MVMObject * MVM_proc_spawn_async(MVMThreadContext *tc, MVMObject *queue, MVMObje
         MVM_exception_throw_adhoc(tc,
             "spawnprocasync target queue must have ConcBlockingQueue REPR");
 
-    /* Encode arguments, taking first as program name. */
+    /* Encode program name and arguments. */
     arg_size = MVM_repr_elems(tc, argv);
     if (arg_size < 1)
         MVM_exception_throw_adhoc(tc, "spawnprocasync must have first arg for program");
-    args = MVM_malloc((arg_size + 1) * sizeof(char *));
-    for (i = 0; i < arg_size; i++) {
+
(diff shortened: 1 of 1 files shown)


src/io/procops.c
```diff
@@ -988,17 +988,20 @@ MVMObject * MVM_proc_spawn_async(MVMThreadContext *tc, MVMObject *queue, MVMObje
...
     arg_size = MVM_repr_elems(tc, argv);
     if (arg_size < 1)
         MVM_exception_throw_adhoc(tc, "spawnprocasync must have first arg for program");
-    args = MVM_malloc((arg_size + 1) * sizeof(char *));
-    for (i = 0; i < arg_size; i++) {
+
+    REPR(argv)->pos_funcs.at_pos(tc, STABLE(argv), argv, OBJECT_BODY(argv), 0, &reg, MVM_reg_obj);
+    prog = MVM_string_utf8_c8_encode_C_string(tc, MVM_repr_get_str(tc, reg.o));
+
+    args = MVM_malloc((arg_size - 1 + 1) * sizeof(char *));
```
niner, line 999: Why `arg_size - 1 + 1`? That will end up as `arg_size` anyway.

niner (approved):

src/io/procops.c
patrickbkr: I tried to be as explicit as possible: arg_size - program name + null marker. The compiler will most certainly optimize it away. And it already managed to trip you up and make you aware that something out of the ordinary is going on.
I don't have strong emotions though. So should you object, I can change that.

niner:
> I tried to be as explicit as possible: arg_size - program name + null
> marker.

Sounds like a good place for a comment explaining this :)

PR #1412, opened 2020-12-28
Get MoarVM building on Solaris

nwc10:
Currently only tested as 32 bit with gcc and gmake on Solaris 11.

### DIFF
--- a/src/jit/graph.c
+++ b/src/jit/graph.c
@@ -336,7 +336,9 @@ static void * op_to_func(MVMThreadContext *tc, MVMint16 opcode) {
     case MVM_OP_pow_I: return MVM_bigint_pow;
     case MVM_OP_rand_I: return MVM_bigint_rand;
     case MVM_OP_abs_n: return fabs;
+#ifndef MVM_HAS_SUBSTANDARD_POW
     case MVM_OP_pow_n: return pow;
+#endif
     case MVM_OP_time_n: return MVM_proc_time_n;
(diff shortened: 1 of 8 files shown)


src/jit/graph.c
```diff
@@ -336,7 +336,9 @@ static void * op_to_func(MVMThreadContext *tc, MVMint16 opcode) {
     case MVM_OP_pow_I: return MVM_bigint_pow;
     case MVM_OP_rand_I: return MVM_bigint_rand;
     case MVM_OP_abs_n: return fabs;
+#ifndef MVM_HAS_SUBSTANDARD_POW
     case MVM_OP_pow_n: return pow;
+#endif
```
jnthn, line 341: I think this addition is strictly not needed, as we'll never call `op_to_func` on this op given the change below.

jnthn (approved):
Looks reasonable to me

src/jit/graph.c
nwc10: I think so too.

The backstory is that I found this location in the code first, `#ifdef`ed it out, and only then spotted the second (thinking "woah..."). Addressing the second, I realised that it called the code with the first `case` statement, so wondered about undoing this change you highlighted, but I wasn't sure, so left it at as it.

1. I'm not sure what would be clearer - what I did, or unwinding this
2. I'm not sure if there's any way to add C **compile** time sanity checks to flag up that the oops at the end of `op_to_func` could get reached

PR #1413, opened 2020-12-29
Refactor the C compiler probe code, and add a probe for log() for negative values.

nwc10:
Solaris `log()` (and `log10()`) aren't POSIX conformant for the same circumstances as `pow()`


Configure.pl
```diff
@@ -555,16 +555,15 @@ sub uniq {
                         . "\t\$(CP) 3rdparty/msinttypes/*.h \"\$(DESTDIR)\$(PREFIX)/include/msinttypes\"\n";
 }
 
+if ($^O eq 'aix' && $config{ld} =~ /xlc/ && $config{ptr_size} == 4) {
```
niner, line 562: It really won't matter but from a very quick research, $config{ld} can be a path and the linker command can be like xlc, xlc_v6, xlc128 or the like, so I'd feel more comfortable with a tightened regex like `m!/xlc`

niner (approved):
Removed code is debugged code!

Configure.pl
nwc10: I'm reworking this branch rebased onto #1417 and given that AIX needs fixing "properly", I'm going to remove this regex from this (rebased) commit to keep things simple.

"properly" means that we need to handle different toolchains on the same OS, which is what #1417 starts to enable.

nwc10:
Last 2 commits are new:

- Add MVM_num_isnegzero(tc, n), which tests whether n is -0.0.
- More pow() corner cases that we spectest for, but Solaris gets wrong. 

The others are functionally identical to the previous versions of this branch, but some have been rearranged now that `Configure.pl` no longer calls `foo_native` or `foo_cross`.

PR #1414, opened 2020-12-31
On Solaris, get the ru_maxrss rusage with uv_resident_set_memory().

nwc10:
Solaris getrusage() doesn't actually set ru_maxrss*
and despite providign an API that can get this value, UV doesn't attempt to
"fake it" in uv_getrusage(). So we have to.

* despite first documenting that the value is in units of page size, and
  explaining how to determine the page size.


niner (commented):
I'm not sure about this one. The code itself is OK and I love the documentation. I see a bit of danger though by giving an answer to a slightly different question than asked. The answer will often fit both questions as a lot of programs will only grow over time, but not shrink (especially since actually freeing pages is currently only implemented on Linux), but still it may be misleading. OTOH we don't expose uv_resident_set_memory anywhere else...

PR #1416, opened 2021-01-01
Switch spawnprocasync to use a separate arg for the program name

patrickbkr:
That's a lot more obvious than putting the program name at the start of the
args list.

Prerequisite for rakudo/rakudo#4154


niner (approved):
This change by itself looks good. It is however clearly not backwards compatible. Will this cause issues with NQP/Rakudo build? Or won't those spawn processes anyway?

Btw. I fixed the memory leak of the prog name in https://github.com/MoarVM/MoarVM/commit/ca15d951fb19a8e3130f6e033bbaf898276a9906

patrickbkr:
@niner Thanks for the memory leak fix and review!

Yes the change is not backwards compatible. When only merging this PR and not the other two PRs in the set (see the link in the PR description), then neither Rakudo nor NQP will build. So the three PRs will need to be merged together and the MOAR_REVISION and NQP_REVISION files bumped respectively. (That's what I did in the past in such cases.) Is there anything else I need to look out for when merging?

PR #1417, opened 2021-01-03
Cleanup the build system probes

nwc10:
Currently MoarVM doesn't build on AIX, and doesn't build well on Solaris. Fixing this requires changing some of the structure of how the build system does probes. Hence...

As a first step, these changes attempt to tidy up how we split the overrides between *OS*, *toolchain* and *compiler* such that any override is consistently used at exactly one level. By locally editing the code to disable some sanity checks I can "run" `Configure.pl` for any OS and confirm whether the generated files have any (unintended) changes. I believe on this branch

1. `./Configure` without parameters will generate the same Makefile *etc* as on master
2. `./Configure` with sensible `--toolchain` or `--compiler` options will still work (ie as we permit all combinations, nonsense like "MSVC on AIX" still won't work, but will likely break with a different error message)

CI will check (default) macOS and WIn32. I can check most *nix. Nothing seems to check Mingw32, but I don't think that that is at risk of being broken by these changes.

I see that @manchicken fixed the rakubrew build recently. I don't want to break this. Am I right in thinking that the configurations that need to work for it are

1. `./Configure.pl`
2. `./Configure.pl --toolchain=gnu`

(with --os=darwin set automatically)? I can't find this documented anywhere, but I might have missed something.

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -709,8 +709,20 @@ sub setup_native {
     print dots("Configuring native build environment");
     print "\n";
 
-    $os = build::probe::win32_compiler_toolchain(\%config, \%defaults)
-        if $os eq 'MSWin32';
+    if ($os eq 'MSWin32') {
+        my $has_nmake = 0 == system('nmake /? >NUL 2>&1');
+        my $has_cl    = `cl 2>&1` =~ /Microsoft Corporation/;
+        my $has_gmake = 0 == system('gmake --version >NUL 2>&1');
+        my $has_gcc   = 0 == system('gcc --version >NUL 2>&1');
+        if ($has_nmake && $has_cl) {
+            $os = 'win32';
+        }
+        elsif ($has_gmake && $has_gcc) {
+            $os = 'mingw32';
+        } else {
+            $os = "";
+        }
+    }
 
     if (!exists $::SYSTEMS{$os}) {
         softfail("unknown OS '$os'");
--- a/Configure.pl
+++ b/Configure.pl
@@ -487,21 +487,16 @@ print "OK\n\n";
 
 if ($config{crossconf}) {
     build::auto::detect_cross(\%config, \%defaults);
-    build::probe::static_inline_cross(\%config, \%defaults);
-    build::probe::thread_local_cross(\%config, \%defaults);
-    build::probe::substandard_pow_cross(\%config, \%defaults);
-    build::probe::unaligned_access_cross(\%config, \%defaults);
-    build::probe::ptr_size_cross(\%config, \%defaults);
 }
 else {
     build::auto::detect_native(\%config, \%defaults);
-    build::probe::static_inline_native(\%config, \%defaults);
-    build::probe::thread_local_native(\%config, \%defaults);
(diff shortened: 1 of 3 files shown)


nwc10:
@coke / @manchicken / (anyone else with a Mac setup) - would you be able to check that this branch doesn't break rakubrew? I don't think that it will, but I'm wary of "unknown unknowns".

niner (approved):
There's a missing } in "Nothing reads $config->{win32_compiler_toolchain so don't set it." 

Otherwise a nice set of cleanups to clearly very historically grown code.

coke:
rakubrew is problematic on OS X, but I tried a manual build of this branch:

Configure.pl  with just a prefix option completed, but:

```
$ make
compiling src/main.o
In file included from src/main.c:4:
src/moar.h:31:10: fatal error: 'atomic_ops.h' file not found
#include <atomic_ops.h>
         ^~~~~~~~~~~~~~
1 error generated.
make: *** [src/main.o] Error 1
$ git status
On branch probe-toolchain-cleanup
Your branch is up to date with 'origin/probe-toolchain-cleanup'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)
        modified:   3rdparty/dyncall (modified content)
        modified:   3rdparty/libatomicops (modified content)

no changes added to commit (use "git add" and/or "git commit -a")
```

nwc10:
Er, erk, thanks.

What isn't clear to me - does rakubrew on MoarVM master with the same prefix build just fine?

in that, this error isn't one I'd expect to see caused by the changes on this branch, if the underlying cause is related to #1425.

coke:
rakubrew doesn't work as is on OS X - it's an unsigned binary

coke:
Whoops - tried a clean checkout, mostly OK - was able to build moarvm & nqp, install ,run a mostly successful make test.

```
t/nqp/113-run-command.t ................ Failed 3/8 subtests
t/nqp/111-spawnprocasync.t ............. Failed 1/4 subtests
t/nqp/114-pod-panic.t ..................
```
2 failures, one panic.

nwc10:
Rebased onto master, with two commit messages updated (The one @niner spotted, and the other updated to state that we now have MinGW in our CI)

And so now we also get to see how this branch behaves with MinGW CI :\-)

Also i now have access to MacOS sufficient to test more there, and have verified that this does not make the situation described in rakudo/rakudo#4234 worse.

nwc10:
The branch doesn't (now?) hang for me on the Mac I have access to, or on the CI, so I have no idea what might have been going wrong.

MasterDuke17:
> The branch doesn't (now?) hang for me on the Mac I have access to, or on the CI, so I have no idea what might have been going wrong.

It was one of the unrelated-to-the-PR random fails because of an error renaming a file during precomp on Windows, so I just re-ran the failed job.

nwc10:
Spotted another goof in a commit message.

So fixed, and rebased to master.

PR #1419, opened 2021-01-10
Fix a bunch of memory leaks

niner:
This PR contains fixes to assortment of memory leaks uncovered by ASAN. This covers all individual leaks that occur during a make test or make spectest but not those that need architectural changes like the whole NativeCall memory management situation, cleanup of background threads or cleanup when exceptions occur deep in the call stack.

The commits starting with "Fix ..." fix leaks that occur during a normal run of programs, i.e. these leaks make programs grow over time. The additional commits improve cleanup for runs with --full-cleanup, i.e. reduce "false" positives reported by ASAN. Proper cleanup may also help for embedding situations, so those are not purely theoretical either.

Note that this branch by itself is not enough to be able to simply run make test or make spectest and find new memory leaks. For that one needs additional patches to the build system to enable --full-cleanup everywhere and patches for proper thread cleanup. The latter are not ready yet for review and may need complete re-thinking before committing to them. If desired, I can make available these local patches. After all they have been good enough to facilitate developing these 19 memory leak fixes.

### DIFF
--- a/src/6model/6model.c
+++ b/src/6model/6model.c
@@ -332,6 +332,10 @@ static void mark_sr_data(MVMThreadContext *tc, MVMFrame *frame, MVMGCWorklist *w
     MVM_gc_worklist_add(tc, worklist, &atd->type);
 }
 
+static void free_sr_data(MVMThreadContext *tc, void *sr_data) {
+    MVM_free(sr_data);
+}
+
 void MVM_6model_istype(MVMThreadContext *tc, MVMObject *obj, MVMObject *type, MVMRegister *res) {
     MVMObject **cache;
     MVMSTable  *st;
@@ -390,7 +394,7 @@ void MVM_6model_istype(MVMThreadContext *tc, MVMObject *obj, MVMObject *type, MV
                 atd->obj = obj;
                 atd->type = type;
                 atd->res = res;
-                MVM_frame_special_return(tc, tc->cur_frame, accepts_type_sr, NULL,
+                MVM_frame_special_return(tc, tc->cur_frame, accepts_type_sr, free_sr_data,
                     atd, mark_sr_data);
             }
             STABLE(code)->invoke(tc, code, typecheck_callsite, tc->cur_frame->args);
--- a/src/6model/6model.c
+++ b/src/6model/6model.c
@@ -61,6 +61,9 @@ static void die_over_missing_method(MVMThreadContext *tc, MVMObject *obj, MVMStr
             c_name, MVM_6model_get_debug_name(tc, obj));
     }
 }
+static void find_method_unwind(MVMThreadContext *tc, void *sr_data) {
+    MVM_free(sr_data);
+}
 static void late_bound_find_method_return(MVMThreadContext *tc, void *sr_data) {
     FindMethodSRData *fm = (FindMethodSRData *)sr_data;
     if (MVM_is_null(tc, fm->res->o) || !IS_CONCRETE(fm->res->o)) {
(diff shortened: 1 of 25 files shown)


src/6model/6model.c
```diff
@@ -332,6 +332,10 @@ static void mark_sr_data(MVMThreadContext *tc, MVMFrame *frame, MVMGCWorklist *w
     MVM_gc_worklist_add(tc, worklist, &atd->type);
 }
 
+static void free_sr_data(MVMThreadContext *tc, void *sr_data) {
+    MVM_free(free_sr_data);
```
MasterDuke17, line 336: gcc version 10.2.0 warns here:
```
In file included from src/moar.h:126,
                 from src/6model/6model.c:1:
In function ‘MVM_free’,
    inlined from ‘free_sr_data’ at src/6model/6model.c:336:5:
src/core/alloc.h:43:5: warning: attempt to free a non-heap object [-Wfree-nonheap-object]
   43 |     free(p);
      |     ^~~~~~~
```

niner:
> +    MVM_free(free_sr_data);
> src/core/alloc.h:43:5: warning: attempt to free a non-heap object

And the compiler is completely right! That sould read `MVM_free(sr_data)`. 
Otherwise I'd try to free the function itself.

niner:
pushed a fixed commit

src/profiler/heapsnapshot.c
```diff
@@ -890,8 +890,8 @@ void serialize_attribute_stream(MVMThreadContext *tc, MVMHeapSnapshotCollection
     }
 
     {
-        char namebuf[8] = {0};
-        memcpy(namebuf, name, 8);
+        char namebuf[8];
+        strncpy(namebuf, name, 8);
```
MasterDuke17, line 910: My bad on the buffer overflow.
clang doesn't warn here, but gcc still does:
```
src/profiler/heapsnapshot.c: In function ‘serialize_attribute_stream’:
src/profiler/heapsnapshot.c:910:9: warning: ‘strncpy’ specified bound 8 equals destination size [-Wstringop-truncation]
  910 |         strncpy(namebuf, name, 8);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~
```
The warning can be silenced in both gcc and clang with:
```
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunknown-pragmas"
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wunknown-warning-option"
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
         strncpy(namebuf, name, 8);
+#pragma GCC diagnostic pop
+#pragma clang diagnostic pop
+#pragma GCC diagnostic pop
```

src/io/procops.c
```diff
@@ -944,6 +949,9 @@ static void spawn_gc_mark(MVMThreadContext *tc, void *data, MVMGCWorklist *workl
 static void spawn_gc_free(MVMThreadContext *tc, MVMObject *t, void *data) {
     if (data) {
         SpawnInfo *si = (SpawnInfo *)data;
+        if (si->prog) {
+            MVM_free_null(si->prog);
```
MasterDuke17, line 953: For curiosity's sake, why the `MVM_free_null` here vs just `MVM_free`? I thought since `si` itself is unconditionally freed at the end, setting the fields to null is not needed?

src/strings/nfg.c
```diff
@@ -471,12 +471,25 @@ void MVM_nfg_init(MVMThreadContext *tc) {
     cache_crlf(tc);
 }
 
+static void nfg_trie_node_destroy(MVMThreadContext *tc, MVMNFGTrieNode *node) {
+    MVMint32 i = 0;
```
MasterDuke17, line 475: Nit-pick, but in some of the other commits you put the declaration right in the first entry of the for.

MasterDuke17 (approved):

PR #1420, opened 2021-01-19
Introduce platform/socket.h

Kaiepi:
There are a couple spots in MoarVM that carry similar platform-dependent socket behaviour, and more are to come. This can be unified.


Kaiepi:
Can someone review this? This is needed so `struct sockaddr` and friends can be available for a REPR I introduce for https://github.com/Raku/problem-solving/issues/111.

lizmat:
Looks good to me, but I should really disqualify myself in these matters.  Pinged jnthn and niner

PR #1428, opened 2021-02-03
Fix compile warnings

MasterDuke17:
You only see the warnings the first commit fixes if FSA_DEBUG is enabled, but the second commit fixes ones always seen.

### DIFF
--- a/src/core/fixedsizealloc.c
+++ b/src/core/fixedsizealloc.c
@@ -13,7 +13,7 @@
 #define FSA_SIZE_DEBUG 0
 #if FSA_SIZE_DEBUG
 typedef struct {
-    MVMuint64 alloc_size;
+    size_t alloc_size;
     void *memory;
 } MVMFixedSizeAllocDebug;
 #endif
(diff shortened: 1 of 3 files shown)


src/core/fixedsizealloc.c
```diff
@@ -294,7 +294,7 @@ void MVM_fixed_size_free(MVMThreadContext *tc, MVMFixedSizeAlloc *al, size_t byt
 #ifdef MVM_VALGRIND_SUPPORT
         if (RUNNING_ON_VALGRIND) {
             char command[128];
-            snprintf(&command, 128, "check_memory defined %p %d", dbg, bytes + 8);
+            snprintf(command, 128, "check_memory defined %p %lu", dbg, bytes + 8);
```
niner, line 297: Shouldn't that be %zu since bytes is a size_t?

src/core/fixedsizealloc.c
```diff
@@ -341,7 +341,7 @@ void MVM_fixed_size_free_at_safepoint(MVMThreadContext *tc, MVMFixedSizeAlloc *a
     if (dbg->alloc_size != bytes) {
 #ifdef MVM_VALGRIND_SUPPORT
         if (RUNNING_ON_VALGRIND) {
-            char command[128]; snprintf(&command, 128, "check_memory defined %p %lu", dbg, bytes + 8); VALGRIND_MONITOR_COMMAND(command);
+            char command[128]; snprintf(command, 128, "check_memory defined %p %lu", dbg, bytes + 8); VALGRIND_MONITOR_COMMAND(command);
```
niner, line 344: same here

src/core/fixedsizealloc.c
MasterDuke17: Guess so, `%ld` is what gcc suggested, but I made it `%lu` to be consistent with the format below. I can change both.

MasterDuke17: Did that, but also changed `MVMFixedSizeAllocDebug->alloc_size` to be `size_t`.

niner (approved):

PR #1436, opened 2021-02-20
Fix possible GC upset caused by half-deserialized STables

niner:
While deserializing STables, we may have to deserialize several objects. Doing
so can trigger a garbage collection run. At this point the STable will not be
fully set up yet. E.g. its type_check_cache_length was set but not actually
filled with appropriate objects. This could lead to the GC following invalid
pointers.

Fix by being more careful about setting up data structures hanging off STables
first and only attaching them to the STable once they are fully ready.

### DIFF
--- a/src/6model/serialization.c
+++ b/src/6model/serialization.c
@@ -2714,11 +2714,17 @@ static void deserialize_stable(MVMThreadContext *tc, MVMSerializationReader *rea
     deserialize_method_cache_lazy(tc, st, reader);
 
     /* Type check cache. */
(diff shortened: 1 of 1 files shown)


niner:
I'm not 100 % sure about invocation_spec. What happens, when some of the objects are already deserialized and the pointers written to invocation_spec and a GC gets triggered? Since the invocation_spec itself is not yet attached to the STable, it won't get processed and may then contain outdated pointers. Except that we deserialize into gen2 by default anyway. So deserialized objects won't move. But then I wonder why we need to use MVM_ASSIGN_REF if both the STable and the assigned objects are actually guaranteed to be in gen2.

PR #1444, opened 2021-03-10
Constrain C pointer serialization to sizes <= INT32_MAX

Kaiepi:
The size of C pointers are passed directly to `assert_can_read` upon deserialization, which takes an `MVMint32`. Constraining serialized C pointer sizes to `INT32_MAX` both prevents overflows and eliminates a couple compiler warnings produced by a comparison of an `MVMint64` to `SIZE_MAX` on some systems, which is unsigned.

Despite this, in theory, C pointer serialization is still flawed; the size of a pointer can overflow the read offset depending on the size of pointer serialized and where it resides in a script. Something more akin to how Raku strings are serialized would be necessary in order to avoid this problem, but we've made it this far without encountering it, so deal with that at another point in time.

CC @MasterDuke17


lizmat:
So, is this ok to merge?  @jnthn @niner

niner (approved):

PR #1445, opened 2021-03-20
Add a setup_notify handler and queue to nqp::signal

niner:
Setting up a signal handler happens on the IO eventloop thread. Thus when
nqp::signal is done, one must not assume that the handler is already in place.
Add setup_notify_queue and setup_notify_schedulee arguments, so we can notify
the caller when the handler is setup to avoid race conditions.


lizmat:
Now that we have a 2021.03 release, is this ok to merge?

PR #1450, opened 2021-03-25
Remove nqp::time_(i|n) in favor of nqp::time...

MasterDuke17:
which just return integer nanoseconds since the epoch. Before we were just throwing away the extra precision possible from using nanoseconds.


lizmat:
Why remove the old nqp::ops in the same commit?   This would force people into a flag day with regards to their use of `nqp::time_n` and `nqp::time_i`.

I would prefer to see the removal of the old `nqp::time_n|i` ops to be some time in the future, possibly years from now.

jnthn:
> This would force people into a flag day with regards to their use of nqp::time_n and nqp::time_i.

It's not a supported API, along with all the other `nqp::op`s. Those using it can already migrate to `now` or similar today.

MasterDuke17:
I don't have strong feelings either way. Sure `nqp::` ops are documented in the NQP repo, but not in docs.raku.org. Also, there are only two uses of any `nqp::` ops in the entire roast. If we did decide to make the change for the next release I would create PRs for the ecosystem (greppable is a bit out of date, but it doesn't show too many https://gist.github.com/Whateverable/6738789120beeaedfe9d5b4873cc4d04).

jnthn:
> It is documented: https://github.com/Raku/nqp/blob/master/docs/ops.markdown#time

That is internal developer documentation, *not* Raku language user documentation. Please don't get me an incentive to *actively* avoid documenting internals; I'm bad enough at doing so already. :-)

> I just don't see why we would drop this unannounced

Because it's an internals change, and setting precedents matters. Use of the `nqp::op`s is even behind a pragma to make it impossible to do by accident. Given the magnitude of changes in the pipeline (`new-disp` could drop easily a dozen ops, likely including `nqp::findmethod`), we're in for a world of pain if we decide we want to avoid a flag day on these as policy. 

In this particular case, if we *do* want a comparability shim, it can be done in NQP (e.g. `nqp::time_n` would get mapped to something in terms of the new `nqp:;time`), which is far cheaper than retaining the ops in MoarVM for longer. I don't mind if somebody does that, but so far as the messaging around use of `nqp::op`s in ecosystem modules goes, I'm not sure it's what serves us best in the long run.

lizmat:
Ok, @jnthn's arguments are clear.

> In this particular case, if we do want a comparability shim, it can be done in NQP (e.g. nqp::time_n would get mapped to something in terms of the new nqp:;time), which is far cheaper than retaining the ops in MoarVM for longer. I don't mind if somebody does that, but so far as the messaging around use of nqp::ops in ecosystem modules goes, I'm not sure it's what serves us best in the long run.

I would agree generally, but in the case of `nqp::time_n`, people may have used that in time critical application *exactly* because of the overhead we're now removing.  If we remove them now, their code won't even compile anymore.

BTW, wouldn't a shim in Rakudo not even be better?  Like some of the nqp::p6 ops ?

MasterDuke17:
> overhead we're now removing

FYI, the PRs as they are currently don't remove significant overhead. All the logic is the same, essentially all I've done is moved the conversion from integer nanoseconds to floating point seconds up to NQP/Rakudo. What's needed is to look where that's happening and see if there's a more efficient way (e.g., maybe some places could now store/operate on integer nanoseconds and only convert to fractional seconds when displayed).

jnthn:
> I would agree generally, but in the case of nqp::time_n, people may have used that in time critical application exactly because of the overhead we're now removing. If we remove them now, their code won't even compile anymore.

In this case there's a way to do it without creating much of a maintenance burden, and you're right that the pragmatic thing to do here is to shim it. I guess my fear is that the "generally" will never happen, and for every such change somebody will make an "in this case" argument. It's already a huge amount of work for those doing significant changes to maintain things up to bug compatibility in the things that aren't considered internals. Given I'm on the hook for two such changes (`new-disp`, `rakuast`), I've got a rather large interest in the precedent being "if you depended on `nqp::op`s or compiler internals, you get to keep all the pieces".

> BTW, wouldn't a shim in Rakudo not even be better? Like some of the nqp::p6 ops ?

Hm, sort of. Most of the `nqp::p6foo` are custom compilations, but there is also a desugar mechanism. Those are registered in the actions, and are the best solution for this, because those work over all backends.

MasterDuke17:
> Hm, sort of. Most of the `nqp::p6foo` are custom compilations, but there is also a desugar mechanism. Those are registered in the actions, and are the best solution for this, because those work over all backends.

Something like?:
```raku
register_op_desugar('time_i', -> {
    QAST::Op.new( :op('div_i'), QAST::Op.new( :op('time' ) ), 1000000000 )
});
register_op_desugar('time_n', -> {
    QAST::Op.new( :op('div_n'), QAST::Op.new( :op('time' ) ), 1000000000e0 )
});
```

jnthn:
> What's needed is to look where that's happening and see if there's a more efficient way (e.g., maybe some places could now store/operate on integer nanoseconds and only convert to fractional seconds when displayed).

Glancing this:

```
my class Instant is Cool does Real {
    has Rat $.tai;
```

I guess it'd be possible to change that to an `int` attribute, and avoid the `Rat` (which is 3 further GC allocations) unless it is requested. Probably even the `Str` method can stick in `num` land.

jnthn:
> QAST::Op.new( :op('div_i'), QAST::Op.new( :op('time' ) ), 1000000000 )

Needs to be `QAST::IVal.new( :value(1000000000) )` (and an `NVal` for the other case). Otherwise looks about right.

MasterDuke17:
> I guess it'd be possible to change that to an `int` attribute, and avoid the `Rat` (which is 3 further GC allocations) unless it is requested. Probably even the `Str` method can stick in `num` land.

Yep, I started to try something like that right away, but then didn't get it quite right and decided to do the straight port first, then follow up with optimizations.

PR #1451, opened 2021-03-28
Fixes for NULL bugs / spesh bugs / wrong backtraces

niner:
MasterDuke's try to log undefined results of decont ops uncovered a deep rabbit hole of NULL bugs, spesh bugs and wrong backtraces. This PR takes care of them, unblocking his work.

### DIFF
--- a/src/core/exceptions.c
+++ b/src/core/exceptions.c
@@ -523,6 +523,10 @@ MVMObject * MVM_exception_backtrace(MVMThreadContext *tc, MVMObject *ex_obj) {
         cur_frame = ((MVMException *)ex_obj)->body.origin;
         throw_address = ((MVMException *)ex_obj)->body.throw_address;
     }
+    else if (ex_obj == tc->instance->VMNull) {
+        cur_frame = tc->cur_frame;
+        throw_address = *tc->interp_cur_op;
+    }
     else {
         MVM_exception_throw_adhoc(tc, "Op 'backtrace' needs an exception object");
     }
--- a/src/core/exceptions.c
+++ b/src/core/exceptions.c
@@ -514,6 +514,7 @@ char * MVM_exception_backtrace_line(MVMThreadContext *tc, MVMFrame *cur_frame,
 /* Returns a list of hashes containing file, line, sub and annotations. */
 MVMObject * MVM_exception_backtrace(MVMThreadContext *tc, MVMObject *ex_obj) {
     MVMFrame *cur_frame;
+    MVMSpeshFrameWalker fw;
     MVMObject *arr = NULL, *annotations = NULL, *row = NULL, *value = NULL;
     MVMuint32 count = 0;
     MVMString *k_file = NULL, *k_line = NULL, *k_sub = NULL, *k_anno = NULL;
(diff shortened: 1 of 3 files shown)


src/core/exceptions.c
```diff
@@ -544,46 +549,51 @@ MVMObject * MVM_exception_backtrace(MVMThreadContext *tc, MVMObject *ex_obj) {
...
+        MVM_spesh_frame_walker_next(tc, &fw); // Go to first inline
+
+        do {
+            cur_frame = MVM_spesh_frame_walker_current_frame(tc, &fw);
+            MVMuint8             *cur_op = count ? cur_frame->return_address : throw_address;
+            MVMuint32             offset = cur_op - MVM_frame_effective_bytecode(cur_frame);
+            MVMBytecodeAnnotation *annot = MVM_bytecode_resolve_annotation(tc, &cur_frame->static_info->body,
+                                                offset > 0 ? offset - 1 : 0);
+            MVMuint32             fshi   = annot ? (MVMint32)annot->filename_string_heap_index : -1;
+            char            *line_number = MVM_malloc(16);
```
MasterDuke17, line 563: Just curious, why not a static char array for `line_number`?

src/core/interp.c
```diff
@@ -5517,12 +5519,16 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
...
                 cur_op += 4;
                 goto NEXT;
-            OP(hllbool):
-                GET_REG(cur_op, 0).o = GET_REG(cur_op, 2).i64
+            OP(hllbool): {
+                MVMObject *bool_value = GET_REG(cur_op, 2).i64
                     ? cu->body.hll_config->true_value
                     : cu->body.hll_config->false_value;
+                if (!bool_value)
+                    MVM_exception_throw_adhoc(tc, "Trying to hllbool a value, but current language doesn't have HLL bools");
```
MasterDuke17, line 5527: Might be nice to add the `cu->body.hll_name` to the error message.

MasterDuke17 (approved):
Not sure I'm qualified to definitively say everything is correct, but LGTM.

nwc10:
The commit 298298aaac65e71 fixes a SEGV in t/spec/S12-attributes/undeclared.t when running with MVM_SPESH_NODELAY=1 MVM_SPESH_BLOCKING=1. However, somehow it **exposes** two more bugs:

`$ ./rakudo-m -Ilib t/spec/S09-multidim/assign.t`

```
ok 36 - Cannot assign to many lists at first dimension (native)
not ok 37 - Cannot assign to many items at first dimension (native)
# Failed test 'Cannot assign to many items at first dimension (native)'
# at t/spec/S09-multidim/assign.t line 79
```

(for that one valgrind says nothing, the failure is unchanged and it is not a JIT bug)

```
$ ./rakudo-m -Ilib t/spec/S22-package-format/local.t
1..19
ok 1 - Could we create '/home/nick/Perl/rakudo/t/spec/S22-package-format/local-file-src'
ok 2 - Could we create '/home/nick/Perl/rakudo/t/spec/S22-package-format/local-file-src/NanooNanoo.pm6'
ok 3 - Could we create /home/nick/Perl/rakudo/t/spec/S22-package-format/local-file-cmp
ok 4 - install Ctrl-C handler for cleanup in END
ok 5 - The object is-a 'CompUnit::Repository::FileSystem'
ok 6 - The object is-a 'IO::Path'
ok 7 - is . looking at the right directory
ok 8 - is the short-id right
ok 9 - Cannot install on CUR::FileSystem
ok 10 - The object is-a 'CompUnit::Repository::FileSystem'
ok 11 - are they the same
ok 12 - The object is-a 'CompUnit::Repository::FileSystem'
ok 13 - are they different
ok 14 - The object is-a 'IO::Path'
ok 15 - is '/home/nick/Perl/rakudo/t/spec/S22-package-format/local-file-src' looking at the right dir
===SORRY!===
Type check failed in binding to parameter '$bytes'; expected Blob but got Supplier::Preserving (Supplier::Preserving...)
# You planned 19 tests, but ran 15
```

Oooh, *this* one is a JIT bug. It goes away with `valgrind`, and with MVM_JIT_EXPR_DISABLE=1

Random guess (based on my only experience of this) is that it's something to do with memory layout - I think that valgrind ends up with malloc returning addresses that have the top 32 bits 0, meaning that they *don't* get corrupted if truncated and expanded again, whereas "regular" runtimes generate heap addresses that do. For my single experience of these symptoms, the problem was caused by trying to write a JIT template that loaded an address with other-than `mov64`. But the fallout was a SEGV when the truncated address didn't work. Which isn't what we see here. But a pointer comparison failing?

niner:
> Might be nice to add the `cu->body.hll_name` to the error message.

I noticed myself but the cost of having to do the whole dance with 
MVM_exception_throw_adhoc_va_free did not seem worth the slight improvement 
for the very improbable case. After all it occured only a single time in all 
of rakudo and can only happen when using nqp ops directly and only in NQP 
code.

niner:
> Just curious, why not a static char array for `line_number`?

I don't know. That part is unchanged from before. Only the indentation 
changed.

PR #1456, opened 2021-04-03
Save malloc+free per frame when creating backtrace

MasterDuke17:
By using a fixed size char array. The number of instructions reported by
callgrind do vary a bit, even with `MVM_SPESH_BLOCKING=1`, but it seems
to save ~1m instructions for `my $a; $a = do given "hi" { when 3 {
    "three" }; when 5 { "five" }; when "hi" { "hello" }; when * > 40 {
        "> 40" }; default { "default" } } for ^1_000; say $a`.


MasterDuke17:
Going to rename and add another commit.

PR #1457, opened 2021-04-04
Some minor optimizations when creating backtraces

MasterDuke17:
By using a fixed size char array in one place and the FSA in another. The number of instructions reported by callgrind do vary a bit, even with `MVM_SPESH_BLOCKING=1`, but it seems to save ~17m instructions for:
```raku
my $a;
$a = do given "hi" {
	when 0      { "zero"    };
	when 1      { "one"     };
	when 2      { "two"     };
	when 3      { "three"   };
	when 4      { "four"    };
	when 5      { "five"    };
	when "hi"   { "hello"   };
	when * > 40 { "> 40"    };
	default     { "default" }
} for ^1_000;
say $a
```

### DIFF
--- a/src/core/exceptions.c
+++ b/src/core/exceptions.c
@@ -560,9 +560,9 @@ MVMObject * MVM_exception_backtrace(MVMThreadContext *tc, MVMObject *ex_obj) {
             MVMBytecodeAnnotation *annot = MVM_bytecode_resolve_annotation(tc, &cur_frame->static_info->body,
                                                 offset > 0 ? offset - 1 : 0);
             MVMuint32             fshi   = annot ? (MVMint32)annot->filename_string_heap_index : -1;
-            char            *line_number = MVM_malloc(16);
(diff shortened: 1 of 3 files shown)


src/core/exceptions.c
```diff
@@ -560,9 +560,9 @@ MVMObject * MVM_exception_backtrace(MVMThreadContext *tc, MVMObject *ex_obj) {
             MVMBytecodeAnnotation *annot = MVM_bytecode_resolve_annotation(tc, &cur_frame->static_info->body,
                                                 offset > 0 ? offset - 1 : 0);
             MVMuint32             fshi   = annot ? (MVMint32)annot->filename_string_heap_index : -1;
-            char            *line_number = MVM_malloc(16);
+            char         line_number[11] = {0};
```
niner, line 563: Would be nice to have a comment on why 11 is the right number. I guess it's because you need a maximum of 10 decimal digits + the NULL character to represent a 32 bit line number. Having that documented makes it clear that it needs adjusting if we ever move to 64 bit line numbers.

niner (approved):

Issue #1434, opened 2021-02-18
Abort when using MVM_SPESH_LOG


MasterDuke17:
```
(gdb) r
Starting program: /home/dan/Source/perl6/install/bin/raku -I lib/ t/09-moar/10-inline-array-pos.t
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[New Thread 0x7ffff6de3640 (LWP 3503209)]
[New Thread 0x7ffff3113640 (LWP 3503223)]
[New Thread 0x7ffff2912640 (LWP 3503224)]
[New Thread 0x7ffff2111640 (LWP 3503225)]
[Detaching after fork from child process 3503226]
[New Thread 0x7ffff1910640 (LWP 3503227)]
1..4
ok 1 - Was AT-POS inlined?
ok 2 - Was ASSIGN-POS inlined?
ok 3 - Was ASSIGN_POS_SLOW_PATH inlined?
ok 4 - Was postcircumfix:<[ ]> inlined?
free(): corrupted unsorted chunks

Thread 2 "spesh optimizer" received signal SIGABRT, Aborted.
[Switching to Thread 0x7ffff6de3640 (LWP 3503209)]
0x00007ffff74bbef5 in raise () from /usr/lib/libc.so.6
(gdb) bt
#0  0x00007ffff74bbef5 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff74a5862 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff74fdf38 in __libc_message () from /usr/lib/libc.so.6
#3  0x00007ffff7505bea in malloc_printerr () from /usr/lib/libc.so.6
#4  0x00007ffff7507502 in _int_free () from /usr/lib/libc.so.6
#5  0x00007ffff750aca8 in free () from /usr/lib/libc.so.6
#6  0x00007ffff7939438 in MVM_free (p=<optimized out>) at src/core/alloc.h:43
#7  0x00007ffff793c7f6 in optimize_call (tc=tc@entry=0x5555555f4bd0, g=g@entry=0x7fffec32eb80, bb=bb@entry=0x7fffec62d590, ins=ins@entry=0x7fffec62c078, p=p@entry=0x7fffec564a20, callee_idx=callee_idx@entry=1, arg_info=0x7ffff6de2720)
    at src/spesh/optimize.c:2079
[...]

PR #1458, opened 2021-04-05
Fix possible segfault on exit when using spesh log

niner:
A normal exit without --full-cleanup did not stop the spesh thread. So spesh
may actually still be active and try to print things to the spesh log when
MVM_vm_exit closes the spesh log file handle. This would lead to a segfault in
vfprintf. Fix by stoping and joining the spesh thread even in MVM_vm_exit if
spesh logging is active.

### DIFF
--- a/src/moar.c
+++ b/src/moar.c
@@ -590,8 +590,13 @@ void MVM_vm_exit(MVMInstance *instance) {
     MVM_io_flush_standard_handles(instance->main_thread);
 
     /* Close any spesh or jit log. */
-    if (instance->spesh_log_fh)
+    if (instance->spesh_log_fh) {
+        /* Need to properly shut down spesh, otherwise we may segfault trying
+         * to write to a closed file handle before we exit completely */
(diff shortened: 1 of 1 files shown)


niner:
Should fix https://github.com/MoarVM/MoarVM/issues/1434

src/moar.c
```diff
@@ -590,8 +590,13 @@ void MVM_vm_exit(MVMInstance *instance) {
     MVM_io_flush_standard_handles(instance->main_thread);
 
     /* Close any spesh or jit log. */
-    if (instance->spesh_log_fh)
+    if (instance->spesh_log_fh) {
+        /* Need to properly shut down spesh, otherwise we it may segfault
```
MasterDuke17, line 594: "we it"

niner:
> "we it"

Thanks, fixed it in the updated commit. Funny, there's actually another place 
with the exact same error in our code base :)

MasterDuke17 (approved):

PR #1460, opened 2021-04-06
Fix when building to a directory which includes a '+' in its file name

### LINKED ISSUE #1459: Perl errors when compiling
I tried to compile MoarVM on Arch Linux, with Perl v5.32.1, everything goes fine until "linking moar":
```
make install
[...]
linking moar
/usr/bin/perl build/mk-moar-pc.pl pkgconfig/moar.pc
String found where operator expected at build/mk-moar-pc.pl line 26, near "my $ldflags = ""
	(Missing semicolon on previous line?)
Ambiguous use of ${package} resolved to $package at build/mk-moar-pc.pl line 26.
String found where operator expected at build/mk-moar-pc.pl line 26, near "my $libs = ""
	(Missing semicolon on previous line?)
String found where operator expected at build/mk-moar-pc.pl line 26, near "my $cflags = ""
	(Missing semicolon on previous line?)
String found where operator expected at build/mk-moar-pc.pl line 26, near "my $static_libs = ""
	(Missing semicolon on previous line?)
Ambiguous use of ${package} resolved to $package at build/mk-moar-pc.pl line 26.
String found where operator expected at build/mk-moar-pc.pl line 26, near "$ldflags = ""
	(Missing semicolon on previous line?)
String found where operator expected at build/mk-moar-pc.pl line 26, near "$ldflags = ""
	(Missing semicolon on previous line?)
Scalar found where operator expected at build/mk-moar-pc.pl line 26, near "${ldopts} ${ldflags}"
	(Missing operator before ${ldflags}?)
String found where operator expected at build/mk-moar-pc.pl line 26, near "$libs = ""
	(Missing semicolon on previous line?)
String found where operator expected at build/mk-moar-pc.pl line 26, near "$cflags = '' if ${includedir} eq ""
	(Missing semicolon on previous line?)
Use of /c modifier is meaningless without /g at build/mk-moar-pc.pl line 26.
String found where operator expected at build/mk-moar-pc.pl line 26, near "or die ""
	(Missing semicolon on previous line?)
Bareword found where operator expected at build/mk-moar-pc.pl line 27, near "or die "Can"
  (Might be a runaway multi-line "" string starting on line 26)
	(Do you need to predeclare or?)
String found where operator expected at build/mk-moar-pc.pl line 27, at end of line
	(Missing semicolon on previous line?)
syntax error at build/mk-moar-pc.pl line 26, near "/."
Global symbol "$ldflags" requires explicit package name (did you forget to declare "my $ldflags"?) at build/mk-moar-pc.pl line 26.
Global symbol "$ldopts" requires explicit package name (did you forget to declare "my $ldopts"?) at build/mk-moar-pc.pl line 26.
Global symbol "$ldflags" requires explicit package name (did you forget to declare "my $ldflags"?) at build/mk-moar-pc.pl line 26.
Global symbol "$ldflags" requires explicit package name (did you forget to declare "my $ldflags"?) at build/mk-moar-pc.pl line 26.
Regexp modifiers "/l" and "/u" are mutually exclusive at build/mk-moar-pc.pl line 26, at end of line
Regexp modifiers "/l" and "/d" are mutually exclusive at build/mk-moar-pc.pl line 26, at end of line
Unknown regexp modifier "/e" at build/mk-moar-pc.pl line 26, at end of line
Can't find string terminator '"' anywhere before EOF at build/mk-moar-pc.pl line 27.
make: *** [Makefile:761: pkgconfig/moar.pc] Error 255
```
The error also happens on Debian bullseye, in a Docker container.
Using the master branch.

demostanis:
Resolves #1459


demostanis:
My code editor defaulted to using tabs instead of space and wasn't showing the difference, should I make a new commit replacing tabs with spaces?

lizmat:
Possibly.  If you do, maybe also fix the other stray TAB that's in there :-)

MasterDuke17:
> My code editor defaulted to using tabs instead of spaces and wasn't showing the difference, should I make a new commit replacing tabs with spaces?

I'd just rebase -i and merge the commits, then force push.

demostanis:
Did I do it right?

lizmat:
Looks ok to me :)

MasterDuke17:
> Did I do it right?

There's no real problem with it, but what I meant was to do a `git rebase -i HEAD~3`, then change the 'pick' for the second (and third, etc) commit (the "remove tabs" one) to 'fixup'. That will combine all the changes back down into one commit. Then you'd need to `git push --force` because you actually changed the commit SHAs. So at the end the PR would have just one commit with all the changes. But it's not required.

demostanis:
Is this better?

patrickbkr:
IMHO this is a bit too specific, as it special-cases a single replacement with a custom separator. Could you switch the quoting to be a default `'...'` quoting and then escape the `'` and `/` instead? Then we could reuse that functionality should other replacements need the same treatment.

demostanis:
Is this better?

patrickbkr (approved):
Looking very good.

demostanis:
Don't merge, I messed up somewhere...

demostanis:
^
src/gen/config.c's backslashes to escape " were being escaped by the lines I added, which caused syntax errors.
It was also escaping backslashes at the end of lines.

demostanis:
What are those errors Azure is getting?

MasterDuke17:
https://dev.azure.com/MoarVM/MoarVM/_build/results?buildId=416&view=logs&j=1f3a7134-3b80-549a-eaff-49041f6b71c8&t=cebb2378-2f87-5e8d-fe8c-4dbf4bfaf11c&l=311 some problem building 3rdparty/libatomicops/src/libatomic_ops.a

demostanis:
Escaping + was much easier... Why is it even failing?

demostanis:
It's best to only escape + sign in the specific line. Escaping ' and \ might introduce errors (e.g `/bin/sh: line 1: -Wextra: command not found`) as some files use " (unlike src/mk-moar-pc.in where the problem first came from).

patrickbkr:
Ah, now I see. You removed the condition and unconditionally escaped replacements. As you discovered that is bound to fail, as several other places also take this code path but need a different or no escaping.

To fix that one would need to review all uses of the function to make sure it's called uniformly. I wouldn't do that now.
But can you instead just re-add the condition so the escaping is only done in this specific case? Then, should the need for the same escaping arise in a different place, one can then just extend the condition. (That's what I originally intended to propose, but obviously didn't make clear enough.)

demostanis:
It compiles fine on my machine, I'm not getting the error Azure gets.
Can anyone else test it?

patrickbkr:
The azure error shows up for relocatable builds, those have `'` characters in the ldflags.

I think I don't understand the regex you wrote. Also, why do you do the replacement for the `prefix` variable instead of the `ldflags` variable? In short, why not just:

    if ($key eq 'ldflags') {
        $config{$key} =~ s/\\/\\\\/g;
        $config{$key} =~ s/\'/\\\'/g;

demostanis:
I did it this way because `@ldflags@` equals to the flags, which include (literally!) `@libdir@`, a config variable comprised of (again, literally) `@prefix@`, a string which might have a quote in it, plus `/bin`. Thus, trying to escape quotes in `@ldflags@` won't escape the ones in `@prefix@`.
As the `configure` subroutine is called multiple times on lines with the same variables, `s///`ing `$config{$key}` might do it many times (~10 for `ldflags`), and escaping `'` ~10 times (which is what your code does).
That's why my regex checks for any previous escaping (`(?<!\\)\\(?![\\\'])` instead of `\\`).
I hope my explanation is understandable.
Maybe the best would be to fix those issues instead of writing complex regexes.

demostanis:
It doesn't

demostanis:
Let's hope this time is the right

patrickbkr:
I now noticed a detail in your solution I missed before. It changes the configuration values themselves. That's a no-go. There is a recommendation to only ever quote / escape at the actual point of use. I think that makes a lot of sense as it makes sure one never runs into the trouble of double quoting.

What do you think about the following?

        my $val = $config{$key};
        # escape \, ' and "
        if ($template =~ /'[^']*@\Q$key\E@[^']*'/) {
            $val =~ s/\\/\\\\/g;
            $val =~ s/'/\\'/g;
        }
        elsif ($template =~ /"[^"]*@\Q$key\E@[^"]*"/) {
            $val =~ s/\\/\\\\/g;
            $val =~ s/"/\\"/g;
        }

        $template =~ s/@\Q$key\E@/$val/;

That's still not completely fool proof, but catches the common failures.

demostanis:
That's for sure a better solution than mine, I have no objections about it.

patrickbkr:
@demostanis Can you make the change?

patrickbkr:
There is some whitespace mess. Can you clean that up?

patrickbkr:
There are still tabs in there :-(

demostanis:
Where do you see them? `awk '/\t/' Configure.pl ` yields nothing.

patrickbkr:
Probably a stale browser cache, looking good now. Thank you for pushing through with this!

PR #1461, opened 2021-04-06
Cleanup of a bunch of Unicode hashes

### LINKED ISSUE #1453: Memory leak of unicode hash table
```
[dan@alexandria p6]$ valgrind --leak-check=full ./install/bin/raku --full-cleanup -e 'my %expressions = D => "got D", M => "got M", l => "got l"; my $a; $a := "%D %M %l".subst(/"%"$<specifier>=<{ %expressions.keys.join("|") }>/, -> ( :$specifier ) { %expressions{$specifier} }, :g) for ^10'
==3058731== Memcheck, a memory error detector
==3058731== Copyright (C) 2002-2017, and GNU GPL'd, by Julian Seward et al.
==3058731== Using Valgrind-3.16.1 and LibVEX; rerun with -h for copyright info
==3058731== Command: ./install/bin/raku --full-cleanup -e my\ %expressions\ =\ D\ =\>\ "got\ D",\ M\ =\>\ "got\ M",\ l\ =\>\ "got\ l";\ my\ $a;\ $a\ :=\ "%D\ %M\ %l".subst(/"%"$\<specifier\>=\<{\ %expressions.keys.join("|")\ }\>/,\ -\>\ (\ :$specifier\ )\ {\ %expressions{$specifier}\ },\ :g)\ for\ ^10
==3058731== 
==3058731== 
==3058731== HEAP SUMMARY:
==3058731==     in use at exit: 145,285 bytes in 6 blocks
==3058731==   total heap usage: 332,497 allocs, 332,491 frees, 123,456,025 bytes allocated
==3058731== 
==3058731== 143,600 bytes in 1 blocks are possibly lost in loss record 6 of 6
==3058731==    at 0x483E77F: malloc (vg_replace_malloc.c:307)
==3058731==    by 0x4A865F7: MVM_malloc (alloc.h:2)
==3058731==    by 0x4A865F7: MVM_fixed_size_alloc (fixedsizealloc.c:203)
==3058731==    by 0x4A865F7: MVM_fixed_size_alloc (fixedsizealloc.c:179)
==3058731==    by 0x4A6B75F: hash_allocate_common (uni_hash_table.c:47)
==3058731==    by 0x4A6B75F: MVM_uni_hash_build (uni_hash_table.c:81)
==3058731==    by 0x4B345B8: generate_property_codes_by_names_aliases (unicode_ops.c:887)
==3058731==    by 0x4B345B8: MVM_unicode_name_to_property_code (unicode_ops.c:912)
==3058731==    by 0x4A5942D: MVM_interp_run (interp.c:1692)
==3058731==    by 0x1097C3: main (main.c:472)
==3058731== 
==3058731== LEAK SUMMARY:
==3058731==    definitely lost: 0 bytes in 0 blocks
==3058731==    indirectly lost: 0 bytes in 0 blocks
==3058731==      possibly lost: 143,600 bytes in 1 blocks
==3058731==    still reachable: 1,685 bytes in 5 blocks
==3058731==         suppressed: 0 bytes in 0 blocks
==3058731== Reachable blocks (those to which a pointer was found) are not shown.
==3058731== To see them, rerun with: --leak-check=full --show-leak-kinds=all
==3058731== 
==3058731== For lists of detected and suppressed errors, rerun with: -s
==3058731== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 0 from 0)
```
It looks like it should be freed here https://github.com/MoarVM/MoarVM/blob/master/src/moar.c#L721-L722, so I don't know what's going on.

MasterDuke17:
Move static variables to be members of MVMInstance instead. This is
safer if there are multiple MoarVMs running in the same process. Instead
of needing to coordinate initializing/destroying these global variables,
each instance can be responsible for its own hashes.

Additionally, free on shutdown some of the generated-on-demand hashes
that weren't being freed before.

Now valgrind and heaptrack don't report any leaks for
`raku --full-cleanup -e 'my %expressions = D => "got D", M => "got M", l => "got l"; my $a; $a := "%D %M %l".subst(/"%"$<specifier>=<{ %expressions.keys.join("|") }>/, -> ( :$specifier ) { %expressions{$specifier} }, :g) for ^10'; say "\c[LATIN CAPITAL LETTER A]"`.

Fixes #1453

### DIFF
--- a/src/core/instance.h
+++ b/src/core/instance.h
@@ -405,6 +405,12 @@ struct MVMInstance {
     /* Normal Form Grapheme state (synthetics table, lookup, etc.). */
     MVMNFGState *nfg;
 
+    /* Unicode hashes. */
+    MVMUniHashTable property_codes_by_names_aliases;
+    MVMUniHashTable property_codes_by_seq_names;
+    MVMUniHashTable codepoints_by_name;
+    MVMUniHashTable *unicode_property_values_hashes;
+
     /************************************************************************
      * Type objects for built-in types and special values
      ************************************************************************/
--- a/src/moar.c
+++ b/src/moar.c
@@ -718,8 +718,15 @@ void MVM_vm_destroy_instance(MVMInstance *instance) {
     uv_mutex_destroy(&instance->mutex_callsite_interns);
     cleanup_callsite_interns(instance);
 
-    /* Release this interpreter's hold on Unicode database */
-    MVM_unicode_release(instance->main_thread);
+    /* Clean up Unicode hashes. */
+    for (int i = 0; i < MVM_NUM_PROPERTY_CODES; i++) {
+        MVM_uni_hash_demolish(instance->main_thread, &instance->unicode_property_values_hashes[i]);
+    }
+    MVM_free_null(instance->unicode_property_values_hashes);
+
+    MVM_uni_hash_demolish(instance->main_thread, &instance->property_codes_by_names_aliases);
+    MVM_uni_hash_demolish(instance->main_thread, &instance->property_codes_by_seq_names);
+    MVM_uni_hash_demolish(instance->main_thread, &instance->codepoints_by_name);
 
     /* Clean up spesh mutexes and close any log. */
     uv_mutex_destroy(&instance->mutex_spesh_install);
--- a/src/strings/unicode.h
+++ b/src/strings/unicode.h
@@ -12,4 +12,3 @@ MVMCodepoint MVM_unicode_find_primary_composite(MVMThreadContext *tc, MVMCodepoi
 #define MVM_unicode_case_change_type_fold  3
 
 void MVM_unicode_init(MVMThreadContext *tc);
-void MVM_unicode_release(MVMThreadContext *tc);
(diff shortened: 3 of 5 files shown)


niner (approved):
I love that doing the right thing is actually less code than it was before :)

MasterDuke17:
> I love that doing the right thing is actually less code than it was before :)

Yep, though it definitely took a while to convince myself all the locking and uv_once, etc. was safe to remove.

jnthn:
> I love that doing the right thing is actually less code than it was before :)

Well, except that now if we did have multiple instances, they'd all have to build these hashes, rather than sharing a single copy, which is safe due to the data being immutable. Nobody really uses the ability to have multiple instances today, though, so far as I know.

MasterDuke17:
> > I love that doing the right thing is actually less code than it was before :)
> 
> Well, except that now if we did have multiple instances, they'd all have to build these hashes, rather than sharing a single copy, which is safe due to the data being immutable. Nobody really uses the ability to have multiple instances today, though, so far as I know.

True, but it does seem to be the case that these might never end up being generated in a bunch of use cases. I'd never seen these show up in a valgrind/heaptrack leak report before trying the example code mentioned in the commit message.

niner:
> Well, except that now if we did have multiple instances, they'd all have to
> build these hashes, rather than sharing a single copy, which is safe due to
> the data being immutable.

One the one hand yes, on the other hand these hashes were allocated with a 
specific instance's fixed size allocator. So when that instance goes away, the 
hashes would have gone anyway (or worse, left with dangling pointers).

> Nobody really uses the ability to have multiple
> instances today, though, so far as I know.

Yes, I figure that's purely theoretic. So we trade off a purely theoretic 
memory management issue against a purely theoretic performance issue :D

jnthn:
> One the one hand yes, on the other hand these hashes were allocated with a
> specific instance's fixed size allocator

Argh, that's indeed not good (and a change from when this was originally written and used uthash, which was malloc all the way...)

PR #1465, opened 2021-04-08
Fix JITed ordfirst/ordat/ordbaseat returning 4294967295 instead of -1

niner:
MVM_string_ord_at returns a MVMGrapheme32 which is a 32 bit signed integer.
However we treated it as a 64 bit integer without proper sign extension, which
turned a -1 into 4294967295. Fix by sign extend the result to a full 64 bit
signed integer.

### DIFF
--- a/src/jit/core_templates.expr
+++ b/src/jit/core_templates.expr
@@ -638,18 +638,20 @@
       (carg $1 int)) ptr_sz))
 
 (template: ordfirst
-  (call (^func &MVM_string_ord_at)
-    (arglist
-      (carg (tc) ptr)
-      (carg $1 ptr)
-      (carg (^zero) int)) int_sz))
+  (scast
+    (call (^func &MVM_string_ord_at)
+      (arglist
+        (carg (tc) ptr)
+        (carg $1 ptr)
+        (carg (^zero) int)) int_sz) int_sz 4))
 
 (template: ordat
-  (call (^func &MVM_string_ord_at)
-    (arglist
-      (carg (tc) ptr)
-      (carg $1 ptr)
-      (carg $2 int)) int_sz))
+  (scast
+    (call (^func &MVM_string_ord_at)
+      (arglist
+        (carg (tc) ptr)
+        (carg $1 ptr)
+        (carg $2 int)) int_sz) int_sz 4))
 
 (template: rindexfrom
   (call (^func &MVM_string_index_from_end)
--- a/src/jit/x64/emit.dasc
+++ b/src/jit/x64/emit.dasc
@@ -2311,6 +2311,7 @@ void MVM_jit_emit_primitive(MVMThreadContext *tc, MVMJitCompiler *compiler, MVMJ
         } else {
             | callp &MVM_string_ord_at;
         }
+        | cdqe;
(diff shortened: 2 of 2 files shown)


MasterDuke17:
Nice. But this seems like the kind of thing that could be happening elsewhere, know of any (easy) way to check?

bdw (approved):
Seems legit to me :+1:
I vaguely recall that I think the JIT should've introduced the casts itself, but apparently not, or not on this level of detail.

niner:
> Nice. But this seems like the kind of thing that could be happening
> elsewhere, know of any (easy) way to check?

Well, for a certain definition of "easy" it is. I wrote a tiny gcc plugin that 
prints all functions with a signed integer return value smaller than 64 bits:

```
import gcc
from gccutils import get_src_for_loc, pformat, cfg_to_dot, invoke_dot

outfile = open('ints', 'at')

class FindInts(gcc.GimplePass):
    def execute(self, fun):
        if fun and fun.cfg:
            if isinstance(fun.decl.result.type, gcc.IntegerType) \
                and fun.decl.result.type.sizeof < 8 \
                and not fun.decl.result.type.unsigned:
                outfile.write(str(fun.decl) + "\n")
                
ps = FindInts(name='find-ints')
ps.register_after('cfg')
```

Then acked for the printed function names in the expr JIT templates:
`for F in $(<ints) ; do ack $F src/jit/core_templates.expr ; done`

Which gives me candidates worth looking at:
```
  (call (^func &MVM_get_exception_category)
  (call (^func &MVM_nativecall_build)
  (call (^func &MVM_string_get_grapheme_at)
    (callv (^func &MVM_6model_find_method_spesh)
    (call (^func &MVM_string_ord_at)
    (call (^func &MVM_string_ord_at)
  (call (^func &MVM_string_ord_basechar_at)
  (call (^func &MVM_unicode_lookup_by_name)
```

niner:
And if you got time to check them. These are for the lego jit:
```
> for F in $(<ints) ; do ack "return $F" src/jit/graph.c ; done
    case MVM_OP_getexcategory: return MVM_get_exception_category;
    case MVM_OP_breakpoint: return MVM_debugserver_breakpoint_check;
        return consume_jumplist(tc, jg, iter, ins);
        return consume_invoke(tc, jg, iter, ins);
    case MVM_OP_getexcategory: return MVM_get_exception_category;
    case MVM_OP_breakpoint: return MVM_debugserver_breakpoint_check;
        return consume_jumplist(tc, jg, iter, ins);
        return consume_invoke(tc, jg, iter, ins);
```

and

```
> for F in $(<ints) ; do ack "$F" src/jit/x64/emit.dasc ; done
        | callp &MVM_6model_find_method_spesh;
            | callp &MVM_string_ord_at;
            | callp &MVM_string_ord_basechar_at;
```

MasterDuke17:
That's not too many, I can probably take a look.

niner:
I got curious and had an hour to work on it and indeed, getcpbyname suffered from the same issue and I missed the expr JIT version of ordbaseat. Both fixed by commit dba4b528eab88ffbed915ad5c9151a1466d31027

@MasterDuke17 thanks for asking the right question!

PR #1468, opened 2021-04-13
Two small tweaks to MVM_coerce_n_s

nwc10:
Remove duplication between `MVM_coerce_n_s` and `dtoa_grisu3`, a call to `strlen` and hence some memory writes.

### DIFF
--- a/src/core/coerce.c
+++ b/src/core/coerce.c
@@ -255,25 +255,14 @@ MVMString * MVM_coerce_u_s(MVMThreadContext *tc, MVMuint64 i) {
 }
 
 MVMString * MVM_coerce_n_s(MVMThreadContext *tc, MVMnum64 n) {
-    if (n == MVM_num_posinf(tc)) {
-        return MVM_string_ascii_decode_nt(tc, tc->instance->VMString, "Inf");
-    }
-    else if (n == MVM_num_neginf(tc)) {
-        return MVM_string_ascii_decode_nt(tc, tc->instance->VMString, "-Inf");
-    }
-    else if (n != n) {
-        return MVM_string_ascii_decode_nt(tc, tc->instance->VMString, "NaN");
-    }
+    char buf[64];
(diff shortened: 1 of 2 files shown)


src/core/coerce.c
```diff
@@ -255,25 +255,14 @@ MVMString * MVM_coerce_u_s(MVMThreadContext *tc, MVMuint64 i) {
...
-    }
-    else if (n == MVM_num_neginf(tc)) {
-        return MVM_string_ascii_decode_nt(tc, tc->instance->VMString, "-Inf");
-    }
-    else if (n != n) {
-        return MVM_string_ascii_decode_nt(tc, tc->instance->VMString, "NaN");
-    }
+    char buf[64];
+    int len = dtoa_grisu3(n, buf, sizeof(buf));
+    if (len < 0)
```
MasterDuke17, line 260: Is this even possible? Are there any floats that grisu3 + snprintf-as-fallback can't stringify?

nwc10: I can't see how it's possible, given the size of buffer provided, and what the formatting string implies. But previously the code was assuming that it needed to handle the documented error return value, so I didn't want to change that. There's probably going to be some-or-other static linting tool that will get grumpy if it's unhandled, so I thought it easier to leave it there.

nwc10:
So why did Lin_MVM_reloc SEGV on Azure for commit 3b8386c3d4cd2e1a00bbfade4896314e1394312c? I'm struggling to replicate this.

MasterDuke17:
> So why did Lin_MVM_reloc SEGV on Azure for commit [3b8386c](https://github.com/MoarVM/MoarVM/commit/3b8386c3d4cd2e1a00bbfade4896314e1394312c)? I'm struggling to replicate this.

I saw some random fails on my PRs today also, I don't think it's at all related to/caused by your commits.

nwc10:
Hold off merging this - there might be a better plan.

nwc10:
Superceded by #1472

PR #1470, opened 2021-04-15
Simplify pow_I and change to never return Nums

MasterDuke17:
To make testing easier, right now I just add a pow2_I op, but if this approach is approved I'll do the bootstrapping dance in NQP to just replace pow_I.

Compared to pow_I:
* I removed one special case (https://github.com/MoarVM/MoarVM/blob/master/src/math/bigintops.c#L624-L626) on the assumption that it's not that likely to happen repeatedly in hot paths, so I'd rather remove a branch for the common case
* Throw on negative exponents and remove the code that dealt with them (which would return a Num)
* Return an Int type object instead of Inf for too big exponents


MasterDuke17:
I just tested this and now (i.e., MoarVM 2023.12-1-gfde6796f8 + Rakudo 2023.12-7-g93009efc4) it's not any faster than 	`nqp::pow_I`. I'm kind of surprised, but at this point it doesn't seem worth more work so I'm going to close it.

PR #1471, opened 2021-04-17
Propagate spesh facts after guard elimination

niner:
This branch fixes 3 places where we turned ops into simple sets without taking the opportunity to copy the spesh facts. This manifested as not even attempting to inline a callee for lack of facts.

### DIFF
--- a/src/spesh/optimize.c
+++ b/src/spesh/optimize.c
@@ -97,6 +97,11 @@ void MVM_spesh_copy_facts(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpeshOperan
     copy_facts(tc, g, to, from);
 }
 
+static void MVM_spesh_turn_into_set(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpeshIns *ins) {
+    ins->info = MVM_op_get_op(MVM_OP_set);
+    copy_facts(tc, g, ins->operands[0], ins->operands[1]);
+}
+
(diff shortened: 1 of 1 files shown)


src/spesh/optimize.c
```diff
@@ -97,6 +97,11 @@ void MVM_spesh_copy_facts(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpeshOperan
     copy_facts(tc, g, to, from);
 }
 
+static void MVM_spesh_turn_into_set(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpeshIns *ins) {
```
zhuomingliang, line 100: `MVM_` could be removed, if it is `static` function.

niner:
> `MVM_` could be removed, if it is `static` function.

I originally named it turn_into_set but that clashed with a variable of the 
same name. I also wanted to move it into spesh/manipulate.c but copy_facts 
isn't available there. So I settled on a static MVM_spesh_turn_into_set 
figuring the prefix while being unusual wouldn't hurt...

zhuomingliang:
> 
> `MVM_` could be removed, if it is `static` function.
> I originally named it turn_into_set but that clashed with a variable of the same name. I also wanted to move it into spesh/manipulate.c but copy_facts isn't available there. So I settled on a static MVM_spesh_turn_into_set figuring the prefix while being unusual wouldn't hurt...

maybe name it `turn_into_set_op`? :)

PR #1473, opened 2021-04-18
Fix inlines missing synthetic deopt points

niner:
Since commit 3345ec36466dbae315b00960d3afedf121b08451 we have synthetic deopt
points on inserted guards. These deopt points take their index from an existing
deopt point they are linked with. This is important as otherwise we might miss
registers we need to restore when deoptimizing.

However these synth deopt points were not retained on spesh graphs created from
existing candidates for the purpose of inlining. So with these exactly what the
synth deopt points should protect us from was happening. We missed out on some
register usage and did not materialize objects optimized away by PEA.

Fix by collecting information about these synth deopt points when writing byte
code, retaining it in the spesh candidate and restoring it when creating a
control flow graph for inlining.

### DIFF
--- a/src/6model/reprs/MVMSpeshCandidate.h
+++ b/src/6model/reprs/MVMSpeshCandidate.h
@@ -75,6 +75,9 @@ struct MVMSpeshCandidateBody {
      *  There is a trailing -1 bytecode offset to mark the end of the data.
      */
     MVMint32 *deopt_usage_info;
+
+    MVMint32 *deopt_synths;
+    MVMint32 num_deopt_synths;
 };
 
 struct MVMSpeshCandidate {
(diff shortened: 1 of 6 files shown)


src/6model/reprs/MVMSpeshCandidate.h
```diff
@@ -75,6 +75,9 @@ struct MVMSpeshCandidateBody {
      *  There is a trailing -1 bytecode offset to mark the end of the data.
      */
     MVMint32 *deopt_usage_info;
+
+    MVMint32 *deopt_synths;
```
MasterDuke17, line 79: Does this need to be freed in `gc_free`?

MasterDuke17: Or added to `unmanaged_size`?

niner: Ah, yes to both. Was so happy to discover the mistake that made it segfault that I forgot that there was still the cleanup to implement.

niner: Pushed a new version that addresses your comments. Thanks for the reminder!

jnthn (approved):

PR #1474, opened 2021-04-19
Use correct size for size argument to send()/write()

MasterDuke17:
In 3734a1dcf876fd583c8ee2bfd3f6991a186dc16d it was changed to cast to
`int`, but that can go negative for large values, which causes a 'Bad
address' error. `raku -e 'say "H" x 2147483647' | wc` now correctly
reports 2147483648 characters written. Fixes the second error reported
in https://github.com/rakudo/rakudo/issues/2197#issuecomment-822633751,
but not the original error of that issue.

### DIFF
--- a/src/io/syncsocket.c
+++ b/src/io/syncsocket.c
@@ -190,7 +190,7 @@ MVMint64 socket_write_bytes(MVMThreadContext *tc, MVMOSHandle *h, char *buf, MVM
     while (bytes > 0) {
         int r;
         do {
-            r = send(data->handle, buf, (int)bytes, 0);
+            r = send(data->handle, buf, bytes > 0x40000000 ? 0x40000000 : bytes, 0);
         } while(r == -1 && errno == EINTR);
         if (MVM_IS_SOCKET_ERROR(r)) {
             MVM_gc_mark_thread_unblocked(tc);
--- a/src/io/syncfile.c
+++ b/src/io/syncfile.c
@@ -74,13 +74,13 @@ static void perform_write(MVMThreadContext *tc, MVMIOFileData *data, char *buf, 
     while (bytes > 0) {
         int r;
         do {
-            r = write(data->fd, buf, (int)bytes);
+            r = write(data->fd, buf, bytes > 0x40000000 ? 0x40000000 : bytes);
         } while (r == -1 && errno == EINTR);
         if (r == -1) {
             int save_errno = errno;
             MVM_gc_mark_thread_unblocked(tc);
-            MVM_exception_throw_adhoc(tc, "Failed to write bytes to filehandle: %s",
-                strerror(save_errno));
+            MVM_exception_throw_adhoc(tc, "Failed to write %"PRId64" bytes to filehandle: %s",
+                bytes, strerror(save_errno));
         }
         bytes_written += r;
         buf += r;


niner:
The cast to int was certainly wrong on a 64 bit machine. But what would now happen on a 32 bit machine?

MasterDuke17:
> But what would now happen on a 32 bit machine?

Wouldn't it give some error about trying to fit a too-big number instead of silently truncating?

niner:
> Wouldn't it give some error about trying to fit a too-big number instead of
> silently truncating?

Best case is a compile time error as we pass a 64 bit number to a function 
only taking a 32 bit number.
Worst case is that the number is silently truncated.

None of these cases seem to be desirable :)

MasterDuke17:
Well, I guess we could propagate the size_t farther back/up the call chain (e.g., into MVM_io_write_bytes), or instead check at the call to write()/send() that whatever we're passed fits into a size_t (since that's what's in their signatures)?

MasterDuke17:
Hm, `bytes_written` and `MVMIOFileData.byte_position` should be size_t, `r` should be ssize_t...

src/io/syncsocket.c
```diff
@@ -190,7 +190,7 @@ MVMint64 socket_write_bytes(MVMThreadContext *tc, MVMOSHandle *h, char *buf, MVM
     while (bytes > 0) {
         int r;
         do {
-            r = send(data->handle, buf, (int)bytes, 0);
+            r = send(data->handle, buf, bytes > 32767 ? 32767 : bytes, 0);
```
nwc10, line 193: I think that this should be 1073741824 too. The `send` call should fail if it what is passed is too large.

In that, the IBM document mentions TCP, UDP and RAW sockets. For UDP and RAW sockets, it matters whether the bytes are transmitted atomically (ie one packet, possibly fragmented at the IP level). The remote end is going to be reading in messages, not streams. Hence what is being sent is too big, it's better for the sender to get the error `EMSGSIZE` (and know this) than send a partial packet (pretend nothing happened), (which the receiver might not realise is truncated) and then send a second (or third) packet which starts midway through the intended message, and might confuse the receiver.

Thinking further, this code is actually incorrect (assuming you can reach it with a UDP socket), as it prevents you sending a UDP packet sized between 32767 and 65527 bytes, which is a legitimate use.

So I think that this should be 1073741824 too. (And it might be better to express that as `0x40000000` in all places - sorry, didn't think of that earlier). Trying to send large over-large payloads with UDP and RAW will fail successfully, whilst large transfers with TCP potentially will use much larger packets. (I think that it's possible for TCP to use packets > 65536ish bytes if the underlying link supports them)(yes there are buffers, but network cards can be really fast, so they might have sent your first 32767 bytes before you manage to enqueue another 32767 in the buffer)

Disclaimer - as best I can remember I've never actually written code to deal with UDP or RAW sockets. I remember playing with ["ping of death"](https://en.wikipedia.org/wiki/Ping_of_death) at the time, but IIRC there was other peoples' source code, and heck, Irix `ping` was trusting enough to send such malformed packets if you simply gave it daft comand line flags - the "feature" was standard on the OS.

MasterDuke17:
OK, I'll give that a test. Somewhat unexpectedly, explicitly using
SO_SNDBUF caused problems because it was so small (7 iirc) and threw a
bunch of exceptions in some of the networking related spectests.

Re UDP packets, I don't think they go through this code, since all the
calls to socket() (
https://github.com/MoarVM/MoarVM/blob/d209337c258729ae9e601acbd80a32a36028793b/src/io/syncsocket.c#L397
and
https://github.com/MoarVM/MoarVM/blob/d209337c258729ae9e601acbd80a32a36028793b/src/io/syncsocket.c#L429)
use SOCK_STREAM.

It's been a very long time since I've done any C network programming, I
even opened up my physical copy of Stevens' UNIX Network Programming, but
couldn't find anything which gave a definitive answer to what the best size
should be.

src/io/syncsocket.c
MasterDuke17: Done, still passes all tests and `raku -e 'say "H" x 2147483647;' | wc` reports the correct value.

nwc10:
I think that it's good. I guess the only thing we can do now is try it in the real world, as I don't think that CI systems are going to find any possible remaining problems.

PR #1475, opened 2021-04-20
Some updates and tweaks to Azure CI pipelines

MasterDuke17:
The main pipelines now use Ubuntu 20.04 (the latest available) and I added
single 18.04 and 16.04 pipelines for GCC and Clang (so four new in total)
so we still get some minimal testing on older versions.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -135,9 +135,9 @@ stages:
           displayName: "Shorten PATH on Windows"
           condition: eq( variables['Agent.OS'], 'Windows_NT' )
 
-        - script: sudo apt-get install libzstd-dev
+        - script: sudo apt-get install libzstd-dev valgrind
           condition: and(succeeded(), eq( variables['Agent.OS'], 'Linux' ))
-          displayName: Install libzstd-dev
+          displayName: Install libzstd-dev and valgrind
 
         - checkout: self
           path: selfrepo
@@ -262,3 +262,10 @@ stages:
 #            ./html-cover.sh 2
 #          condition: and(succeeded(), eq( variables['COVERAGE'], 'yes' ))
 #          displayName: Create coverage report
+
+        - script: |
+            valgrind --leak-check=full ../install/bin/raku --full-cleanup -e '' 2>valgrind.log
+            cat valgrind.log
+            ../install/bin/raku -ne 'BEGIN my @lost; @lost.push($0) if /"lost: " (\d+)/; END exit @lost.elems != 3 || all(@lost) != 0' valgrind.log
+          condition: and(succeeded(), ne( variables['RELOCATABLE'], 'yes' ), ne( variables['COVERAGE'], 'yes' ), eq( variables['Agent.OS'], 'Linux' ))
+          displayName: Check for memory leaks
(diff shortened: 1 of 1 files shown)


patrickbkr (approved):
lgtm

azure-pipelines.yml
```diff
@@ -243,3 +261,11 @@ stages:
             cp -v MoarVM-cover/merge-profraw.sh ../nqp/ &&
             ./html-cover.sh 2
           condition: and(succeeded(), eq( variables['COVERAGE'], 'yes' ))
+          displayName: Create coverage report
+
+        - script: |
+            valgrind --leak-check=full ../install/bin/raku --full-cleanup -e '' 2>valgrind.log
+            cat valgrind.log
+            ../install/bin/raku -ne 'BEGIN my @lost; @lost.push($0) if /"lost: " (\d+)/; END exit @lost.elems != 3 || all(@lost) != 0' valgrind.log
```
patrickbkr, line 269: To make sure I understand this correctly: Three entries with each 0 lost bytes is expected?

azure-pipelines.yml
```diff
@@ -49,55 +49,72 @@ stages:
...
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--has-libffi --cc=clang'
          MVM_coverage:
-           IMAGE_NAME: 'ubuntu-18.04'
+           IMAGE_NAME: 'ubuntu-20.04'
            NQP_OPTIONS: '--backends=moar'
            MOAR_OPTIONS: '--compiler=clang --coverage --optimize=0 --debug=3 --cc=clang'
-           COVERAGE: 1
-           USE_CLANG_LATEST: 1
+           COVERAGE: 'not working right now'
```
patrickbkr, line 99: Given it's not working at the moment, can you disable the job entirely (commenting out?), to not needlessly burn up CPU cycles / speed up the run?

lizmat (approved):

patrickbkr (commented):
Did we decide to just accept the occasional test failures on Windows caused by the `-j2` for now? (This is not meant as a suggestive question. I really am unsure if I've maybe missed a bit of discussion about this.)

azure-pipelines.yml
MasterDuke17: Yep, exactly correct. Looks like this:
```
==826695==    definitely lost: 0 bytes in 0 blocks
==826695==    indirectly lost: 0 bytes in 0 blocks
==826695==      possibly lost: 0 bytes in 0 blocks
```

patrickbkr: OK. Thanks for the explanation!

MasterDuke17:
> Did we decide to just accept the occasional test failures on Windows caused by the `-j2` for now? (This is not meant as a suggestive question. I really am unsure if I've maybe missed a bit of discussion about this.)

I don't think there was any definite yeas or nays. @ugexe said he prefers to only use a single job in his CI pipelines for reproducibility, but I don't believe he actually voted against this. I prefer using multiple jobs as that is more likely to uncover potential problems, even though it may make reproducing them harder. Since there are a couple approves for this PR I'll leave it as is (adding the `-j2`), but if there's a subsequent consensus against that's fine, we can revert.

In the multiple runs since that first one, I haven't seen that problem re-appear.

patrickbkr:
Looking good. 👍🏽

azure-pipelines.yml
MasterDuke17: Done in 5fad749

PR #1478, opened 2021-04-24
Fix missing gc_mark of simstackframe's arg_types

niner:
Found through: AddressSanitizer: heap-buffer-overflow src/spesh/stats.c:40 in
incomplete_type_tuple x62f000000040 is located 960 bytes to the left of 51200-
byte region [0x62f000000400,0x62f00000cc00) allocated by thread 0 in setup_bin
src/gc/gen2.c:27


niner:
@jnthn I wonder how something so obvious could have hidden for so long in plain sight. Am I missing something that makes this implicitly OK, or is the bug really as straight forward as this?

PR #1481, opened 2021-04-27
Submodule default shallow clone

nwc10:
Using *shallow* clones for the submodules saves about 38M of disk space.

If you need the history, it can easily be restored locally by running `git submodule foreach git fetch --unshallow`.

Most of us rarely look at the submodule history, so this seems like a worthwhile trade off.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,3 +20,13 @@
 	path = 3rdparty/ryu
 	url = https://github.com/MoarVM/ryu
 	ignore = untracked
+
+# adding `shallow = true` to each submodule would save about 12M on a
+# git clone --recursive
+#
+# however, adding it tickles a bug in git not fixed until 2.32.0 in that with
+# `shallow = true` here, `git submodule --quiet update` *isn't* "--quiet". If
+# `shallow = true` is found here, then the `--quiet` flag is not passed down
+# from `git submodule` to `git fetch` because `git submodule` takes a different
+# code path from the default case. *Even though* the submodule fetch it invokes
+# isn't shallow!
--- a/tools/update-submodules.pl
+++ b/tools/update-submodules.pl
@@ -19,6 +19,22 @@ Getopt::Long::GetOptions('git-cache-dir=s' => \$git_cache_dir);
 
 print 'Updating submodules .................................... ';
 
+# This git --version parsing logic adapted from the perl5 commit
+# 962ff9139336a4a2 which I wrote almost 10 years ago, and seems to have been
+# reliable and trouble free ever since:
+
+my $version_string = exec_with_output('git', '--version') // "";
+unless ($version_string =~ /\Agit version (\d+\.\d+\.\d+)(.*)/) {
+    print "\n===SORRY=== ERROR: `git --version` failed to return a parseable version string\n";
+    print "The program's output was: $version_string\n";
+}
+
+# CentOS 7 (and presumably similar vintage RedHat) ships 1.8.3
+# 1.8.4 and later have the --depth flag.
+# However, --quiet wasn't quiet with --depth, until commit 62af4bdd423f5f39,
+# which is in 2.32.0
+my $git_new_enough = eval "v$1 ge v2.32.0";
+
 exec_and_check('git', 'submodule', 'sync', '--quiet', 'Submodule sync failed for an unknown reason.');
 exec_and_check('git', 'submodule', '--quiet', 'init', 'Submodule init failed for an unknown reason.');
 
@@ -57,6 +73,12 @@ if ($git_cache_dir) {
         exec_and_check('git', 'submodule', '--quiet', 'update', '--reference', $modrefdir, $smodpath, 'Git submodule update failed.');
     }
 }
+elsif ($git_new_enough) {
+    # This saves about 38M.
+    # If you need the history, run `git submodule foreach git fetch --unshallow`
+    exec_and_check('git', 'submodule', '--quiet', 'update', '--depth', '1',
+                   'Git submodule update with --depth 1 failed.');
+}
 else {
     exec_and_check('git', 'submodule', '--quiet', 'update', 'Git submodule update failed.');
 }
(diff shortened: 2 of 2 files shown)


MasterDuke17:
While I rarely look deep into a submodule's history, I have looked a couple commits back (i.e., we've added a commit or two on top and I want to find the last upstream commit). Maybe make the depth 3 or 5 instead of 1? Does that still have most of the space saving?

nwc10:
Playing with `--depth 5` I realise that this invocation of `git submodule` is very chatty (even the original `--depth `)

For some reason the `--quiet` is getting lost. (I've seen this before with something - I think it was `--quiet` from `git fetch` doesn't get passed onwards to the `git gc --auto` that it invokes)

So I need to figure out how to shut this up.

nwc10:
Oh my, this is bonkers. If you have this diff:

```diff --git a/.gitmodules b/.gitmodules
index 566a9450d..1737c2c7c 100644
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,18 +1,27 @@
 [submodule "3rdparty/libuv"]
        path = 3rdparty/libuv
        url = https://github.com/libuv/libuv.git
+       shallow = true
 [submodule "3rdparty/dyncall"]
        path = 3rdparty/dyncall
        url = https://github.com/MoarVM/dyncall.git
+       shallow = true
 [submodule "3rdparty/dynasm"]
        path = 3rdparty/dynasm
        url = https://github.com/MoarVM/dynasm.git
+       shallow = true
 [submodule "3rdparty/libtommath"]
        path = 3rdparty/libtommath
        url = https://github.com/MoarVM/libtommath
+       shallow = true
 [submodule "3rdparty/cmp"]
        path = 3rdparty/cmp
        url = https://github.com/MoarVM/cmp.git
+       shallow = true
 [submodule "3rdparty/libatomicops"]
        path = 3rdparty/libatomicops
        url = https://github.com/MoarVM/libatomic_ops
+       shallow = true
+
+# shallow = true saves about 12M on a git clone --recursive
+# If you need the history, run `git submodule foreach git fetch --unshallow`
```

then now your `git submodule --quiet update` **isn't** because it gets all chatty (seemingly to stderr if its a tty)

nwc10:
It's a bug in git. The fix is:

```diff --git a/git-submodule.sh b/git-submodule.sh
index eb90f18229..4678378424 100755
--- a/git-submodule.sh
+++ b/git-submodule.sh
@@ -420,9 +420,9 @@ fetch_in_submodule () (
        cd "$1" &&
        if test $# -eq 3
        then
-               echo "$3" | git fetch --stdin ${2:+"$2"}
+               echo "$3" | git fetch ${GIT_QUIET:+--quiet} --stdin ${2:+"$2"}
        else
-               git fetch ${2:+"$2"}
+               git fetch ${GIT_QUIET:+--quiet} ${2:+"$2"}
        fi
 )

```

Now I just have to figure out how to submit this.

(This makes it look easy. I've spent a couple of hours with full-on `strace` output to work out what happens where in the maze of forks and threads. That `git fetch` command is only called for shallow checkouts, and in turn **it** is passing a `progress true` flag down to the `git-remote-https` process which in turn is being chatty to its **fd 1** which I infer is going down a pipe to another process that then relays it to stderr, or is being `dup`ed to stderr by its invoker. I'm still not sure why the `depth` isn't being done correctly for just `git submodule update` despite it clearly reading `.gitmodules`, or why `git --clone recursive` pulls tags for the submodules with `shallow = true` but `git submodule update --depth 1` does not. But I don't need to fix **those** to save 38M by default.)

MasterDuke17:
Is https://git.kernel.org/pub/scm/git/git.git/commit/?id=a819e2b3ef7c27d04befa0a345c9e6e3c7a7965e related?

nwc10:
It's git/git@62af4bdd423f5f3988dad97abcdd7adfba0756c4 and it turns out that it's made it into Git v2.32.0-rc

(If I read it right, it will also be backported to git maint releases, so we might be able to use it *before* Debian Bookworm)

MasterDuke17:
Git v2.32.0 has been released (I have it now on my Arch desktop). I don't know if the fix was backported, or where it needs to land for this to be mergeable?

nwc10:
I don't know what git policy on backporting is. (I assume that it's documented somewhere, but I don't know where that is).

I can see from the tags that maintenance releases are made, but also that none have been made in the past 5 months, so I'm guessing that it's unlikely that it will appear in a backport: https://git.kernel.org/pub/scm/git/git.git/refs

I don't have access to any machine that has had git v2.32.0 or v2.33.0 installed, so I can't test it against a "packaged" git. (I didn't really want to test it just on something I built from source - I wanted to confirm that I didn't do anything "special" in my build that hid real errors). So I'm reluctant to take this further *personally* until I can test like this.

As in - I think that it's good, and if you're happy to take it over and merge it, that's fine, but I'm not comfortable doing that yet. (And part of this is that I can't actually benefit from this change yet, as it's supposed to save disk space on test machines, and hence if I need my own git build to "save" space, I'm not saving anything.)

nwc10:
(Finally) I got back to this. Mostly because I

1. now have access to a machine with extremely slow external network connection...
2. and `git` with the bugfixed version

(and actually it turns out that at least one more system I have access to has a recent enough `git`)

Turns out I hadn't found and fixed all the "issues" in this yet (ie worked within the limitations of the `git` bug that is now fixed)

Basically

1. you can't put `shallow = true` in `.gitmodules` because this causes the `git` invocations from `tools/update-submodules.pl` to ignore `--quiet` and hence generate a *lot* of chatty progress output (*despite* not actually doing a shallow checkout or update)
2. I needed to change the version number in `tools/update-submodules.pl` from 1.8.4 to 2.32.0

I believe that this branch (well, single commit) is now correct, and about as good as we can get. `git clone --recursive` won't default to shallow submodules, but any "regular" build from a "regular" checkout will be shallow on new enough systems (so progressively more systems)

I'm not sure if I spent more time figuring this stuff out than I will ever spend on waiting for network traffic...

niner (approved):

MasterDuke17 (approved):

PR #1483, opened 2021-04-29
Remove unnecessary Windows specific tasks from CI

MasterDuke17:
We can use Powershell on all OSes, so do that, but then only set the few
Windows-specific options needed to get the special dev Powershell at runtime.

Also stick the install destination in a variable, so for relocatable
builds we can change that after moving the installation and subsequent
jobs can be simplified.


patrickbkr (approved):
LGTM

PR #1492, opened 2021-05-16
 Try to trap concurrent hash accesses.

nwc10:
Previously MoarVM could well SEGV if two user threads access the same hash without locking. Now if it detects concurrent access it will oops instead with a diagnostic. It's not clear to me that we can really recover if we detect this, as concurrent write accesses might have already corrupted the hash structures too far for us to resolve problems.

### DIFF
--- a/src/core/fixedsizealloc.c
+++ b/src/core/fixedsizealloc.c
@@ -58,6 +58,13 @@ void MVM_fixed_size_create_thread(MVMThreadContext *tc) {
 void MVM_fixed_size_destroy(MVMFixedSizeAlloc *al) {
     int bin_no;
 
+    /* Free anything on the overflow lists. */
+    MVMFixedSizeAllocSafepointFreeListEntry *cur = al->free_at_next_safepoint_overflows;
+    while (cur) {
+        MVM_free(cur->to_free);
+        cur = cur->next;
+    }
+
     for (bin_no = 0; bin_no < MVM_FSA_BINS; bin_no++) {
         int page_no;
         int num_pages = al->size_classes[bin_no].num_pages;
--- a/src/core/str_hash_table.c
+++ b/src/core/str_hash_table.c
@@ -40,7 +40,7 @@ MVM_STATIC_INLINE void hash_demolish_internal(MVMThreadContext *tc,
 
     size_t total_size
         = entries_size + sizeof(struct MVMStrHashTableControl) + metadata_size;
-    MVM_fixed_size_free(tc, tc->instance->fsa, total_size, start);
+    MVM_fixed_size_free_at_safepoint(tc, tc->instance->fsa, total_size, start);
 }
 
 /* Frees the entire contents of the hash, leaving you just the hashtable itself,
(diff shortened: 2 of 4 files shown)


nwc10:
@jnthn: Your example does this:

```$ ./rakudo-m -e 'my %h; await start { for ^1_000_000 { %h{$_}++ } } xx 5'                                                                   MoarVM oops: MVM_str_hash_fetch_nocheck called with a stale hashtable pointer
   at -e:1  (<ephemeral file>:)
 from -e:1  (<ephemeral file>:)
 from SETTING::src/core.c/Promise.pm6:263  (/home/nick/Perl/rakudo/blib/CORE.c.setting.moarvm:)
 from SETTING::src/core.c/ThreadPoolScheduler.pm6:882  (/home/nick/Perl/rakudo/blib/CORE.c.setting.moarvm:)
 from SETTING::src/core.c/ThreadPoolScheduler.pm6:251  (/home/nick/Perl/rakudo/blib/CORE.c.setting.moarvm:)
 from SETTING::src/core.c/ThreadPoolScheduler.pm6:245  (/home/nick/Perl/rakudo/blib/CORE.c.setting.moarvm:)
 from SETTING::src/core.c/ThreadPoolScheduler.pm6:242  (/home/nick/Perl/rakudo/blib/CORE.c.setting.moarvm:run-one)
 from SETTING::src/core.c/ThreadPoolScheduler.pm6:284  (/home/nick/Perl/rakudo/blib/CORE.c.setting.moarvm:)
 from SETTING::src/core.c/Thread.pm6:54  (/home/nick/Perl/rakudo/blib/CORE.c.setting.moarvm:THREAD-ENTRY)
```

Previously it went `SEGV`, or occasionally malloc aborted after detecting a double free.

jnthn:
> Now if it detects concurrent access it will oops instead with a diagnostic.

Certainly an improvement, if nothing else because the stack trace points at where to start looking in order to fix the problem, whereas a SEGV gives no such clue.

> It's not clear to me that we can really recover if we detect this, as concurrent write accesses might have already corrupted the hash structures too far for us to resolve problems.

Certainly not something for this PR, but it'd be interesting to consider if it's possible to make this situation recoverable. So far as I understand, this only really detects cases where we have a conflict that leads to that hash growing in size, and the control structure is replaced at these grow points, thus why we can mark the old one state?

nwc10:
Rebasing onto #1493

nwc10:
> Certainly not something for this PR, but it'd be interesting to consider if it's possible to make this situation recoverable. So far as I understand, this only really detects cases where we have a conflict that leads to that hash growing in size, and the control structure is replaced at these grow points, thus why we can mark the old one state?

Yes, correct in your understanding. For what it is doing, it can only detect a subset of data races. if two threads concurrently add then delete keys but never increase the size, this won't get spotted (and `oops`ed), but MoarVM likely could go wrong in all sorts of strange ways (possibly up to SEGV). But I realised that the approach would add very little overhead, and probably catch most naughtiness eventually.

I'm not sure that it's possible to detect **and recover** in the general case (ie try to do anything more sophisticated than this) without actual locking (on every write). In that, the worst case that I can think of is already pretty bad - you have two threads both intending to insert a fresh key. Both keys hash into the middle of a run of occupied slots. So both threads need to walk along from the ideal position, find where their key belongs (based on the *Robin Hood* offset invariants), and then potentially move some of the later keys "to the right". If one thread has already decided to move a block of memory at the same time that the second thread is reading in the middle of it, then all bets are off as to what the second thread actually reads.

Hence I can't see how you can recover from that. So you'd have to lock before every write. (And unlike a hash based on chained buckets, you'd have to hold a global lock for the hash, instead of being able to have a finer grained lock for each chain. Or some *CAS* approach to inserting into the chains)

Even the "less bad" failure cases (I suspect) can end up breaking some invariants, such as trying to add two keys at the same time in two threads might exceed the maximum probe distance.

jnthn (approved):

PR #1497, opened 2021-05-21
Define _GNU_SOURCE for GNU builds

tbrowder:
The definition results in quelling the compile warnings for libuv on Debian Buster with gcc

The changed MoarVM source files have #ifndef/#endif guards added to prevent redefinition warnings

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -230,7 +230,7 @@ if ($args{'has-sha'}) {
 }
 else { $config{shaincludedir} = '3rdparty/sha1' }
 
-# After upgrading from libuv from 0.11.18 to 0.11.29 we see very weird erros
+# After upgrading from libuv from 0.11.18 to 0.11.29 we see very weird errors
 # when the old libuv files are still around. Running a `make realclean` in
 # case we spot an old file and the Makefile is already there.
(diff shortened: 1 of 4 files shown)


niner (changes requested):
The change itself looks good, but the commit message does not explain why the change is needed. You do explain that in the PR but that will not make it into the git history and be lost to the inquiring developer.

Configure.pl
```diff
@@ -430,6 +430,9 @@ sub uniq {
 push @cflags, '-DMVM_HEAPSNAPSHOT_FORMAT=' . $config{heapsnapformat};
 push @cflags, $ENV{CFLAGS} if $ENV{CFLAGS};
 push @cflags, $ENV{CPPFLAGS} if $ENV{CPPFLAGS};
+
+push @cflags, '-D_GNU_SOURCE'; # mainly for libuv
```
MasterDuke17, line 434: What about making this `push @cflags, '-D_GNU_SOURCE' unless $args{'has-libuv'};`?

tbrowder: Okay, will do.

tbrowder:
> the commit message does not explain why the change is needed

I made a more complete explanatory comment in the last commit.

niner (approved):

MasterDuke17 (approved):

PR #1498, opened 2021-05-21
Quell compiler warnings on Linux with gcc

tbrowder:
Ensure two reference variables are initialized before use

### DIFF
--- a/src/core/interp.c
+++ b/src/core/interp.c
@@ -3355,7 +3355,7 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                 if (REPR(maybe_cu)->ID == MVM_REPR_ID_MVMCompUnit) {
                     MVMCompUnit *cu = (MVMCompUnit *)maybe_cu;
                     if (cu->body.mainline_frame) {
-                        MVMObject *coderef;
+                        MVMObject *coderef = 0;
                         for (MVMuint32 i = 0; i < cu->body.num_frames; i++) {
                             if (((MVMCode*)cu->body.coderefs[i])->body.sf == cu->body.mainline_frame) {
                                 coderef = cu->body.coderefs[i];
--- a/src/core/interp.c
+++ b/src/core/interp.c
@@ -3355,7 +3355,7 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                 if (REPR(maybe_cu)->ID == MVM_REPR_ID_MVMCompUnit) {
                     MVMCompUnit *cu = (MVMCompUnit *)maybe_cu;
                     if (cu->body.mainline_frame) {
-                        MVMObject *coderef = 0;
+                        MVMObject *coderef = NULL;
                         for (MVMuint32 i = 0; i < cu->body.num_frames; i++) {
                             if (((MVMCode*)cu->body.coderefs[i])->body.sf == cu->body.mainline_frame) {
                                 coderef = cu->body.coderefs[i];
--- a/src/strings/shiftjis.c
+++ b/src/strings/shiftjis.c
@@ -350,7 +350,7 @@ MVMuint32 MVM_string_shiftjis_decodestream(MVMThreadContext *tc, MVMDecodeStream
         MVMuint8 *bytes = (MVMuint8 *)cur_bytes->bytes;
         while (pos < cur_bytes->length || repl_pos) {
             MVMGrapheme32 graph;
-            MVMCodepoint codepoint;
+            MVMCodepoint codepoint = 0;
             MVMuint8 byte;
             int graph_is_set = 0;
             int handler_rtrn = 0;
--- a/src/strings/shiftjis.c
+++ b/src/strings/shiftjis.c
@@ -429,7 +429,7 @@ MVMuint32 MVM_string_shiftjis_decodestream(MVMThreadContext *tc, MVMDecodeStream
             last_accept_bytes = cur_bytes;
             last_accept_pos   = pos;
             total++;
-            if (MVM_string_decode_stream_maybe_sep(tc, seps, codepoint)) {
+            if (handler_rtrn == DECODE_CODEPOINT && MVM_string_decode_stream_maybe_sep(tc, seps, codepoint)) { 
                 reached_stopper = 1;
                 goto done;
             }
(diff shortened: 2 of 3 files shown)


src/core/interp.c
```diff
@@ -3355,7 +3355,7 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                 if (REPR(maybe_cu)->ID == MVM_REPR_ID_MVMCompUnit) {
                     MVMCompUnit *cu = (MVMCompUnit *)maybe_cu;
                     if (cu->body.mainline_frame) {
-                        MVMObject *coderef;
+                        MVMObject *coderef = 0;
```
niner, line 3358: But....wow, this is such an obvious bug it's kinda embarrassing that I didn't see it. I even had a look at this warning and it confused me endlessly, because there's just no way a longjmp or vfork could happen between initialization and use of the value. But actually it can, because initialization happens conditionally within the following loop. So the variable may be initialized in a compunitmainline op, then a longjmp clobbers it, then we run compunitmainline again and this time don't find a mainline_frame. It's a purely theoretical one, but a bug nonetheless. Thanks for finally solving this mystery!

One nit to pick though: while the = 0 is technically correct, we usually spell out NULL in MoarVM and I find that just more explicit.

src/strings/shiftjis.c
```diff
@@ -350,7 +350,7 @@ MVMuint32 MVM_string_shiftjis_decodestream(MVMThreadContext *tc, MVMDecodeStream
         MVMuint8 *bytes = (MVMuint8 *)cur_bytes->bytes;
         while (pos < cur_bytes->length || repl_pos) {
             MVMGrapheme32 graph;
-            MVMCodepoint codepoint;
+            MVMCodepoint codepoint = 0;
```
niner, line 353: The compiler is right: codepoint can be uninitialized if decode_handler returns DECODE_ERROR and we are in replacement mode. In this case when we pass codepoint to MVM_string_decode_stream_maybe_sep it's either uninitialized memory or the previous iteration's value. Neither would be helpful.

The one thing I'm not sure about though is if 0 is a proper value to initialize with. After all 0 is a valid code point, i.e. that of the NULL character. That's why we have the graph_is_set flag to avoid the same ambiguity there. Actually, NULL is an entirely appropriate separator character in some situations. Think `xargs -0`

niner (changes requested):

src/core/interp.c
tbrowder: I'll change to NULL then.

src/strings/shiftjis.c
tbrowder: What about moving the var up a level and initializing to NULL?

niner:
> What about moving the var up a level and initializing to NULL?

That would mean that in the error case it would contain the value of the 
previous iteration. Whether this would work or not would depend fully on how 
this algorithm and the Shift-JIS encoding work. I know too little about both 
to say anything with certainty.

src/strings/shiftjis.c
MasterDuke17: @samcv ping

samcv: Line 432 is currently:
```
if (MVM_string_decode_stream_maybe_sep(tc, seps, codepoint)) {
```
should be
```
if (handler_rtrn == DECODE_CODEPOINT && MVM_string_decode_stream_maybe_sep(tc, seps, codepoint)) {
```

tbrowder: I'll try that.

samcv: Please don't initialize `codepoint` to `0`. If you do that, then we won't get any compiler warnings that it is used before initialization. Because if `codepoint` is used before initialized, that indicates a programming error.

tbrowder: I did still get the original warnings when I did NOT initialize it. I'll try again without initializing it...

tbrowder: I used a gcc pragma to turn off the uninitialized warning. No warnings for our source now.

MasterDuke17: However, now there are warnings about unknown pragmas in clang and msvc. Unfortunately you need to do something like https://github.com/MoarVM/MoarVM/blob/master/src/profiler/heapsnapshot.c#L916-L929 to be clean across all three compilers.

tbrowder: Thanks, I'll apply the pragmas in like manner.

tbrowder: I have tried several ways to separate pre-processing options for 'gcc' and 'clang' by macro defs '__GNUC__' and '__clang__' but I haven't been successful. It's a chicken or egg situation since I can fix it for either compiler but not the other. At this point I'm not sure what to try next.  

My gut tells me that something is awry with the C preprocessor or at least how or when it is called in the build process. Clang and GCC pragma defs are leaking into each other's pre-processing definitions.

tbrowder: Some progress: appears both compilers define __GNUC__ so I have to find the real distinguishing macro between the two to continue.

tbrowder: That's above my pay grade to determine whether the compiler is correct on the uninitialized warning, but at least the warnings can be squelched if desired.

MasterDuke17:
@samcv ping. AIUM (As I understand myself) in https://github.com/MoarVM/MoarVM/pull/1498#discussion_r641624709 I'm not sure the proposed fix was correct. Are you able to take a look at this again?

samcv:
@MasterDuke17 please ping me again in 7 days if I haven't replied to this message yet. Sorry for the delay in answering the question. Will try to get to it, otherwise please ping me (after 7 days). Thanks.

samcv:
Thanks for the ping @MasterDuke17. After thinking about this further. I think we should continue with the merge request as it is. So we set the compiler to not tell us this warning, and keeping my suggestion about `handler_rtrn == DECODE_CODEPOINT`. I am going to approve the MR now.

samcv (approved):

MasterDuke17:
The changes in src/strings/decode_stream.h don't seem to be necessary anymore. At HEAD of master, neither gcc 10.3.0 nor clang 12.0.0 show a warning there. There aren't any warnings in src/strings/shiftjs.c at HEAD of master either, but since @samcv has said that's a good change, @tbrowder can you drop the first, fourth, and fifth commits from this and just keep the second and third?

tbrowder:
I think so, but I am in the middle of selling my house so it might be a few day before I can get to it.

tbrowder:
I have not forgotten, working on it now...

PR #1504, opened 2021-06-03
Close socket on async error during an async connection

Kaiepi:
Related to https://github.com/rakudo/rakudo/issues/4370.

### DIFF
--- a/src/io/asyncsocket.c
+++ b/src/io/asyncsocket.c
@@ -564,6 +564,8 @@ static void on_connect(uv_connect_t* req, int status) {
             MVM_repr_push_o(tc, arr, tc->instance->boot_types.BOOTStr);
             MVM_repr_push_o(tc, arr, tc->instance->boot_types.BOOTInt);
         });
+        uv_close((uv_handle_t*)ci->socket, free_on_close_cb);
+        ci->socket = NULL;
     }
     MVM_repr_push_o(tc, t->body.queue, arr);
     MVM_free(req);


Zer0-Tolerance:
Hi , any idea when this will be merged ?

Zer0-Tolerance:
Hi ,
Raku is still having issue when closing connection after timeout on OSX in version 2021.07, any idea how this can be fixed ?

Kaiepi:
@Zer0-Tolerance Asynchronous connections don't close on timeout on master, but do with these changes, but again, I'm testing on OpenBSD, and I don't have an OS X system to test with offhand.

Kaiepi:
The golf for this is similar to [before](https://github.com/rakudo/rakudo/pull/4391#issuecomment-895145873):

```raku
use v6;
shell "ktrace dump ktrace.out -p $*PID";
say try { await IO::Socket::Async.connect: '127.0.0.1', 12345 } // $!;
sleep;
```

But the process for producing a dump and viewing it with the more BSD-ish flavour of `ktrace`:

```sh
$ ktrace -p $*PID
$ kdump | less
```

Looks to be equivalent to this on [newer versions of OS X](https://www.unix.com/man-page/mojave/1/ktrace/):

```sh
$ ktrace dump ktrace.out -p $*PID
$ ktrace trace -R ktrace.out | less
```

niner:
Thanks for the fix! And sorry for the long delay.

PR #1506, opened 2021-06-07
Set sc.idx during deserialization to avoid costly lookup later

niner:
When we deserialize a collectable, we actually already know which SC it
belongs to and the index in this SC. So far we've already told the the
collectable about the SC, but not about the index. So the first call to
MVM_sc_find_object_idx had to do a linear search of the SC to find the index.
Fix by setting the idx field in the collectable header on deserialization.

This brings installation of rakudo core modules down from 7.89s to 7.21s
and reduces CPU time of rakudo make test from 138.98s to 134.27s.

### DIFF
--- a/src/6model/serialization.c
+++ b/src/6model/serialization.c
@@ -2283,6 +2283,7 @@ static void stub_stable(MVMThreadContext *tc, MVMSerializationReader *reader, MV
 
     /* Set the STable's SC. */
     MVM_sc_set_stable_sc(tc, st, reader->root.sc);
+    MVM_sc_set_idx_in_sc((MVMCollectable*) st, i);
 
     /* Set STable read position, and set current read buffer to the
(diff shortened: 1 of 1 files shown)


nwc10:
I'm confident that I don't know enough to review the correctness of the commit itself (ie are there any subtle corner cases possible that testing doesn't cover) but I am wondering what planet CodeFactor is on.

Specifically, CodeFactor seems to have found "new" problems in 3 unchanged functions, and one issue fixed in a fourth unchanged function. I **assume** that code factor is generally useful, but its failure modes seem very strange. (i've seen similar "changed" reports on unchanged code before.)

Oh, thinking aloud - do they have different versions installed? ie do our reports come from different variants of their freemium product, but as github's state is common, running the report against a different variant produces this sort of noise?

MasterDuke17 (approved):

Issue #433, opened 2016-10-18
MVM_dir_mkdir mishandles EEXIST


geekosaur:
mkdir_p will return -1/EEXIST if a component of the path to the new directory is a non-directory. MVM_dir_mkdir assumes -1/EEXIST means that the directory was not created because it already existed, and does not consider other possibilities.

Found this while trying to diagnose a different issue (on WIndows MVM_dir_mkdir may succeed in creating a directory while returning failure) --- but since that issue is also characterized by an unexpected failure with EEXIST, it may well be related.

geekosaur:
ftr this was fixed about 10 minutes after I filed it.

PR #1507, opened 2021-06-15
Fix mkdir error handling

McSinyx:
Before this, if a file is in place of the target path (not on Windows)
or a parent directory is successfully created but the children,
`mkdir_p` still reports success.  On *nix, this can be tested in Raku via:

```raku
mkdir '/usr/bin/env';
mkdir '/tmp/' ~ 'universe' x 42;
```

Related: GH-433.

### DIFF
--- a/src/io/dirops.c
+++ b/src/io/dirops.c
@@ -44,15 +44,11 @@ static int mkdir_p(MVMThreadContext *tc, char *pathname, MVMint64 mode) {
             ch = *p;
             *p  = '\0';
 #ifdef _WIN32
-            if (CreateDirectoryW(pathname, NULL)) {
-                created = 1;
-            }
+            created = CreateDirectoryW(pathname, NULL);
 #else
-            if (uv_fs_stat(NULL, &req, pathname, NULL) <= 0) {
-                if (mkdir(pathname, mode) != -1) {
-                    created = 1;
-                }
-            }
+            created = (uv_fs_stat(NULL, &req, pathname, NULL) == 0
+                       && S_ISDIR(req.statbuf.st_mode)
+                       || mkdir(pathname, mode) == 0);
 #endif
             if (!(*p = ch)) break;
         }
@@ -102,7 +98,7 @@ void MVM_dir_mkdir(MVMThreadContext *tc, MVMString *path, MVMint64 mode) {
     MVM_free(wpathname);
 #else
 
-    if (mkdir_p(tc, pathname, mode) == -1 && errno != EEXIST) {
+    if (mkdir_p(tc, pathname, mode) == -1) {
         int mkdir_error = errno;
         MVM_free(pathname);
         MVM_exception_throw_adhoc(tc, "Failed to mkdir: %s", strerror(mkdir_error));


MasterDuke17:
```
src/io/dirops.c: In function ‘mkdir_p’:
src/io/dirops.c:48:33: error: ‘uv_fs_t’ {aka ‘struct uv_fs_s’} has no member named ‘statbug’; did you mean ‘statbuf’?
   48 |                 && !S_ISDIR(req.statbug.st_mode)
      |                                 ^~~~~~~
```

McSinyx:
I'm so sorry for my carelessness :sweat_smile:  It's built successfully locally now, is there anyway to perform integration testing?

MasterDuke17:
https://dev.azure.com/MoarVM/MoarVM/_build/results?buildId=605&view=logs&j=1f3a7134-3b80-549a-eaff-49041f6b71c8&t=428659df-f0a2-5fc4-7f6c-dcc94dfc92b7&l=96 looks like rakudo's `make install` is failing.

src/io/dirops.c
```diff
@@ -35,31 +35,28 @@ static int mkdir_p(MVMThreadContext *tc, wchar_t *pathname, MVMint64 mode) {
...
     char *p = pathname, ch;
-    uv_fs_t req;
 #endif
+
     int created = 0;
 
-    for (;; ++p)
-        if (!*p || IS_SLASH(*p)) {
+    while (*p++)
+        if (*p == '\0' || IS_SLASH(*p)) {
```
zhuomingliang, line 43: I can't understand here, if `*p == '\0'` is true, why we need `*p = '\0';` after `ch = *p;` ?

zhuomingliang: I think other codes that's not about the bug should not be changed. Because it may causes more bugs. :)

McSinyx: I changed the condition into comparing with `\0` to make it clear what we are looking for either end of string or a slash to call mkdir on (micro-optimization speaking I should change it into `IS_SLASH(*p) || *p == '\0'`).  If I understand the original code correctly, `ch` is a temporary character to store the current one when it is changed into the null terminator so that the mkdir call will be on the parent directory only.

McSinyx: > I think other codes that's not about the bug should not be changed. Because it may causes more bugs. :)

I won't argue with that, reverting irrelevant changes.

zhuomingliang: Well, could do make a long dir `/tmp/path/to/long/dir` in the gdb and print `p`  to understand it. I did it long time ago :)

zhuomingliang:
After a calm thinking, It may be not a bug.  You could fix it with a `try`:

```
try mkdir '/usr/bin/env';
try mkdir '/tmp/' ~ 'universe' x 42;
```

McSinyx:
What I meant was that both of those should fail (well for the latter it should fail on ext4 with the filename limit of 255 at least) instead of silencing the errors.

McSinyx:
Regarding [the CI failure](https://dev.azure.com/MoarVM/MoarVM/_build/results?buildId=609&view=logs&j=4def8e43-62b3-5cca-1c32-fbd5a1f0442a&t=6f17dd53-63e8-5611-e660-0bfbef84498f), is [rakudo/t/04-nativecall/21-callback-other-thread.t](https://github.com/rakudo/rakudo/blob/master/t/04-nativecall/21-callback-other-thread.t) flaky?  It doesn't seem to be related.

MasterDuke17:
> Regarding [the CI failure](https://dev.azure.com/MoarVM/MoarVM/_build/results?buildId=609&view=logs&j=4def8e43-62b3-5cca-1c32-fbd5a1f0442a&t=6f17dd53-63e8-5611-e660-0bfbef84498f), is [rakudo/t/04-nativecall/21-callback-other-thread.t](https://github.com/rakudo/rakudo/blob/master/t/04-nativecall/21-callback-other-thread.t?rgh-link-date=2021-06-15T14%3A36%3A40Z) flaky? It doesn't seem to be related.

Yeah, there's a known problem/race-condition with renaming files on Windows that random jobs run into. I restarted the job.

zhuomingliang:
I think this patch is not good enough. Could you try this one？
```
if (uv_fs_stat(NULL, &req, pathname, NULL) <= 0) {
                if (mkdir(pathname, mode) != -1) {
                    created = 1;
                } else {
                    created = 0;
                    break;
                }
} else {
     created = 0;
     break;
}
```

McSinyx:
`uv_fs_stat(NULL, &req, pathname, NULL) <= 0` will always yield 1 because `stat` returns either 0 or -1.  We also need to take into account if the current path exists and is a directory.  Me adding `chmod` earlier was a mistake though, in neither NetBSD nor GNU coreutils `mkdir -p` would try to change the mode of a directory it didn't create.

zhuomingliang:
Good, it could be nice to have a test in https://github.com/raku/roast before someone merge it.

McSinyx:
Thanks!  This means Rakudo 2021.06 was out, and `repl` looks like fun!

PR #1511, opened 2021-06-28
Fix memory corruption by accessing freed spesh stats.

niner:
When cleaning up spesh stats that didn't get updated for a while, we free the
MVMSpeshStats structure and remove the reference from the MVMStaticFrameSpesh
object holding it. It may still be referenced from a MVMSpeshSimStackFrame
however and when we continue processing this sim stack, we access some random
memory.

Look through all tc's spesh_sim_stacks for frames still referencing the
MVMSpeshStats and just not free it if we find one.

### DIFF
--- a/src/spesh/stats.c
+++ b/src/spesh/stats.c
@@ -657,26 +657,59 @@ void MVM_spesh_stats_update(MVMThreadContext *tc, MVMSpeshLog *sl, MVMObject *sf
  * updated in a while, clears them out. */
 void MVM_spesh_stats_cleanup(MVMThreadContext *tc, MVMObject *check_frames) {
     MVMint64 elems = MVM_repr_elems(tc, check_frames);
-    MVMint64 insert_pos = 0;
-    MVMint64 i;
-    for (i = 0; i < elems; i++) {
-        MVMStaticFrame *sf = (MVMStaticFrame *)MVM_repr_at_pos_o(tc, check_frames, i);
-        MVMStaticFrameSpesh *spesh = sf->body.spesh;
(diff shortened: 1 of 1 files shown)


src/spesh/stats.c
```diff
@@ -668,9 +668,37 @@ void MVM_spesh_stats_cleanup(MVMThreadContext *tc, MVMObject *check_frames) {
...
+                    }
+                }
+                current = current->body.next;
+            }
+
+            uv_mutex_unlock(&tc->instance->mutex_threads);
+
+            if (!found) {
+                MVM_spesh_stats_destroy(tc, ss);
+                MVM_free(spesh->body.spesh_stats);
```
MasterDuke17, line 699: MVM_free_null?

src/spesh/stats.c
```diff
@@ -668,9 +668,37 @@ void MVM_spesh_stats_cleanup(MVMThreadContext *tc, MVMObject *check_frames) {
...
+
+            MVMThread *current = tc->instance->threads;
+            int found = 0;
+            while (current && !found) {
+                MVMThreadContext *cur_tc = current->body.tc;
+                if (cur_tc) {
+                    MVMSpeshSimStack *sims = cur_tc->spesh_sim_stack;
+                    if (sims) {
+                        MVMuint32 i = 0;
+                        for (; i < sims->used; i++) {
```
MasterDuke17, line 683: Since we’re C99 now, why not just put the declaration in the for?

MasterDuke17 (approved):
Without claiming to know much about this code, the change seems reasonable.

Issue #821, opened 2018-03-22
UTF8 decodestream doesn't throw after IO::Handle.slurp


samcv:
This may have to be handled on the Rakudo side, but I'm putting this here for now.
```perl6
Buf.new(0xe2, 0x99).decode #> throws
# Doesn't throw
my $fh = open '/tmp/temppp', :w; $fh.write: Buf.new(0xe2, 0x99); $fh.close; my $fh2 = open '/tmp/temppp', :r; $fh2.slurp
# throws
'/tmp/temppp'.IO.slurp
```

MasterDuke17:
Fixed by https://github.com/MoarVM/MoarVM/pull/1519

PR #1519, opened 2021-07-22
Stream Decoder: Disallow incomplete code at EOF

dyknon:
Break unwanted infinite loop caused by
$ echo 0: c2 | xxd -r | rakudo -ne ''

Related to: MoarVM/MoarVM#821

### DIFF
--- a/src/strings/decode_stream.c
+++ b/src/strings/decode_stream.c
@@ -174,6 +174,9 @@ static void reached_eof(MVMThreadContext *tc, MVMDecodeStream *ds) {
     /* Decode all the things. */
     if (ds->bytes_head)
         run_decode(tc, ds, NULL, NULL, DECODE_EOF);
+    if (ds->bytes_head)
+        MVM_exception_throw_adhoc(tc,
+            "Incomplete character found at the end of a stream");
 
     /* If there's some things left in the normalization buffer, take them. */
     MVM_unicode_normalizer_eof(tc, &(ds->norm));


MasterDuke17:
Should the return value of the `run_decode()` call be checked also (or instead)?

dyknon:
I don't think return values of the `run_decode()` are useful to detect this error.
The return value depends on the number of decoded characters, not the bytes left in the input buffer.

MasterDuke17:
I'm inclined to merge this, but hopefully a final question. Would it make sense to put the remaining bytes in the error message, similarly to https://github.com/MoarVM/MoarVM/blob/master/src/strings/utf8.c#L238-L259 ?

MasterDuke17:
@dyknon ping

dyknon:
That's a good idea.
But why does it uses conversion specifications "%02hhx" for dumped bytes.
Because of the default argument promotion, "%02x" is enough, and it is a more compatible way, I think.
Sorry for my slow response.

MasterDuke17:
> But why does it uses conversion specifications "%02hhx" for dumped bytes.
> Because of the default argument promotion, "%02x" is enough, and it is a more compatible way, I think.

Apparently I originally had `"%02x"`, but changed it in https://github.com/MoarVM/MoarVM/commit/ff007f2ff93c74ab87c2861cdb83f37af3c31f78. However, that was long enough ago that I don't remember what was wrong about `"%02x"`, probably some warning from one of the three compilers we use in our CI pipelines (gcc, clang, MSVC). Feel free to give it a try, but we'll need to make sure it's warning free.

dyknon:
OK. I use `%02hhx`.
Change error message to include remaining bytes.

patrickbkr:
So, is this ready for merging?

MasterDuke17:
> So, is this ready for merging?

I think so.

dyknon:
I'm ready.

PR #1523, opened 2021-07-27
Templates for (bind|get)attr(s)?_*

MasterDuke17:
NQP builds ok and passes `make m-test` and Rakudo builds ok and passes `make m-test m-spectest`. Looking at a couple random spesh logs I don't see the `expr bail: Cannot get template for: ` messages for these ops anymore.


bdw (approved):
Looks good... testing will have to prove if it works :-)

PR #1524, opened 2021-07-30
Handle sp_runbytecode_s in rewrite_obj_return

MasterDuke17:


src/spesh/inline.c
```diff
@@ -1056,6 +1056,23 @@ static void return_to_box(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpeshBB *re
...
+                                      unbox_op == MVM_OP_unbox_u ? MVM_reg_int64 :
+                                      unbox_op == MVM_OP_unbox_n ? MVM_reg_num64 :
+                                                                   MVM_reg_str;
+    /* Replace the return with an unboxing instruction. */
+    MVMSpeshOperand ver_target      = MVM_spesh_manipulate_new_version(tc, g, target.reg.orig);
+    MVMSpeshOperand *unbox_operands = MVM_spesh_alloc(tc, g, 2 * sizeof(MVMSpeshOperand));
+    return_ins->info                = MVM_op_get_op(unbox_op);
+    return_ins->operands            = unbox_operands;
+    unbox_operands[0]               = ver_target;
+    unbox_operands[1]               = return_ins->operands[0];
```
jnthn, line 1071: Hm, this looks wrong. I think we're meant to update `return_ins->operands = unbox_operands` *after* the population of `unbox_operands`.

src/spesh/inline.c
```diff
@@ -1122,12 +1139,16 @@ static void rewrite_obj_return(MVMThreadContext *tc, MVMSpeshGraph *g,
     case MVM_OP_sp_runbytecode_v:
         MVM_spesh_manipulate_delete_ins(tc, g, return_bb, return_ins);
         break;
+    case MVM_OP_sp_runbytecode_s:
+        return_to_unbox(tc, g, return_bb, return_ins, runbytecode_ins->operands[0],
+            MVM_OP_unbox_s);
+        break;
```
jnthn, line 1145: Could add it for `_i` and `_n` too, given I think it's mostly copy-paste.

src/spesh/inline.c
MasterDuke17: Well, in happens in this order in `return_to_box` https://github.com/MoarVM/MoarVM/pull/1524/files#diff-b4fbc9fb84a8bb3a3b9d3371048903867537c428ae7be8a781d92574bbfb84c5R1036-R1039

jnthn: Yes, but it's being stuck on a freshly allocated `MVMSpeshIns` there; here we're updating the one instruction.

MasterDuke17: Ah, right. However, just looking at `return_to_set` made me think maybe should we instead have a `return_to_op`? Then `return_to_set` and `return_to_unbox` (which already takes an op as an argument) can be combined.

MasterDuke17: Then the diff would look like this https://gist.github.com/MasterDuke17/a5842d1e8fa5a45cafc7c1b5d8d07a38

jnthn: @MasterDuke17 Yes, that's a good way to do it. :+1:

PR #1531, opened 2021-08-27
Convert the systems calls hash to MVMFixKeyHashTable

nwc10:
The new-disp systems call hash had had to be implemented as `MVMStrHashTable` with a second helper struct to indirect, because

1. `MVMFixKeyHashTable` would always allocate storage for the hash entry from the FSA
2. The systems call hash entries needed to be static storage in the C code

This branch

1. Adds a new mode to `MVMFixKeyHashTable` so that it can also use static storage
2. Converts the systems call hash to use this
3. Marks the hash keys with `MVM_gc_root_add_permanent_desc` and hence eliminates a loop from `MVM_gc_root_add_instance_roots_to_worklist`

The code does now assign to the static storage used for the `MVMDispSysCall` structs, meaning that they can't be marked as `const` in the C code, and hence more efficiently initialised at process start up and shared between running processes. (*nix system, at least, can mark constant data as read only and shared). However, the trade off seems worth it, as we eliminate a loop at every GC run, and likely some level of CPU cache pressure.

The same approach *could* be extended to convert `container_registry` from `MVMFixKeyHashTable` to `MVMStrHashTable` 
suggesting that this approach is suitable generic. However, I've not done that yet becase

1. It requires changing `MVM_6model_add_container_config` to take a *non*-`const` `MVMContainerConfigurer *`
2. That function is marked `MVM_PUBLIC`

However it seems that *nothing* outside of `containers.c` uses that function, so possible it could

1. Be dropped from the public API
2. Even made `static`

This would enable us to remove another iteration loop from `MVM_gc_root_add_instance_roots_to_worklist`

### DIFF
--- a/src/core/instance.h
+++ b/src/core/instance.h
@@ -279,7 +279,7 @@ struct MVMInstance {
     MVMDispRegistry disp_registry;
 
     /* MoarVM system calls hash (VM-provided functionality). */
-    MVMStrHashTable syscalls;
+    MVMFixKeyHashTable syscalls;
 
     /* Identity mapping for arguments (argument 0 is at position 0, argument
      * 1 is at position 1, etc.) Allocated at startup with an initial number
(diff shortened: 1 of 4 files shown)


nwc10:
> It can be. It was public because Rakudo used to have an implementation of its `Scalar` container implemented alongside its extops. However, a generalized version of that already moved into MoarVM, and we're on the path to eliminate extops entirely.

OK, #1533 removes it from the API, when then permits some further simplifications.

PR #1540, opened 2021-09-16
Comment that `MVM_callsite_intern` and `is_common` are coupled.

nwc10:
If you change one you need to change the other.


MasterDuke17 (approved):

jnthn (approved):
This is helpful; a more robust way might be to have the interned flag on the callsite be just a bit field and add another bit for common, and have an `MVM_callsite_intern_common` or some such.

PR #1542, opened 2021-09-17
Convert the use of a Variable Length Array to a regular allocation

nwc10:
This should restore the build on MSVC.

### DIFF
--- a/build/setup.pm
+++ b/build/setup.pm
@@ -342,7 +342,7 @@ our %COMPILERS = (
         ld => undef,
         as => 'as',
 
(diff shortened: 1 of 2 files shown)


nwc10:
@jnthn are my assumptions about the FSA being a better fit reasonable?

I haven't actually benchmarked them :-)

nwc10:
@coke I thought it clearer (for history) to merge my changes to your original commit into one change. It's not clearer for attribution - this is still fine with you?

nwc10:
`alloca` would better.

coke:
@nwc10 - sure. The config change was ok, but my code change was a WIP that needed fixing anyway, thanks!

nwc10:
celo_duro suggested `alloca`. It seems the simplest approach - see #1543

PR #1543, opened 2021-09-17
Convert the use of a Variable Length Array to alloca

nwc10:
This should restore the build on MSVC.


coke:
Looks good on windows, builds ok, no (new) test failures on windows new-disp

jnthn (approved):

PR #1545, opened 2021-09-20
Add _n cases to jitting some of the new ops

MasterDuke17:
NQP build ok and passes `make m-test` and Rakudo builds ok and passes `make m-test m-spectest`. These ops are no longer seen in a spesh log of compiling CORE.c.setting as not being jitted.


jnthn (approved):
At first I wondered if this needed some special code path for dealing with floating point, but since we only store a location to write the return value to in these ops, it seems there's nothing to handle in that regard.

PR #1548, opened 2021-09-22
Add GCC computed goto for dispatcher.New disp cgoto

zhuomingliang:
Works locally, though the code is a bit ugly.


src/disp/program.c
```diff
@@ -2677,302 +2677,331 @@ MVMuint32 MVM_disp_program_record_end(MVMThreadContext *tc, MVMCallStackDispatch
...
+#if MVM_CGOTO
+#define DISPATCH(op)
+#define OP(name) name
+#define NEXT do { \
+    if(i < dp->num_ops) {       \
+        op = dp->ops[i++];      \
+        goto *LABELS[op.code];  \
+    } else                      \
+        goto acception;         \
+    } while(0)
```
jnthn, line 2691: I think we can get rid of the `i < dp->num_ops` check here and just put a `return 1` into the terminals; every dispatch program ends with one of the `MVMDispOpcodeResult*` ops.

zhuomingliang: `goto acception` can be replaced with `return 1`,  I don't understand the part of getting rid of `i < dp->num_ops` because I know nothing about dispatcher :)

PR #1549, opened 2021-09-22
add dispatcher-drop-n-args to optimize allocations

timo:
Instead of creating a MVMCapture and MVMCallsite for each step of
removing arguments, we now offer a syscall that drops multiple arguments
that live at the same index in one go.

The result is that the transformations tree can now contain null entries
for the capture entry, which we have to interpret and deal with.


jnthn:
Manually merged a slight tweaked version; thanks!

PR #1552, opened 2021-09-30
Jit some ops that aren't terribly hot, but do cause bails in a spesh log

MasterDuke17:

This also includes a fix for scobjcount, which was incorrectly jitted (but had never been a problem before because other not-jitted-at-the-time ops were preventing the jitted version of scobjcount from being used).


MasterDuke17:
Not needed after https://github.com/MoarVM/MoarVM/pull/1555 was merged instead.

PR #1555, opened 2021-10-02
Jit some not so common ops

MasterDuke17:


MasterDuke17:
Ok, so we know that jitting setdebugtypename [`b8e230e` (#1555)](https://github.com/MoarVM/MoarVM/pull/1555/commits/b8e230ea8e74d4081c3381cc10a64cd63077c87b) has a problem. I'm going to remove that commit for now and test the rest from https://github.com/MoarVM/MoarVM/pull/1552.

PR #1558, opened 2021-10-05
Change CI job strategy from matrix to parallel

MasterDuke17:
We aren't passing multiple options to the matrix strategy for it to do
all combinations of, so this might give us more control about how the
jobs start (e.g., start the Windows jobs first since they're slower).

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -26,7 +26,7 @@ stages:
     # leaves little space for the name.
     - job: T
       strategy:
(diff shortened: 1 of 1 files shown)


MasterDuke17:
Whoops, found some new documentation that explains matrix vs parallel better, this isn't right (obviously, because no jobs ran (but it just silently doesn't do anything, thanks Azure for making it so easy to debug)).

PR #1559, opened 2021-10-06
Assorted debug server fixes

jnthn:
* Fix missed breakpoints on Windows due to inconsistent application of path normalization
* Avoid breakpoints being hit many times for the same line
* Avoid failures to resume that were liable to occur in multi-threaded programs
* Make sure all debugger debug output is guarded


timo (approved):
LGTM, thank you very much

PR #1581, opened 2021-10-27
Migrate special return to callstack and simplify return handling

jnthn:
Work on `new-disp` introduced new callstack unwind logic, but conservatively tried not to touch the overall workflow of frame unwinding too much. However, once we more fully embrace the new callstack model, simplifications ore on offer.

A prerequisite for that is changing the special return mechanism to use callstack unwinding rather than a check in `remove_one_frame`. There is a further benefit to this: we also can allocate space for any data associated with special return handlers on the callstack rather than using the FSA or `malloc`.

One usage of the special return mechanism was for running finalizers. This was the only use that did not install the special return handler on the currently executing frame, and thus was not possible to port use the new mechanism, which installs the special return record on the callstack. After playing with a few ideas, the simplest one - with relatively low cost - turned out to be just checking if anything needs finalizing, and trying to run a handler if so. This is probably more robust than what it replaces, since we don't go trying to find a frame to put the handler on down the call chain.

With the special return handling migrated, it then became possible to greatly simplify `remove_one_frame`, it could now rely on the unwinding process to leave the interpreter in the correct state. The "thunked" flag could also go away thanks to this. Interpreter boundary checks went away thanks to having a callstack record type to mark such boundaries. Some work was also eliminated for callstack frames. Finally, an inlining and check elimination the compiler failed to do decreases the cost of returning from JIT-compiled frames.

This seems to be worth ~5% off a recursive Fibonacci benchmark.

### DIFF
--- a/src/6model/parametric.c
+++ b/src/6model/parametric.c
@@ -77,17 +77,11 @@ static void finish_parameterizing(MVMThreadContext *tc, void *sr_data) {
         }
         uv_mutex_unlock(&tc->instance->mutex_parameterization_add);
     });
-
-    /* Clean up parametric return data, now we're finished with it. */
-    MVM_free(prd);
 }
-static void mark_parameterize_sr_data(MVMThreadContext *tc, MVMFrame *frame, MVMGCWorklist *worklist) {
-    ParameterizeReturnData *prd = (ParameterizeReturnData *)frame->extra->special_return_data;
+static void mark_parameterize_sr_data(MVMThreadContext *tc, void *sr_data, MVMGCWorklist *worklist) {
+    ParameterizeReturnData *prd = (ParameterizeReturnData *)sr_data;
     MVM_gc_worklist_add(tc, worklist, &(prd->parametric_type));
     MVM_gc_worklist_add(tc, worklist, &(prd->parameters));
-}
-static void free_parameterize_sr_data(MVMThreadContext *tc, void *sr_data) {
-    MVM_free(sr_data);
 }
 void MVM_6model_parametric_parameterize(MVMThreadContext *tc, MVMObject *type, MVMObject *params,
                                         MVMRegister *result) {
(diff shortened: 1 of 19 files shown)


niner:
Just got this segfault during core.c build:
```
Program terminated with signal SIGSEGV, Segmentation fault.

warning: Section `.reg-xstate/12159' in core file too small.
#0  0x00007f90aef01b57 in MVM_hll_current (tc=tc@entry=0xe08e30) at src/core/hll.c:185
185         return tc->cur_frame->static_info->body.cu->body.hll_config;
[Current thread is 1 (Thread 0x7f90ae8b8b80 (LWP 12159))]
(gdb) bt
#0  0x00007f90aef01b57 in MVM_hll_current (tc=tc@entry=0xe08e30) at src/core/hll.c:185
#1  0x00007f90aef2122e in MVM_gc_finalize_run_handler (tc=tc@entry=0xe08e30) at src/gc/finalize.c:87
#2  0x00007f90aeefafc8 in MVM_callstack_unwind_frame (tc=tc@entry=0xe08e30, exceptional=exceptional@entry=0 '\000') at src/core/callstack.c:737
#3  0x00007f90aeef475b in remove_one_frame (tc=0xe08e30, unwind=<optimized out>) at src/core/frame.c:882
#4  0x00007f90aeef6c8a in MVM_frame_try_return (tc=0xe08e30) at src/core/frame.c:950
#5  0x00007f90aeedf1d4 in MVM_interp_run (tc=0xe08e30, initial_invoke=0x0, initial_invoke@entry=0x7f90aeff58b0 <toplevel_initial_invoke>, invoke_data=0x0, invoke_data@entry=0x7f90aeff58b0 <toplevel_initial_invoke>, outer_runloop=outer_runloop@entry=0x0) at src/core/interp.c:570
#6  0x00007f90aeff6985 in MVM_vm_run_file (instance=<optimized out>, filename=<optimized out>) at src/moar.c:505
#7  0x000000000040160d in main (argc=11, argv=0x7ffcdcc55db8) at src/main.c:307
(gdb) p tc->cur_frame
$1 = (MVMFrame *) 0x0
```

jnthn:
Rebased to fix a conflict and also the NULL dereference observed.

niner (approved):

PR #1585, opened 2021-10-28
Dispatcher replace arg syscall

timo:
Lets us replace a single argument with a tracked value or a literal object value.

This saves the creation of one BOOTCapture when you used to drop and then insert with the same index.

The replace-arg syscall when inserted into the nqp dispatchers saves just 20 captures in `rakudo -e ''` but the replace-arg-literal-obj syscall saves about 1.5k.

### DIFF
--- a/src/6model/reprs/MVMCapture.c
+++ b/src/6model/reprs/MVMCapture.c
@@ -429,7 +429,7 @@ MVMObject * MVM_capture_insert_arg(MVMThreadContext *tc, MVMObject *capture_obj,
  *
  * The callsite argument type is expected to be the same. */
 MVMObject * MVM_capture_replace_arg(MVMThreadContext *tc, MVMObject *capture_obj, MVMuint32 idx,
-        MVMRegister value) {
+        MVMCallsiteEntry kind, MVMRegister value) {
     MVMCapture *capture = validate_capture(tc, capture_obj);
     if (idx > capture->body.callsite->num_pos)
         MVM_exception_throw_adhoc(tc, "Capture argument index out of range");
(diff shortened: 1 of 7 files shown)


src/6model/reprs/MVMCapture.c
```diff
@@ -423,3 +423,45 @@ MVMObject * MVM_capture_insert_arg(MVMThreadContext *tc, MVMObject *capture_obj,
...
+
+    /* Allocate a new capture before we begin; this is the only GC allocation
+     * we do. */
+    MVMCallsite *callsite = capture->body.callsite;
+    MVMCallsite *new_callsite = MVM_callsite_copy(tc, capture->body.callsite);
+    MVMObject *new_capture;
+    MVMCallsiteEntry kind = new_callsite->arg_flags[idx];
+    MVMROOT(tc, capture, {
+        if (kind & (MVM_CALLSITE_ARG_OBJ | MVM_CALLSITE_ARG_STR)) {
+            MVMROOT(tc, value.o, {
```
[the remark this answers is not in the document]
timo: it could be an int or num value, in which case rooting it could be quite explosive

src/disp/program.c
```diff
@@ -1210,6 +1210,98 @@ MVMObject * MVM_disp_program_record_capture_insert_arg(MVMThreadContext *tc,
...
+
+    /* First, create an entry as if we had dropped the argument.
+     * We save the work of creating the capture here, because it is not
+     * used by anything - it is anonymous and can not be addressed. */
+    MVMDispProgramRecordingCapture dropped_arg_record = {
+        .capture = NULL,
+        .transformation = MVMDispProgramRecordingDrop,
+        .index = idx,
+    };
+    MVM_VECTOR_INIT(dropped_arg_record.captures, 0);
```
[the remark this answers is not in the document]
timo: i think the one below can stay at 0 if the capture isn't modified any further in the future, but the first one definitely can have a 1 here

src/core/callsite.c
```diff
@@ -431,3 +432,35 @@ MVMCallsite * MVM_callsite_insert_positional(MVMThreadContext *tc, MVMCallsite *
...
+MVMCallsite * MVM_callsite_replace_positional(MVMThreadContext *tc, MVMCallsite *cs, MVMuint32 idx,
+        MVMCallsiteFlags flag) {
+    /* Can only do this with positional arguments and non-flattening callsite. */
+    if (idx > cs->num_pos)
+        MVM_exception_throw_adhoc(tc, "Cannot replace positional in callsite: index out of range");
+    if (cs->has_flattening)
+        MVM_exception_throw_adhoc(tc, "Cannot transform a callsite with flattening args");
+
+    /* Allocate a new callsite and set it up. */
+    MVMCallsite *new_callsite = MVM_malloc(sizeof(MVMCallsite));
```
jnthn, line 447: Could this not have a shortcut for when the new flag is exactly the same as the existing one, so we'd produce an identical callsite anyway?

timo: i tried that and got into memory management trouble. we could check if it's interned and in that case we can just return it unchanged, though. otherwise we'll need to clone, or else we'll double-free stuff

jnthn: Ah, fair point.

jnthn (approved):

PR #1586, opened 2021-10-29
Eliminate hllbool/boot-boolify-boxed-int pairs in spesh

niner:
No need to turn an int into an HLL bool just to turn it back to an int when
using it as a condition for jumps. Eliminate those pairs same as we do with
box/unbox.

### DIFF
--- a/src/spesh/optimize.c
+++ b/src/spesh/optimize.c
@@ -96,6 +96,12 @@ void MVM_spesh_copy_facts(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpeshOperan
                           MVMSpeshOperand from) {
     copy_facts(tc, g, to, from);
 }
+static int is_syscall(MVMThreadContext *tc, MVMSpeshFacts *facts,
+                      void (*syscall) (MVMThreadContext *tc, MVMArgs arg_info)) {
+    return facts->flags & MVM_SPESH_FACT_KNOWN_VALUE
+        && REPR(facts->value.o)->ID == MVM_REPR_ID_MVMCFunction
+        && ((MVMCFunction*)facts->value.o)->body.func == syscall;
+}
 
 static void MVM_spesh_turn_into_set(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpeshIns *ins) {
     ins->info = MVM_op_get_op(MVM_OP_set);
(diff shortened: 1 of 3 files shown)


src/spesh/optimize.c
```diff
@@ -2525,6 +2528,56 @@ static void try_eliminate_box_unbox_pair(MVMThreadContext *tc, MVMSpeshGraph *g,
...
+    MVMSpeshUseChainEntry *user_entry = MVM_spesh_get_facts(tc, g, set_ins->operands[0])->usage.users;
+    while (user_entry) {
+        MVMSpeshIns *user = user_entry->user;
+        if (user->info->opcode == MVM_OP_sp_runcfunc_i) {
+            MVMSpeshFacts *dispatch_facts = MVM_spesh_get_facts(tc, g, user->operands[1]);
+            if (dispatch_facts->flags & MVM_SPESH_FACT_KNOWN_VALUE
+                    && REPR(dispatch_facts->value.o)->ID == MVM_REPR_ID_MVMCFunction
+                    && ((MVMCFunction*)dispatch_facts->value.o)->body.func
+                        == MVM_disp_syscall_boolify_boxed_int_impl
+            ) {
```
jnthn, line 2542: Lifting this out to an `is_syscall` function would be convenient, as I suspect it's far from the last one we'll want to lower.

src/spesh/optimize.c
```diff
@@ -2525,6 +2528,56 @@ static void try_eliminate_box_unbox_pair(MVMThreadContext *tc, MVMSpeshGraph *g,
...
+                (user->info->opcode == MVM_SSA_PHI && user->info->num_operands == 2)) {
+            walk_set_looking_for_unbool(tc, g, bb, ins, user);
+        }
+        user_entry = user_entry->next;
+    }
+    if (MVM_spesh_usages_is_used(tc, g, ins->operands[0])) {
+    }
+    else {
+        MVM_spesh_manipulate_delete_ins(tc, g, bb, ins);
+    }
```
jnthn, line 2579: Just `if (!MVM_spesh_usages_is_used(tc, g, ins->operands[0]))` to avoid the empty if branch?

niner:
I addressed all comments by amending the commit. Please have another look.

jnthn (approved):

PR #1590, opened 2021-11-02
Migrate special return to callstack

jnthn:
This is one part of #1581, to see if it also trips up the CI.


jnthn:
Closing as debugging is complete; #1581 has been updated with the fixed implementation.

PR #1592, opened 2021-11-05
libffi nativecall fixes

nwc10:


src/6model/reprs/NativeCall.c
```diff
@@ -47,6 +47,13 @@ static void copy_to(MVMThreadContext *tc, MVMSTable *st, void *src, MVMObject *d
         memcpy(dest_body->arg_info, src_body->arg_info, src_body->num_args * sizeof(MVMObject*));
     }
     dest_body->ret_type = src_body->ret_type;
+#ifdef HAVE_LIBFFI
+    dest_body->ffi_ret_type = src_body->ffi_ret_type;
+    if (src_body->ffi_arg_types) {
+        dest_body->ffi_arg_types = MVM_malloc(sizeof(ffi_type *) * (dest_body->num_args ? dest_body->num_args : 1));
+        memcpy(dest_body->ffi_arg_types, src_body->ffi_arg_types, sizeof(ffi_type *) * (dest_body->num_args ? dest_body->num_args : 1));
```
niner, line 54: Could factor out the size calculation to avoid the repetition and also shorten those lines.

niner (approved):

src/6model/reprs/NativeCall.c
nwc10: Agree, but that's also true of the (unchanged) code immediately above, so I thought I'd keep it consistent with the (poor) style.

More pragmatically - once this new-disp work is done and settled, I'd like to get back to **all** this code and address some bigger issues that I see with it

1. Do we need to special case 0 as 1, or can we just store a NULL pointer if there are 0 arguments?
2. The FSA would be more efficient than malloc, as we don't reallocate and we know the size at free time
3. Likely we should merge (at least) the two allocations needed for dyncall into one memory block (pointers followed by 16-bit integers), and possibly also combine that with this third allocation for libffii only
4. change the handling of the `values` array in libffi - it seems nuts to malloc an array of pointers, and **then** malloc a whole load of small things that they point to, one per argument.

PR #1594, opened 2021-11-07
Jit ctx(outer|caller)(skipthunks)?

MasterDuke17:
Currently this causes the rakudo installation to die with `Frame has no lexical with name '::?PACKAGE'` and a ton of fails in `make m-test`. However, everything passes with `MVM_JIT_EXPR_DISABLE=1`. Since there aren't any templates for these ops, there likely is some weird interaction with other templates. Isolating this change into its own branch for ease of testing.


MasterDuke17:
Hm. After a rebase, now rakudo installs, but `MVM_JIT_EXPR_DISABLE=1` doesn't seem to make any difference for the tests (i.e., still a ton of fails).

PR #1595, opened 2021-11-09
New disp nativecall

niner:
This is a dispatcher based rework of how native routines are called. As much work as possible like unboxing arguments is pushed into the HLL dispatcher code where it can be speshed away. JIT compilation of native calls is completely re-done to be more in line with other ops. We no longer JIT compile calls when building the native call site, but let spesh handle this via the new sp_runnativecall family of ops. With the old native call JIT code gone, we can now remove the last remainder of the old invocation protocol (including the args buffer).


nwc10:
All the CI tests fail because they attempt to use MoarVM from this branch with nqp and rakudo master. What would be needed are nqp and rakudo on their respective `new-disp-nativecall` branches.

nwc10:
Builds on the M1 Mac on the gcc compiler farm. (`t/04-nativecall/02-simple-args.t` fails `not ok 12 - passed uint8 0xFE` both on this branch and master due to dyncall making buggy assumptions about ABI requirements for unsigned types smaller than `int` and sign extension which don't matter on x86 or x86_64)

Builds on x86_64 and ppc64 with both libffi and dyncall, and arm32 (I only tested 32 bit `with` dyncall). Not exhaustive, but looks like it's good on 32 and 64 bit linux, big and little endian.

Probably hasn't been tested by anyone on win32, and CI won't help us.

src/core/validation.c
```diff
@@ -26,9 +26,6 @@ enum {
     MARK_regular  = ' ',
     MARK_special  = '.',
     MARK_sequence = ':',
-    MARK_head     = '+',
-    MARK_body     = '*',
-    MARK_tail     = '-'
```
jnthn, line 31: Oh nice, this isn't a simplification I realized we'd get :-)

src/disp/program.c
```diff
@@ -819,6 +826,72 @@ static MVMuint32 value_index_how(MVMThreadContext *tc, MVMDispProgramRecording *
     return MVM_VECTOR_ELEMS(rec->values) - 1;
 }
 
+/* Ensures we have a values used entry for the specified HOW read. */
```
jnthn, line 829: s/HOW read/unbox/

src/disp/program.c
```diff
@@ -819,6 +826,72 @@ static MVMuint32 value_index_how(MVMThreadContext *tc, MVMDispProgramRecording *
...
+static MVMuint32 value_index_unbox_int(MVMThreadContext *tc, MVMDispProgramRecording *rec,
+        MVMuint32 from_value) {
+    /* Look for an existing such value. */
+    MVMuint32 i;
+    for (i = 0; i < MVM_VECTOR_ELEMS(rec->values); i++) {
+        MVMDispProgramRecordingValue *v = &(rec->values[i]);
+        if (v->source == MVMDispProgramRecordingUnboxValue &&
+                v->how.from_value == from_value)
```
jnthn, line 837: I think the `->how` here is copy pasto, and it should be `->unbox`. Also I think it has to compare `kind` as well as `from_value`.

src/disp/program.c
```diff
@@ -819,6 +826,72 @@ static MVMuint32 value_index_how(MVMThreadContext *tc, MVMDispProgramRecording *
...
+    MVMDispProgramRecordingValue new_value;
+    memset(&new_value, 0, sizeof(MVMDispProgramRecordingValue));
+    new_value.source = MVMDispProgramRecordingUnboxValue;
+    new_value.unbox.from_value = from_value;
+    new_value.unbox.kind = MVM_CALLSITE_ARG_INT;
+    MVM_VECTOR_PUSH(rec->values, new_value);
+    return MVM_VECTOR_ELEMS(rec->values) - 1;
+}
+
+/* Ensures we have a values used entry for the specified HOW read. */
```
jnthn, line 851: s/HOW read/unbox/

src/disp/program.c
```diff
@@ -819,6 +826,72 @@ static MVMuint32 value_index_how(MVMThreadContext *tc, MVMDispProgramRecording *
...
+static MVMuint32 value_index_unbox_num(MVMThreadContext *tc, MVMDispProgramRecording *rec,
+        MVMuint32 from_value) {
+    /* Look for an existing such value. */
+    MVMuint32 i;
+    for (i = 0; i < MVM_VECTOR_ELEMS(rec->values); i++) {
+        MVMDispProgramRecordingValue *v = &(rec->values[i]);
+        if (v->source == MVMDispProgramRecordingUnboxValue &&
+                v->how.from_value == from_value)
```
jnthn, line 859: Same concern as above

src/disp/program.c
```diff
@@ -819,6 +826,72 @@ static MVMuint32 value_index_how(MVMThreadContext *tc, MVMDispProgramRecording *
...
+
+/* Ensures we have a values used entry for the specified unboxed str. */
+static MVMuint32 value_index_unbox_str(MVMThreadContext *tc, MVMDispProgramRecording *rec,
+        MVMuint32 from_value) {
+    /* Look for an existing such value. */
+    MVMuint32 i;
+    for (i = 0; i < MVM_VECTOR_ELEMS(rec->values); i++) {
+        MVMDispProgramRecordingValue *v = &(rec->values[i]);
+        if (v->source == MVMDispProgramRecordingUnboxValue &&
+                v->how.from_value == from_value)
```
jnthn, line 881: And another `how`

src/disp/program.c
```diff
@@ -1012,6 +1085,96 @@ MVMObject * MVM_disp_program_record_track_attr(MVMThreadContext *tc, MVMObject *
...
+    /* Ensure the tracked value is an object type. */
+    if (((MVMTracked *)tracked_in)->body.kind != MVM_CALLSITE_ARG_OBJ)
+        MVM_oops(tc, "Can only use dispatcher-track-unbox-int on a tracked object");
+
+    /* Resolve the tracked value. */
+    MVMCallStackDispatchRecord *record = MVM_callstack_find_topmost_dispatch_recording(tc);
+    MVMuint32 value_index = find_tracked_value_index(tc, &(record->rec), tracked_in);
+
+    /* Obtain the object and ensure it is a concrete P6opaque; also track its
+     * type and concreteness since unboxing safety depends on this. */
```
jnthn, line 1098: I think the P6opaque bit here is inaccurate.

src/disp/program.c
```diff
@@ -1012,6 +1085,96 @@ MVMObject * MVM_disp_program_record_track_attr(MVMThreadContext *tc, MVMObject *
...
+MVMObject * MVM_disp_program_record_track_unbox_num(MVMThreadContext *tc, MVMObject *tracked_in) {
+    /* Ensure the tracked value is an object type. */
+    if (((MVMTracked *)tracked_in)->body.kind != MVM_CALLSITE_ARG_OBJ)
+        MVM_oops(tc, "Can only use dispatcher-track-unbox-num on a tracked object");
+
+    /* Resolve the tracked value. */
+    MVMCallStackDispatchRecord *record = MVM_callstack_find_topmost_dispatch_recording(tc);
+    MVMuint32 value_index = find_tracked_value_index(tc, &(record->rec), tracked_in);
+
+    /* Obtain the object and ensure it is a concrete P6opaque; also track its
```
jnthn, line 1127: P6opaque mention here probably also not correct

src/disp/program.c
```diff
@@ -1012,6 +1085,96 @@ MVMObject * MVM_disp_program_record_track_attr(MVMThreadContext *tc, MVMObject *
...
+MVMObject * MVM_disp_program_record_track_unbox_str(MVMThreadContext *tc, MVMObject *tracked_in) {
+    /* Ensure the tracked value is an object type. */
+    if (((MVMTracked *)tracked_in)->body.kind != MVM_CALLSITE_ARG_OBJ)
+        MVM_oops(tc, "Can only use dispatcher-track-unbox-str on a tracked object");
+
+    /* Resolve the tracked value. */
+    MVMCallStackDispatchRecord *record = MVM_callstack_find_topmost_dispatch_recording(tc);
+    MVMuint32 value_index = find_tracked_value_index(tc, &(record->rec), tracked_in);
+
+    /* Obtain the object and ensure it is a concrete P6opaque; also track its
```
jnthn, line 1157: P6opaque again

src/disp/program.c
```diff
@@ -1012,6 +1085,96 @@ MVMObject * MVM_disp_program_record_track_attr(MVMThreadContext *tc, MVMObject *
...
+    /* Read the value. */
+    MVMRegister attr_value;
+    attr_value.s = MVM_repr_get_str(tc, read_from);
+
+    /* Ensure that we have this value read in the values table, and make
+     * a tracked object if not. */
+    MVMuint32 result_value_index = value_index_unbox_str(tc, &(record->rec), value_index);
+    if (!record->rec.values[result_value_index].tracked)
+        record->rec.values[result_value_index].tracked = MVM_tracked_create(tc,
+                attr_value, MVM_CALLSITE_ARG_STR);
```
jnthn, line 1174: Missing `MVMROOT` of `attr_value.s` (`MVM_tracked_create` allocates).

src/disp/program.c
```diff
@@ -2864,7 +3080,7 @@ MVMuint32 MVM_disp_program_record_end(MVMThreadContext *tc, MVMCallStackDispatch
                 run_dispatch(tc, record, record->outcome.delegate_disp,
                         record->outcome.delegate_capture, thunked);
             else
-                MVM_exception_throw_adhoc(tc, "Dispatch callback failed to delegate to a dispatcher");
+                MVM_oops(tc, "Dispatch callback failed to delegate to a dispatcher");
```
jnthn, line 3083: Why? This is a mistake in the dispatcher implementation, not an internal VM error.

src/jit/graph.c
```diff
@@ -3877,6 +3765,82 @@ static MVMint32 consume_ins(MVMThreadContext *tc, MVMJitGraph *jg,
         jg_append_label(tc, jg, reentry_label);
         break;
     }
+    case MVM_OP_sp_runnativecall_v:
+    case MVM_OP_sp_runnativecall_i: {
```
jnthn, line 3769: We're not JIT-compiling native calls that return floating point, string, or pointer types yet, iiuc?

src/jit/graph.c
```diff
@@ -3877,6 +3765,82 @@ static MVMint32 consume_ins(MVMThreadContext *tc, MVMJitGraph *jg,
...
+        node->u.runnativecall.num_args        = callsite->flag_count - 1;
+        node->u.runnativecall.return_type     =
+            op == MVM_OP_sp_runnativecall_v
+                ? MVM_RETURN_VOID
+                : op == MVM_OP_sp_runnativecall_i
+                    ? MVM_RETURN_INT
+                    : op == MVM_OP_sp_runnativecall_s
+                        ? MVM_RETURN_STR
+                        : op == MVM_OP_sp_runnativecall_n
+                            ? MVM_RETURN_NUM
```
jnthn, line 3825: I believe these are unreachable.

src/jit/graph.c
```diff
@@ -3877,6 +3765,82 @@ static MVMint32 consume_ins(MVMThreadContext *tc, MVMJitGraph *jg,
...
+        node->u.runnativecall.entry_point     = body->entry_point;
+        node->u.runnativecall.map             = &ins->operands[2 + start];
+        node->u.runnativecall.reentry_label   = reentry_label;
+        jg_append_node(jg, node);
+        /* append reentry label */
+        jg_append_label(tc, jg, reentry_label);
+        /* Check the entry_point after we've used it to generate the code to avoid a race
+         * condition where the check turned out fine, but entry_point got overwritten before we
+         * got to use it for generating the code. */
+        if (!body->entry_point) {
```
jnthn, line 3837: Yowser...this can happen? Reposession?

src/spesh/manipulate.c
```diff
@@ -481,9 +481,10 @@ MVMSpeshBB *MVM_spesh_manipulate_split_BB_at(MVMThreadContext *tc, MVMSpeshGraph
     /* Step two: update all idx fields. */
     new_bb->idx = bb->idx + 1;
     {
-        MVMSpeshBB *ptr = linear_next;
+        MVMSpeshBB *ptr = g->entry;
         while (ptr != NULL) {
-            ptr->idx += 1;
+            if (ptr != new_bb && ptr->idx > bb->idx)
+                ptr->idx += 1;
             ptr = ptr->linear_next;
```
jnthn, line 488: Why was this wrong originally? I don't see it.

src/spesh/disp.c
```diff
@@ -1057,6 +1110,217 @@ static int translate_dispatch_program(MVMThreadContext *tc, MVMSpeshGraph *g,
...
+                    cur_op++;
+                }
+
+                /* Add the operand with the bytecode to run. */
+                rb_ins->operands[cur_op] = temporaries[op->res_code.temp_invokee];
+                MVM_spesh_usages_add_by_reg(tc, g, rb_ins->operands[cur_op], rb_ins);
+                cur_op++;
+
+                MVMuint8 *is_rw_operand = MVM_spesh_alloc(tc, g, rb_op->num_operands * sizeof(MVMuint8));
+                memset(is_rw_operand, 0, rb_op->num_operands * sizeof(MVMuint8));
```
jnthn, line 1191: Pretty sure `MVM_spesh_alloc` gives zeroed memory.

src/spesh/disp.c
```diff
@@ -1057,6 +1110,217 @@ static int translate_dispatch_program(MVMThreadContext *tc, MVMSpeshGraph *g,
...
+                MVMSpeshOperand *rw_operands = MVM_spesh_alloc(tc, g, rb_op->num_operands * sizeof(MVMSpeshOperand));
+
+                MVMSpeshFacts *object_facts = MVM_spesh_get_facts(tc, g, temporaries[op->res_code.temp_invokee]);
+                MVMNativeCallBody *body = MVM_nativecall_get_nc_body(tc, object_facts->value.o);
+                if (object_facts->flags & MVM_SPESH_FACT_KNOWN_VALUE) {
```
jnthn, line 1196: Not sure it's safe to use `object_facts->value.o` before checking it's a known value.

src/spesh/disp.c
```diff
@@ -1057,6 +1110,217 @@ static int translate_dispatch_program(MVMThreadContext *tc, MVMSpeshGraph *g,
...
+                }
+
+
+                int box_return_value = 0;
+                int has_return_value = (ins->info->opcode != MVM_OP_dispatch_v);
+                if (ins->info->opcode == MVM_OP_dispatch_o) {
+                    /* Let's see if we can turn it into a more JITable instruction */
+                    MVMSpeshFacts *object_facts = MVM_spesh_get_facts(tc, g, temporaries[op->res_code.temp_invokee]);
+                    MVMNativeCallBody *body = MVM_nativecall_get_nc_body(tc, object_facts->value.o);
+                    if (object_facts->flags & MVM_SPESH_FACT_KNOWN_VALUE) {
```
jnthn, line 1143: Also here we use `value.o` a bit early, though since I noticed it already, I wonder if we shouldn't check this property (and get the invokee facts) earlier.

src/spesh/disp.c
```diff
@@ -1057,6 +1110,217 @@ static int translate_dispatch_program(MVMThreadContext *tc, MVMSpeshGraph *g,
...
+                        emit_tri_op(tc, g, bb->linear_next, &post_call_instructions, MVM_OP_box_i,
+                            box_target_reg,
+                            rb_ins->operands[0],
+                            rb_ins->operands[3]);
+
+                        MVMSpeshBB *box_bb = MVM_spesh_manipulate_split_BB_at(
+                            tc, g, bb->linear_next, post_call_instructions);
+
+                        bb->linear_next->children[1] = target_bb;
+                        bb->linear_next->succ[1] = target_bb;
```
jnthn, line 1287: Hm, wondering if `MVM_spesh_manipulate_add_successor` wouldn't be useful here?

src/spesh/disp.c
```diff
@@ -1057,6 +1110,217 @@ static int translate_dispatch_program(MVMThreadContext *tc, MVMSpeshGraph *g,
...
+                        MVMSpeshFacts *result_facts = MVM_spesh_get_facts(tc, g, ins->operands[0]);
+                        result_facts->flags |= MVM_SPESH_FACT_KNOWN_TYPE;
+                        result_facts->type = return_type_facts->type;
+                    }
+                    MVM_spesh_manipulate_release_temp_reg(tc, g, rb_ins->operands[0]);
+                }
+
+                /* Make sure we delay release of temporaries since optimization
+                 * can add further ones. */
+                delay_temps_release = 1;
```
jnthn, line 1321: I believe this needs `optimize.c` to cooperate (to actually release them), but don't see a change there.

jnthn (commented):
Really pleased to see native calls integrated into the new dispatch mechanism.

src/core/nativecall_dyncall.c
```diff
@@ -820,3 +820,495 @@ MVMObject * MVM_nativecall_invoke(MVMThreadContext *tc, MVMObject *res_type,
...
+                            (MVMint64)*(DCpointer *)free_rws[num_rws]);
+                        break;
+                    default:
+                        MVM_telemetry_interval_stop(tc, interval_id, "nativecall invoke failed");
+                        MVM_exception_throw_adhoc(tc, "Internal error: unhandled dyncall argument type");
+                }
+                num_rws++;
+            }
+            /* Perform CArray/CStruct write barriers. */
+            MVM_nativecall_refresh(tc, value);
```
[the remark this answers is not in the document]
niner: The cases above only cover RW args. We have to refresh non-RW CStructs and CArrays as well, as it's their contents that could get modified. Same happens on master: https://github.com/MoarVM/MoarVM/blob/master/src/core/nativecall_dyncall.c#L801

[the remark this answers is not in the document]
jnthn: Ah, got it.

niner:
> > +    MVMRegister attr_value;
> 
> +    attr_value.s = MVM_repr_get_str(tc, read_from);
> +
> +    /* Ensure that we have this value read in the values table, and make
> +     * a tracked object if not. */
> +    MVMuint32 result_value_index = value_index_unbox_str(tc,
> &(record->rec), value_index); +    if
> (!record->rec.values[result_value_index].tracked)
> +        record->rec.values[result_value_index].tracked =
> MVM_tracked_create(tc, +                attr_value, MVM_CALLSITE_ARG_STR);
> 
> Missing `MVMROOT` of `attr_value.s` (`MVM_tracked_create` allocates).

Are you sure about needing MVMROOT? attr_value.s is not used anymore after the 
call to MVM_tracked_create and none of the other callers of MVM_tracked_create 
in the file MVMROOT their register values, so apparently MVMROOT is not needed 
to keep the object alive either.

niner:
> MVM_exception_throw_adhoc(tc, "Internal error: unhandled dyncall argument
> type");
> 
> Wonder if including the type in the error could help if we ever hit this

Yes, it would. However, omitting such detail is pretty much par for the course 
in MoarVM. Often the first thing I do when investigating a new bug is adding 
something like "got %s instead" to the error output. I rarely commit those 
howerver as the errors usually only appear due to MoarVM or rakudo bugs and 
once they are fixed, the added detail is essentially dead code that makes 
refactoring more tedious and is easy to miss.

jnthn:
> Are you sure about needing MVMROOT? attr_value.s is not used anymore after the
> call to MVM_tracked_create and none of the other callers of MVM_tracked_create
> in the file MVMROOT their register values, so apparently MVMROOT is not needed
> to keep the object alive either.

Checked again, and it appears I somehow misread. I agree no `MVMROOT` is needed here.

> Yes, it would. However, omitting such detail is pretty much par for the course
> in MoarVM.

It has been; I guess the argument for doing so is more clues in the event we have to debug things we can't reproduce, although how far it helps is another question.

MasterDuke17:
> MVM_exception_throw_adhoc(tc, "Internal error: unhandled dyncall argument type"); Wonder if including the type in the error could help if we ever hit this
> Yes, it would. However, omitting such detail is pretty much par for the course in MoarVM. Often the first thing I do when investigating a new bug is adding something like "got %s instead" to the error output. I rarely commit those howerver as the errors usually only appear due to MoarVM or rakudo bugs and once they are fixed, the added detail is essentially dead code that makes refactoring more tedious and is easy to miss.

Fwiw, I’ve made a couple commits in the past that just went through adding context and debugging information where it looked like it was missing, and I usually make such changes part of any other PR if needed.

nwc10:
Historically I have tended to think of it more like this;

> It has been; I guess the argument for doing so is more clues in the event we have to debug things we can't reproduce, although how far it helps is another question.

than this:

> ... as the errors usually only appear due to MoarVM or rakudo bugs and once they are fixed, the added detail is essentially dead code that makes refactoring more tedious and is easy to miss.

I found it immensely frustrating to get Perl 5 bug reports where the reporter accurately pasted the "panic" message, and then you go look into the source code and realise that you can't tell why the panic was reached. Unless it's obvious how to replicate the test case, you're immediately into a round trip delay with the bug reporter, before you can get more information to progress. it's very much like "assertion failed", where the first thing you want to do is re-run the code in the debugger, and then go look to see what the **actual** value was, because all you know so far is something that it's **not**.

So I tended to do the following

1. make sure each "panic" message was **unique** - if there are 2+ in the same C function, make sure that the text is different enough that it's clear which was hit
2. try to put in as much data into the format string so that one had a reasonable idea **why** the panic was hit, on the assumption that worst-case the bug isn't repeatable (at least not locally), and maybe the only report you get is the panic message

However, I rarely hit the case where there was enough refactoring going on such that the panic messages got in the way of it. I guess this reflects age and stability of code bases, and so I can't really comment on the different trade offs to make when this starts to matter.

niner:
> 1. make sure each "panic" message was **unique** - if there are 2+ in the
> same C function, make sure that the text is different enough that it's
> clear which was hit 2. try to put in as much data into the format string so
> that one had a reasonable idea **why** the panic was hit, on the assumption
> that worst-case the bug isn't repeatable (at least not locally), and maybe
> the only report you get is the panic message

Ok, truth is, I was just too lazy to amend those commits. But I just did so 
anyway, because yes, I wish we had error messages like you describe 
everywhere. So I may as well be part of the solution instead of part of the 
problem. Adjusted all error messages to be unique and to contain the failing 
value.

jnthn:
> make sure each "panic" message was unique - if there are 2+ in the same C function, make sure that the text is different enough that it's clear which was hit

I've certainly been trying to do this one, at least...

> So I may as well be part of the solution instead of part of the problem.

...yup, I'll keep it in mind too. :-)

niner:
> > -                MVM_exception_throw_adhoc(tc, "Dispatch callback failed
> > to delegate to a dispatcher");
> +                MVM_oops(tc, "Dispatch callback failed to delegate to a
> dispatcher");
> 
> Why? This is a mistake in the dispatcher implementation, not an internal VM
> error.

Just an accidental commit. Changed it to oops locally because that's easier to 
set a break point on.

> > +    case MVM_OP_sp_runnativecall_v:
> +    case MVM_OP_sp_runnativecall_i: {
> 
> We're not JIT-compiling native calls that return floating point, string, or
> pointer types yet, iiuc?

Floats and strings, no. Pointer types are ints though and spesh turns 
dispatch_o into sp_runnativecall_i when the result is a pointer type and adds 
appropriate boxing ops (which PEA may optimize away). Keeps the JIT 
implementation as simple as possible _and_ gives spesh some more 
opportunities.

> > +                    : op == MVM_OP_sp_runnativecall_s
> 
> +                        ? MVM_RETURN_STR
> +                        : op == MVM_OP_sp_runnativecall_n
> +                            ? MVM_RETURN_NUM
> 
> I believe these are unreachable.

As of now, yes. But since we'll want to JIT compile those in the future as 
well, I figure, we may as well keep it rather than remove it now just to add 
it again later.

> > +        /* Check the entry_point after we've used it to generate the code
> > to avoid a race
> +         * condition where the check turned out fine, but entry_point got
> overwritten before we +         * got to use it for generating the code. */
> +        if (!body->entry_point) {
> 
> Yowser...this can happen? Reposession?

Yes, exactly.

> > -        MVMSpeshBB *ptr = linear_next;
> 
> +        MVMSpeshBB *ptr = g->entry;
>          while (ptr != NULL) {
> -            ptr->idx += 1;
> +            if (ptr != new_bb && ptr->idx > bb->idx)
> +                ptr->idx += 1;
>              ptr = ptr->linear_next;
> 
> Why was this wrong originally? I don't see it.

It assumes that the linear order of BBs is numbered monotonically and 
increasingly. But inlining may give us BBs 1, 2, 5, 6, 3, 4.

> > +                        bb->linear_next->children[1] = target_bb;
> 
> +                        bb->linear_next->succ[1] = target_bb;
> 
> Hm, wondering if `MVM_spesh_manipulate_add_successor` wouldn't be useful
> here?

Unfortunately not as MVM_spesh_manipulate_add_successor allocates the new BB 
but here we get the BB from MVM_spesh_manipulate_split_BB_at.

I addressed all other comments, mostly by amending the commits. Thanks a lot 
for the thorough review. Certainly improved the result a lot!

src/disp/program.c
```diff
@@ -815,6 +822,72 @@ static MVMuint32 value_index_how(MVMThreadContext *tc, MVMDispProgramRecording *
...
 
+/* Ensures we have a values used entry for the specified unboxed int. */
+static MVMuint32 value_index_unbox_int(MVMThreadContext *tc, MVMDispProgramRecording *rec,
+        MVMuint32 from_value) {
+    /* Look for an existing such value. */
+    MVMuint32 i;
+    for (i = 0; i < MVM_VECTOR_ELEMS(rec->values); i++) {
+        MVMDispProgramRecordingValue *v = &(rec->values[i]);
+        if (v->source == MVMDispProgramRecordingUnboxValue &&
+                v->unbox.from_value == from_value)
```
jnthn, line 833: I think this should check `v->unbox.kind == MVM_CALLSITE_ARG_INT` too; we do have allomorphs and a dispatch program that unboxes the same thing two ways is possible.

src/disp/program.c
```diff
@@ -815,6 +822,72 @@ static MVMuint32 value_index_how(MVMThreadContext *tc, MVMDispProgramRecording *
...
+
+/* Ensures we have a values used entry for the specified unboxed num. */
+static MVMuint32 value_index_unbox_num(MVMThreadContext *tc, MVMDispProgramRecording *rec,
+        MVMuint32 from_value) {
+    /* Look for an existing such value. */
+    MVMuint32 i;
+    for (i = 0; i < MVM_VECTOR_ELEMS(rec->values); i++) {
+        MVMDispProgramRecordingValue *v = &(rec->values[i]);
+        if (v->source == MVMDispProgramRecordingUnboxValue &&
+                v->unbox.from_value == from_value)
```
jnthn, line 855: `v->unbox.kind` check here also

src/disp/program.c
```diff
@@ -815,6 +822,72 @@ static MVMuint32 value_index_how(MVMThreadContext *tc, MVMDispProgramRecording *
...
+
+/* Ensures we have a values used entry for the specified unboxed str. */
+static MVMuint32 value_index_unbox_str(MVMThreadContext *tc, MVMDispProgramRecording *rec,
+        MVMuint32 from_value) {
+    /* Look for an existing such value. */
+    MVMuint32 i;
+    for (i = 0; i < MVM_VECTOR_ELEMS(rec->values); i++) {
+        MVMDispProgramRecordingValue *v = &(rec->values[i]);
+        if (v->source == MVMDispProgramRecordingUnboxValue &&
+                v->unbox.from_value == from_value)
```
jnthn, line 877: `v->unbox.kind` check here also

src/core/callstack.c
```diff
@@ -198,7 +198,6 @@ MVMCallStackFrame * MVM_callstack_allocate_frame(MVMThreadContext *tc, MVMuint32
 
     /* Current arguments callsite must be NULL as it's used in GC. Extra must
      * be NULL so we know we don't have it. Flags should be zeroed. */
-    allocated->frame.cur_args_callsite = NULL;
```
jnthn, line 201: Comment is out of date after this code change.

jnthn (commented):
A few final things I spotted...

niner:
> I think this should check `v->unbox.kind == MVM_CALLSITE_ARG_INT` too; we do
> have allomorphs and a dispatch program that unboxes the same thing two ways
> is possible.

Good point! I amended the three commits.

> Comment is out of date after this code change.

dito

jnthn (approved):
I've nothing more to add; excellent work. :+1:

lizmat:
But we'll leave this until after the 2021.11 release, right?

jnthn:
> But we'll leave this until after the 2021.11 release, right?

I thought we were still looking for a release manager, in which case a 2021.11 this weekend looks hugely doubtful?

lizmat:
We are indeed still looking...

Do I read correctly that you'd rather have this merged asap if we don't have a release manager?

jnthn:
> Do I read correctly that you'd rather have this merged asap if we don't have a release manager?

Well, only if that's what @niner prefers too. :-)

niner:
> Well, only if that's what @niner prefers too. :-)

Honestly, I can't upgrade to a current rakudo release until NativeCall has at 
least comparable performance to before new-disp. Remember that 2021.10 is just 
about half as fast as 2021.09.

A 50 % performance loss on test-t would have been a definite release blocker. 
Actually it was even a merge blocker. new-disp wasn't even merged to the 
*development* master until test-t performance was on par with the previous 
implementation. Somehow csv-ip5xs has never gotten a similar treatment despite 
it measuring performance of rakudo's NativeCall subsystem which is e.g. at the 
heart of every Cro web application. Indeed, the 50 % performance loss was not 
even mentioned by anyone on the channel, much less considered for release 
planning.

So yes, I see good reason for merging before the next release.

lizmat:
Then I will squash merge and bump, unless someone beats me to it in the next 30 minutes.

jnthn:
> Somehow csv-ip5xs has never gotten a similar treatment despite
> it measuring performance of rakudo's NativeCall subsystem which is e.g. at the
> heart of every Cro web application.

I'm curious about that assertion; Cro is mostly pure Raku code. It was also one of the things that I measured ahead of the `new-disp`, and found a 10% *improvement* in request throughput. That test was for a non-TLS connection, and NativeCall is of course more involved if TLS is at play. If your Cro webapps are doing a lot of `Inline::Perl5` usage I can appreciate it's rather more important, but that isn't "every Cro web application".

niner:
@jnthn I mean OpenSSL for TLS support in Cro web applications, since TLS is ubiquitous nowadays.

patrickbkr:
@lizmat Can you recall why this was reverted?

lizmat:
I have no active recollection of that.

jdv:
Isn't it just https://github.com/MoarVM/MoarVM/pull/1609?  See https://irclogs.raku.org/moarvm/2021-11-18.html.

PR #1596, opened 2021-11-09
Use alloca in MVM_string_memmem_grapheme32str...

MasterDuke17:
based on how it's done in knuth_morris_pratt_string_index().

I tested this with a one-liner that just does some `.contains()` and
`.starts-with()` over a spesh log that's ~200k lines long. Timing it
(with `MVM_SPESH_BLOCKING=1`) reported ~2.36s before this change and
~2.26 after. Additionally, heaptrack reported ~4.7M calls to allocation
functions and ~1.8M temporary allocations before, and ~3.0M calls to
allocation functions and ~110k temporary allocations after.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -544,24 +544,32 @@ static MVMint64 MVM_string_memmem_grapheme32 (MVMThreadContext *tc, MVMGrapheme3
 static MVMint64 MVM_string_memmem_grapheme32str (MVMThreadContext *tc, MVMString *Haystack, MVMString *needle, MVMint64 H_start, MVMStringIndex H_graphs, MVMStringIndex n_graphs) {
     MVMGrapheme32 *needle_buf = NULL;
     MVMint64 rtrn;
+    int needle_is_malloced = 0;
     if (needle->body.storage_type != MVM_STRING_GRAPHEME_32) {
         MVMStringIndex i;
-        needle_buf = MVM_malloc(needle->body.num_graphs * sizeof(MVMGrapheme32));
+        size_t needle_size = n_graphs * sizeof(MVMGrapheme32);
+        /* Allocate max 3K onto the stack, otherwise malloc */
+        if (needle_size < 3000)
+            needle_buf = alloca(needle_size);
+        else {
+            needle_buf = MVM_malloc(needle_size);
+            needle_is_malloced = 1;
+        }
         if (needle->body.storage_type != MVM_STRING_GRAPHEME_8) {
             MVMGraphemeIter n_gi;
             MVM_string_gi_init(tc, &n_gi, needle);
-            for (i = 0; i < needle->body.num_graphs; i++) {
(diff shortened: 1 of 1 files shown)


src/strings/ops.c
```diff
@@ -544,9 +544,17 @@ static MVMint64 MVM_string_memmem_grapheme32 (MVMThreadContext *tc, MVMGrapheme3
 static MVMint64 MVM_string_memmem_grapheme32str (MVMThreadContext *tc, MVMString *Haystack, MVMString *needle, MVMint64 H_start, MVMStringIndex H_graphs, MVMStringIndex n_graphs) {
     MVMGrapheme32 *needle_buf = NULL;
     MVMint64 rtrn;
+    int needle_is_malloced = 0;
     if (needle->body.storage_type != MVM_STRING_GRAPHEME_32) {
         MVMStringIndex i;
-        needle_buf = MVM_malloc(needle->body.num_graphs * sizeof(MVMGrapheme32));
+        size_t needle_size = (1 + n_graphs) * sizeof(MVMGrapheme32);
+        /* Allocate max 8K onto the stack, otherwise malloc */
+        if (needle_size < 3000)
```
jnthn, line 552: Comment says 8K, allocation amount is 3K?

MasterDuke17: That was just copied from `knuth_morris_pratt_string_index()`, looks like the comment is wrong there too. I'll correct the comments and force push.

MasterDuke17:
Callgrind reports ~21,050,000,000 instructions before and ~20,408,000,000 instructions after.

src/strings/ops.c
```diff
@@ -544,24 +544,32 @@ static MVMint64 MVM_string_memmem_grapheme32 (MVMThreadContext *tc, MVMGrapheme3
...
         else {
-            for (i = 0; i < needle->body.num_graphs; i++) {
+            for (i = 0; i < n_graphs; i++) {
                 needle_buf[i] = needle->body.storage.blob_8[i];
             }
         }
     }
     rtrn = MVM_string_memmem_grapheme32(tc, Haystack->body.storage.blob_32, needle_buf ? needle_buf : needle->body.storage.blob_32, H_start, H_graphs, n_graphs);
-    if (needle_buf) MVM_free(needle_buf);
+    if (needle_buf && needle_is_malloced) MVM_free(needle_buf);
```
zhuomingliang, line 572: Change to `needle_is_malloced && needle_buf ` will be better. For most case, if `needle_is_malloced ` is true, it should not  test `needle_buf` again.

MasterDuke17: Actually, it doesn't even need the `&& needle_buf`, right? If `needle_is_malloced` is set, we called MVM_malloc with a size >= 3000. If that were to fail it'd blow up earlier. Ah, just looked and `knuth_morris_pratt_string_index` just does the `*_is_malloced` test, so switching all to just that.

PR #1597, opened 2021-11-10
New disp nativecall libffi azure hack

nwc10:
Make with the CI!


niner:
I'm pretty sure this is not needed any more

PR #1598, opened 2021-11-10
getlexperinvtype_o is now inlinable

MasterDuke17:
"getlexperinvtype_o basically just calls MVM_frame_find_lexical_by_name
which in https://github.com/MoarVM/MoarVM/commit/e2e63d08342eb60f518ed5edf143b477626de81f
was converted to using the frame walker which correctly handles inlining. So
the reason for marking this op :noinline might actually be gone."
                                             - nine, #moarvm, 2021/11/10 18:39:19


lizmat:
If it tests clean, I'm inclined to merge this with the other two by @MasterDuke17 and bump MoarVM

PR #1600, opened 2021-11-13
Fix MVM_string_ascii_encode_substr to use given offset value

MasterDuke17:
Since 6eaf6aad3393af33023e0e8346114bb7e4a27b37, any value passed in
`start` is not used to offset into the string.

### DIFF
--- a/src/strings/ascii.c
+++ b/src/strings/ascii.c
@@ -161,7 +161,7 @@ char * MVM_string_ascii_encode_substr(MVMThreadContext *tc, MVMString *str, MVMu
     result = MVM_malloc(result_alloc + 1);
     if (str->body.storage_type == MVM_STRING_GRAPHEME_ASCII) {
         /* No encoding needed; directly copy. */
-        memcpy(result, str->body.storage.blob_ascii, lengthu);
+        memcpy(result, str->body.storage.blob_ascii + start, lengthu);
         result[lengthu] = 0;
         if (output_size)
             *output_size = lengthu;
@@ -170,6 +170,10 @@ char * MVM_string_ascii_encode_substr(MVMThreadContext *tc, MVMString *str, MVMu
         MVMuint32 i = 0;
         MVMCodepointIter ci;
         MVM_string_ci_init(tc, &ci, str, translate_newlines, 0);
+        while (MVM_string_ci_has_more(tc, &ci) && start > 0) {
+            MVM_string_ci_get_codepoint(tc, &ci);
+            start--;
+        }
         while (MVM_string_ci_has_more(tc, &ci)) {
             MVMCodepoint ord = MVM_string_ci_get_codepoint(tc, &ci);
             if (i == result_alloc) {


timo (approved):
LGTM

jnthn:
Hm, is it me or does only the fast path care about `lengthu` and the one with the codepoint iterator ignores it?

MasterDuke17:
> Hm, is it me or does only the fast path care about `lengthu` and the one with the codepoint iterator ignores it?

I just tested with this further change:
```patch
diff --git a/src/strings/ascii.c b/src/strings/ascii.c
index a2a39326a..4fe1fa071 100644
--- a/src/strings/ascii.c
+++ b/src/strings/ascii.c
@@ -174,8 +174,9 @@ char * MVM_string_ascii_encode_substr(MVMThreadContext *tc, MVMString *str, MVMu
             MVM_string_ci_get_codepoint(tc, &ci);
             start--;
         }
-        while (MVM_string_ci_has_more(tc, &ci)) {
+        while (MVM_string_ci_has_more(tc, &ci) && lengthu > 0) {
             MVMCodepoint ord = MVM_string_ci_get_codepoint(tc, &ci);
+            lengthu--;
             if (i == result_alloc) {
                 result_alloc += 8;
                 result = MVM_realloc(result, result_alloc + 1);
```
and get six fails in t/spec/S15-nfg/crlf-encoding.t.
```
not ok 9 - CRLF encodes to 2 bytes (ascii)
# Failed test 'CRLF encodes to 2 bytes (ascii)'
# at t/spec/S15-nfg/crlf-encoding.t line 10
# expected: '2'
#      got: '1'
```
```
not ok 11 - second byte is correct (ascii)
# Failed test 'second byte is correct (ascii)'
# at t/spec/S15-nfg/crlf-encoding.t line 12
# expected: '10'
#      got: (Failure)
```
```
not ok 13 - decoding it gives back the correct grapheme (ascii)
# Failed test 'decoding it gives back the correct grapheme (ascii)'
# at t/spec/S15-nfg/crlf-encoding.t line 16
# expected: "\r\n"
#      got: "\r"
not ok 14 - Wrote file of correct length (ascii)
# Failed test 'Wrote file of correct length (ascii)'
# at t/spec/S15-nfg/crlf-encoding.t line 22
# expected: '12'
#      got: '10'
not ok 15 - Read file and got expected number of chars (ascii)
# Failed test 'Read file and got expected number of chars (ascii)'
# at t/spec/S15-nfg/crlf-encoding.t line 27
# expected: '10'
#      got: '9'
not ok 16 - Chars read from file were correct (ascii)
# Failed test 'Chars read from file were correct (ascii)'
# at t/spec/S15-nfg/crlf-encoding.t line 28
# expected: "goat\r\nboat\r\n"
#      got: "goat\r\nboat"
```

jnthn:
@MasterDuke17 Indeed, because the units of `lengthu` are graphemes, but we iterate codepoints. It seems that actually none of the other encoders really pay attention to `length` either, though, which makes me wonder why we have it...

MasterDuke17:
> @MasterDuke17 Indeed, because the units of `lengthu` are graphemes, but we iterate codepoints. It seems that actually none of the other encoders really pay attention to `length` either, though, which makes me wonder why we have it...

Yeah, not sure how far down this rabbit hole I want to go, but just pushed a potential fix for the grapheme vs codepoint issue (t/spec/S15-nfg/crlf-encoding.t passes, couldn't do a full spectest because I am on battery).

nwc10:
`MVM_string_utf8_encode_substr` also fails to implement `start` and `length` - it validates that they are in range, but otherwise ignores them.

It's been this way for 7 years, which makes me think that we should simplify the APIs we offer, instead of fixing them. @jnthn - if we want to do that, how do we go about it? What actually depends on the C level APIs marked `MVM_PUBLIC`?

PR #1604, opened 2021-11-16
Optimize passing CPointer to nativecall ever so slightly

timo:
We emit an unbox_i operation in a compiled nativecall body in order to
get the pointer value to pass to the native function. Without a spesh
method on the CPointer repr, this would interpret as, and jit into, a
call to CPointer's get_int.

Instead of a call, we just emit a spesh op to do the memory offset and
read for us, which the JIT also likes very much.

### DIFF
--- a/src/6model/reprs/CPointer.c
+++ b/src/6model/reprs/CPointer.c
@@ -117,6 +117,36 @@ const MVMREPROps * MVMCPointer_initialize(MVMThreadContext *tc) {
     return &CPointer_this_repr;
 }
 
+static void spesh(MVMThreadContext *tc, MVMSTable *st, MVMSpeshGraph *g, MVMSpeshBB *bb, MVMSpeshIns *ins) {
+    MVMuint16             opcode    = ins->info->opcode;
+    switch (opcode) {
+    case MVM_OP_unbox_i: {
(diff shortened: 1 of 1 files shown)


src/6model/reprs/CPointer.c
```diff
@@ -117,6 +117,36 @@ const MVMREPROps * MVMCPointer_initialize(MVMThreadContext *tc) {
...
+            ins->operands[2].lit_i16 = offsetof(MVMCPointer, body.ptr);
+            MVM_spesh_graph_add_comment(tc, g, ins, "unbox_i into %s", ins->info->name);
+        }
+        else {
+            MVM_spesh_graph_add_comment(tc, g, ins, "no devirt: not known concrete");
+        }
+        break;
+    }
+    default:
+        MVM_spesh_graph_add_comment(tc, g, ins, "reprop %s unsupported in P6Opaque %s",
```
niner, line 144: Should probably say "in CPointer"

niner (changes requested):

niner (approved):

coke:
This should not have been merged so close to the release.

coke:
Reverted via linked PR - please feel free to create another PR to RE-apply this, which we can do after the release today.

PR #1605, opened 2021-11-16
Mark nqp::radix as :pure

MasterDuke17:
I suspect this was just overlooked in
ed9db7251d18d15bbf645e0a8b6cdb2654a32ece (where nqp::radix_I *was* marked
:pure).


lizmat:
Apart from correctness, would this make a performance difference somehow?

MasterDuke17:
> Apart from correctness, would this make a performance difference somehow?

It could, though [this](https://github.com/MoarVM/MoarVM/blob/master/src/spesh/dead_ins_elimination.c#L21-L30) is the only one obvious from a quick grep.

jnthn (approved):

PR #1608, opened 2021-11-18
Convert temp allocations into alloca

MasterDuke17:
These were all pointed out by heaptrack as sources of temporary
allocations during a Rakudo compile of CORE.c.setting.

Before heaptrack reported 29,995,046 calls to allocation functions and
2,600,772 temporary allocations. After it reports 29,526,794 calls to
allocation functions and 2,442,004 temporary allocations.

Most of these probably don't impact performance all that much, but the
change to parse_int_frac_exp does make a small-but-noticeably-positive
difference when coercing a native str to a native num. `my str $a =
100000.rand.Str; my num $b; $b = $a for ^10_000_000; say now - INIT now;
say $b` reports ~2.17s before and ~2s after.

### DIFF
--- a/src/6model/serialization.c
+++ b/src/6model/serialization.c
@@ -530,7 +530,7 @@ static void write_hash_str_var(MVMThreadContext *tc, MVMSerializationWriter *wri
         MVMint32 is_malloced = 0;
         MVMuint32 keys_size = sizeof(MVMString *) * elems;
         MVMString **keys;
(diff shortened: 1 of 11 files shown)


MasterDuke17:
I'll have to revisit the heaptrack reports now that we've removed the FSA. But the impact on `parse_int_frac_exp` is still noticeable, going from ~2.12s to ~2.08s.

PR #1611, opened 2021-11-25
Fix various callstack region record related failures

niner:
In several places we could encounter a callstack region record while expecting something else like a frame record.
These issues were uncovered by setting MVM_CALLSTACK_DEFAULT_REGION_SIZE to 72 (the lowest legal value without breaking valid code).
With these fixes, not even forcing allocation of a new region on every record allocation unearths any new failures (besides excessive memory usage).

### DIFF
--- a/src/core/callstack.c
+++ b/src/core/callstack.c
@@ -886,6 +886,13 @@ void MVM_callstack_unwind_failed_dispatch_run(MVMThreadContext *tc) {
     move_to_prev_record(tc);
 }
 
+void MVM_callstack_unwind_nested_runloop(MVMThreadContext *tc) {
+    assert(tc->stack_top->kind == MVM_CALLSTACK_RECORD_NESTED_RUNLOOP);
+    tc->stack_current_region->alloc = (char *)tc->stack_top;
+    tc->stack_top = tc->stack_top->prev;
+    unwind_region_start_or_flattening(tc);
+}
+
 /* Walk the linked list of records and mark each of them. */
 #define add_collectable(tc, worklist, snapshot, col, desc) \
     do { \
--- a/src/core/callstack.c
+++ b/src/core/callstack.c
@@ -551,6 +551,10 @@ static void unwind_region_start_or_flattening(MVMThreadContext *tc) {
         tc->stack_top = tc->stack_top->prev;
     }
 }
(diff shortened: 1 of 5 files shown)


niner:
That one CI failure seems unrelated. It's 15-gh_1202.t again.

src/core/callstack.c
```diff
@@ -503,19 +503,22 @@ void MVM_callstack_continuation_append(MVMThreadContext *tc, MVMCallStackRegion
 
 /* Walk the frames in the region, looking for the first bytecode one. */
 MVMFrame * MVM_callstack_first_frame_in_region(MVMThreadContext *tc, MVMCallStackRegion *region) {
```
jnthn, line 505: Hm, this name no longer describes what it does. I'm also a bit concerned about whether this can generate false positives now; for example, I think continuation handling uses this and relies upon the frame it finds really being in the region, otherwise it could find a root frame that isn't going to be within the continuation.

jnthn: While my comment on the name still kinda applies (maybe `first_frame_from_region` might be better), I I no longer think there's an actual bug here; I somehow misread the `->next` as if it were `->prev`.

src/core/callstack.c
```diff
@@ -631,8 +634,158 @@ static void handle_bind_control(MVMThreadContext *tc, MVMCallStackBindControl *c
...
+    MVMCallStackDispatchRecord *disp_record =
+        (MVMCallStackDispatchRecord *)tc->stack_top;
+    if (disp_record->resumption_state.disp)
+        MVM_disp_resume_destroy_resumption_state(tc, &(disp_record->resumption_state));
+    if (disp_record->produced_dp && !disp_record->produced_dp_installed)
+        MVM_disp_program_destroy(tc, disp_record->produced_dp);
+    if (disp_record->temps)
+        MVM_free(disp_record->temps);
+    tc->stack_current_region->alloc = (char *)tc->stack_top;
+    tc->stack_top = tc->stack_top->prev;
```
jnthn, line 657: And since we're factoring out to static inlines, this could also call `move_to_prev_record`

jnthn: (And the same comment applies in a few other places too)

jnthn (approved):
I don't see any correctness issues. A few optional style points.

PR #1612, opened 2021-11-27
Fix uninitialized register after PEA and inlining

niner:
The register in question should have been written to by a PHI node and carried
the appropriate deopt usage markers. The other side of the PHI (just one
version) didn't. So PEA optimized away the writer.

Fix by copying the dopt_users from the PHIs target to the PHI's sources when
rewriting returns. This way PEA knows that it has to keep the writer alive.

I don't know if this is an appropriate fix for this issue. An alternative could be to
teach PEA to look through PHI nodes. But PEA is a lot more complicated and I'm
not sure enough that it's worth the time spent in the learning curve to fix it there.


jnthn (approved):
The fix makes sense to me: we'd have placed deopt usages on both if it occurred "naturally", so this is making sure it happens in the synthetic case also. My only suggestion would be to comment what we're doing here in the code, or better yet to extract a function doing this and place it in `usages.[ch]`.

niner:
I extracted a function like suggested and pushed the fix to master. Unfortunately I only re-read your comment after pushing and missed that the function should live in usages.c :/

PR #1618, opened 2021-12-09
Use faster itoa implementation

MasterDuke17:
On my system this version consistently was faster than the existing one
when running the benchmark from https://github.com/miloyip/itoa-benchmark

This was also still the case in a trivial benchmark of NQP and Raku
code. For example,
`MVM_SPESH_BLOCKING=1 ./nqp-m -e 'my str $s; my int $i := 0; my $n := nqp::time; while $i++ < 10_000_000 { $s := $i }; say(nqp::div_n(nqp::time - $n, 1000000000e0)); say($s)'`
reports ~0.56 before and ~0.43 after. Using callgrind (though just up
to 1_000_000) I see ~958,400,000 instructions before and ~873,700,000
instructions after.

### DIFF
--- a/src/core/coerce.c
+++ b/src/core/coerce.c
@@ -11,13 +11,10 @@ MVMint64 MVM_coerce_istrue_s(MVMThreadContext *tc, MVMString *str) {
 }
 
 /* ui64toa and i64toa
- * Copyright(c) 2014-2016 Milo Yip (miloyip@gmail.com)
- * https://github.com/miloyip/itoa-benchmark
- * With minor modifications.
+MIT License
 
- Copyright (C) 2014 Milo Yip
-
-MIT License:
+Copyright (c) 2017 James Edward Anhalt III - https://github.com/jeaiii/itoa
+with minor modifications for formatting and so it will compile as C code
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


nwc10:
I built it on ppc64 (so big endian) and arm (ie 32 bit CPU) and everything still works. So there weren't any bad assumptions about x86_64 hidden in that code. (There might be bad assumptions about Debian in the code - I've heard rumours that other OSes exist, but I didn't investigate further \:-))

I've looked at the C code. I can't say that I understand it - either exactly how it works, or why it should be faster.

I did wonder whether even with static functions whether the C compiler can spot all possible optimisations. Are you in a position to easily benchmark whether changing the functions to `const char *` changes anything?

(But if that's not easy, don't worry about it. If it is, then I think it probably should be a distinct commit, as it's not strictly something needed to convert the original author's C++ code to C)

nwc10 (approved):

MasterDuke17:
> I've looked at the C code. I can't say that I understand it - either exactly how it works, or why it should be faster.

The version here https://github.com/jeaiii/itoa/blob/main/itoa/to_text_from_integer.h has some more comments and explanation. The version I copied https://github.com/jeaiii/itoa/blob/main/itoa/itoa_jeaiii.cpp (because it was closer to what was in the benchmark repo) does also, I just didn't copy them for some reason. I only tried the 1-char-at-a-time version, because there was a comment somewhere that the 2-char-at-a-time version didn't work on 32-bit and/or big endian systems. However, I don't see that comment now, maybe it was in the older version used in the benchmark repo? I don't have access to big endian (or easy access to 32-bit), if you're curious you could try these https://github.com/jeaiii/itoa/blob/main/itoa/itoa_jeaiii.cpp#L60-L83 defines instead.

> Are you in a position to easily benchmark whether changing the functions to `const char *` changes anything?

Assuming you mean the function return values, I just made that change and tried my above example with callgrind, but didn't notice any difference.

jeaiii:
2 chars at a time should work as far as a know. If there is a system where it doesn't work, I'll help make it work.

PR #1620, opened 2021-12-16
Deserialize empty hashes robustly

nwc10:
Fixes a bug whereby we wrongly hit an assertion failure if attempting to iterate an empty hash that had been serialized/deserialized. This was recently discovered because it broke the build of `LibXML`. However, the bug has been present for months, so it is totally unclear how it was only just encountered. (Other than "assertions default to disabled", and the code path beyond the assertion would still give correct results)

Also optimise the hash memory allocation when deserializing an empty hash, and the code executed when serializing it.

### DIFF
--- a/src/6model/serialization.c
+++ b/src/6model/serialization.c
@@ -523,24 +523,27 @@ static int cmp_strings(const void *s1, const void *s2) {
 }
 static void write_hash_str_var(MVMThreadContext *tc, MVMSerializationWriter *writer, MVMObject *hash) {
     MVMuint32 elems = (MVMuint32)MVM_repr_elems(tc, hash);
-    MVMString **keys = MVM_malloc(sizeof(MVMString *) * elems);
-    MVMObject *iter = MVM_iter(tc, hash);
-    MVMuint64 i = 0;
 
     /* Write out element count. */
     MVM_serialization_write_int(tc, writer, elems);
 
-    /* Write elements, as key,value,key,value etc. */
-    while (MVM_iter_istrue(tc, (MVMIter *)iter)) {
-        MVM_repr_shift_o(tc, iter);
-        keys[i++] = MVM_iterkey_s(tc, (MVMIter *)iter);
+    if (elems) {
+        MVMString **keys = MVM_malloc(sizeof(MVMString *) * elems);
+        MVMObject *iter = MVM_iter(tc, hash);
+        MVMuint64 i = 0;
+
+        /* Write elements, as key,value,key,value etc. */
+        while (MVM_iter_istrue(tc, (MVMIter *)iter)) {
+            MVM_repr_shift_o(tc, iter);
+            keys[i++] = MVM_iterkey_s(tc, (MVMIter *)iter);
+        }
+        qsort(keys, elems, sizeof(MVMString*), cmp_strings);
+        for (i = 0; i < elems; i++) {
+            MVM_serialization_write_str(tc, writer, keys[i]);
+            MVM_serialization_write_ref(tc, writer, MVM_repr_at_key_o(tc, hash, keys[i]));
+        }
+        MVM_free(keys);
     }
-    qsort(keys, elems, sizeof(MVMString*), cmp_strings);
-    for (i = 0; i < elems; i++) {
-        MVM_serialization_write_str(tc, writer, keys[i]);
-        MVM_serialization_write_ref(tc, writer, MVM_repr_at_key_o(tc, hash, keys[i]));
-    }
-    MVM_free(keys);
 }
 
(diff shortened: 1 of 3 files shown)


src/6model/serialization.c
```diff
@@ -523,24 +523,27 @@ static int cmp_strings(const void *s1, const void *s2) {
 }
 static void write_hash_str_var(MVMThreadContext *tc, MVMSerializationWriter *writer, MVMObject *hash) {
     MVMuint32 elems = (MVMuint32)MVM_repr_elems(tc, hash);
-    MVMString **keys = MVM_malloc(sizeof(MVMString *) * elems);
```
niner, line 526: This has the nice side effect of getting rid of a possible 0-sized malloc. IIRC not all systems really like those.

niner (approved):

src/6model/serialization.c
MasterDuke17: Additionally, this could now be a good spot for using alloca or the FSA, though I don’t know how hot it is or the usual size.

nwc10: > IIRC not all systems really like those.

I'm really not sure if this is the case "any more". In that, C89 says that you're allowed to `malloc(0)`, and I never met a system which was C89 non-conformant here. (SunOS `sprintf` returns `char *` and `fflush(NULL)` is a problem, but I can't remember the details of why, but basically C89 non-conformance died with SunOS, until Android made some daft choices with things like `setlocale`)

Sure, there's a comment in the Perl 5 source code about this, but the Perl 5 source code has continuous lineage back to Perl/perl5@8d063cd8450e59ea1c611a2f4f5a21059a2804f1

```
commit 8d063cd8450e59ea1c611a2f4f5a21059a2804f1 (tag: perl-1.0)
Author: Larry Wall <lwall@jpl-devvax.jpl.nasa.gov>
Date:   Fri Dec 18 00:00:00 1987 +0000

    a "replacement" for awk and sed
```

hence some of the comments can pre-date 1989, let alone whenever lagging vendors achieved C89 compliance \:-)

nwc10: > Additionally, this could now be a good spot for using alloca or the FSA, though I don’t know how hot it is or the usual size.

I thought briefly about `alloca` and suspect that it wouldn't be a good idea. We can't trap failure from `alloca` (IIRC it's actually undefined behaviour if it fails). Potentially we could have a very large hash to serialise - on a system with a relatively small stack size, `alloca` might fail (craptacularly) whilst `malloc` works, because there is plenty of heap.

In theory the FSA should always be "faster" because deallocation needs the size too, which must make it more efficient than `free`, but it's not clear that anyone has managed to meaningfully benchmark any speedup when trying the change in other places. This might be because `malloc`/`free` cause the same memory to be re-used more often, so if we just use `malloc`/`free` everywhere, a smaller working set of memory is "hot", whereas if half the code is using the FSA, then our working set is larger, and that is offsetting the small gain in CPU usage at `free` time.

MasterDuke17: > I thought briefly about `alloca` and suspect that it wouldn't be a good idea. We can't trap failure from `alloca` (IIRC it's actually undefined behaviour if it fails). Potentially we could have a very large hash to serialise - on a system with a relatively small stack size, `alloca` might fail (craptacularly) whilst `malloc` works, because there is plenty of heap.

Assuming it's not guaranteed to be a small number, we'd need the usual
```C
MVMString **keys;
int is_malloced = 0;
size_t size = sizeof(MVMString *) * elems;
if (size < whatever_max_we_choose_for_alloca) {
    keys = alloca(size);
}
else {
    keys = MVM_malloc(size);
    is_malloced = 1;
}
<...>
if (is_malloced)
    MVM_free(keys);
```

> In theory the FSA should always be "faster" because deallocation needs the size too, which must make it more efficient than `free`, but it's not clear that anyone has managed to meaningfully benchmark any speedup when trying the change in other places. This might be because `malloc`/`free` cause the same memory to be re-used more often, so if we just use `malloc`/`free` everywhere, a smaller working set of memory is "hot", whereas if half the code is using the FSA, then our working set is larger, and that is offsetting the small gain in CPU usage at `free` time.

Have you seen https://github.com/MoarVM/MoarVM/issues/1623? Turns out if we use mimalloc it might be faster to get rid of the FSA entirely.

nwc10: > Assuming it's not guaranteed to be a small number, we'd need the usual

It's not guaranteed to be small.

I don't think that the complexity is worth it.

> Have you seen #1623? Turns out if we use mimalloc it might be faster to get rid of the FSA entirely.

I was aware of it but hadn't digested it fully.

PR #1631, opened 2022-01-02
New unsigned ops

niner:
This lays the foundation of properly supporting unsigned native integers. These are all the changes of the broader work that can be merged without requiring any adjustments in NQP or Rakudo. They are split off the main branch so they will already be in place when the inevitable rebootstrap will happen. This should reduce the number of required rebootstraps.

This is mostly new unsigned variants of existing integer ops, but also the introduction of a uint_box_type and a syscall for late setting of HLL config keys (currently only supporting the uint_box_type. The syscall helps avoiding bootstrapping issues in Rakudo.


src/6model/containers.c
```diff
@@ -114,6 +114,7 @@ static const MVMContainerSpec code_pair_spec = {
     code_pair_fetch_s,
     code_pair_store,
     code_pair_store_i,
+    code_pair_store_i,
     code_pair_store_n,
```
nwc10, line 118: It's not clear from this commit why we can get away with using `code_pair_store_i` (ie no need for `code_pair_store_u`) and similarly no need for for `value_desc_cont_store_u` when we *do* need `native_ref_store_u`

niner: Excellent question! The short answer is: we don't.

The slightly longer one is that the tests we have do let us get away with this, but we really shouldn't. I was thinking that I'll leave this alone for now. The branches are already huge and I'd like to see them debugged and merged first before I (or someone else) start attacking the remaining uint issues. The compiler does warn about this particular sloppiness, so it cannot fall through the cracks anyway.

Of course, if as a reviewer you disagree, I can change my strategy.

nwc10: > The compiler does warn about this particular sloppiness, so it cannot fall through the cracks anyway.

For the version of gcc that I'm using (with some or other flags) it doesn't warn for me.

I guess I'd be happier simply with a comment near line 116 roughly as you wrote. Maybe something like

*FIXME - to address all the uint issues we also need code_pair_store_u and value_desc_cont_store_u. However, our current tests don't expose this, and we don't need them for the rebootstrapping. Hence leave this code (and subtle bugs) as-is until we have the large and intertangled changes done and stable, to reduce the size and risk of those changes.*

Also, if you're rebasing the branch again, I think I'd prefer it if the one "word" commit messages such as `MVM_repr_bind_pos_u` became `Add MVM_repr_bind_pos_u`. Although thinking a bit more (but sorry, not fully baked thoughts) I *think* that it might be better to merge at least some of the commits like 92f24b80e9c4e1133ff36e893344ef28da2909f7 which merely *add* API calls, but don't change any code to call them.

(Keeping he other commits that add members to structs, and change the initialisation code in lockstep as individual commits makes sense to me, as those sort of changes are error prone, and hence having a discrete step to bisect is useful. Likewise the commits that add code to switch statements - they are also useful as individual bisects. I checked the former manually and didn't see any errors. I didn't check the latter, but "works on my machine" so I'll pretend that I did a good job of reviewing them)

src/6model/containers.c
```diff
@@ -545,6 +585,28 @@ static void native_ref_store_i(MVMThreadContext *tc, MVMObject *cont, MVMint64 v
...
+        MVM_exception_throw_adhoc(tc, "This container does not reference a native integer");
+    switch (repr_data->ref_kind) {
+        case MVM_NATIVEREF_LEX:
+            MVM_nativeref_write_lex_i(tc, cont, value);
+            break;
+        case MVM_NATIVEREF_ATTRIBUTE:
+            MVM_nativeref_write_attribute_i(tc, cont, value);
+            break;
+        case MVM_NATIVEREF_POSITIONAL:
+            MVM_nativeref_write_positional_u(tc, cont, value);
```
MasterDuke17, line 600: I'd appreciate a comment explaining why this case is `_u`, but the others are `_i`.

src/6model/reprconv.c
```diff
@@ -249,6 +262,13 @@ void MVM_repr_bind_pos_i(MVMThreadContext *tc, MVMObject *obj, MVMint64 idx, MVM
         idx, val, MVM_reg_int64);
 }
 
+void MVM_repr_bind_pos_u(MVMThreadContext *tc, MVMObject *obj, MVMint64 idx, MVMuint64 value) {
+    MVMRegister val;
+    val.i64 = value;
```
MasterDuke17, line 267: I suspect it doesn't matter that you're setting `.i64` instead of `.u64`?

src/6model/reprs/VMArray.c
```diff
@@ -1502,6 +1502,14 @@ void *MVM_VMArray_find_fast_impl_for_jit(MVMThreadContext *tc, MVMSTable *st, MV
                 return vmarray_at_pos_int64;
             }
             break;
+        case MVM_OP_atpos_u:
+            if (kind != MVM_reg_uint64) {
+                return NULL;
+            }
+            if (repr_data->slot_type == MVM_ARRAY_I64) {
```
MasterDuke17, line 1509: `MVM_ARRAY_I64`?

niner: Great catch! Looks like this is something that would just fail silently.

niner:
Thanks for the great suggestions! I squashed some commits (introducing and using some new internal API), fixed the issues pointed out and added FIXME comments for where we're still missing some _u functions. I've split the explanation for why between comments and commit messages.

src/6model/reprs/VMArray.c
```diff
@@ -227,36 +227,36 @@ void MVM_VMArray_at_pos(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, vo
                 value->n64 = (MVMnum64)body->slots.n32[body->start + real_index];
             break;
         case MVM_ARRAY_U64:
-            if (kind != MVM_reg_int64)
-                MVM_exception_throw_adhoc(tc, "MVMArray: atpos expected int register");
+            if (kind != MVM_reg_uint64 && kind != MVM_reg_int64)
+                MVM_exception_throw_adhoc(tc, "MVMArray: atpos U64 expected int register, got %d instead", kind);
```
MasterDuke17, line 231: How come this is the only one with the `kind` added to the message?

src/core/hll.c
```diff
@@ -107,6 +107,9 @@ void set_max_inline_size(MVMThreadContext *tc, MVMObject *config_hash, MVMHLLCon
             config->max_inline_size = MVM_repr_get_int(tc, size);
     });
 }
+void MVM_hll_set_config_key(MVMThreadContext *tc, MVMHLLConfig *hll, MVMString *key, MVMObject *value) {
+    hll->uint_box_type = value;
+}
```
MasterDuke17, line 112: Shouldn't this use `key` somehow?

src/6model/reprs/VMArray.c
niner: It's the only one where I needed debugging help. Since my discussion with jnthn about the topic I've started committing changes like this.

src/core/hll.c
niner: If you've ever needed any proof of my laziness, this is is as good as any :D
Yes, it should. Just wasn't required to make progress as the syscall is currently only ever used for setting uint_box_type (and not even that yet in this branch)

src/6model/reprs/VMArray.c
MasterDuke17: No objection to the addition here, it just seems out of place to only have it in the one case. But I think I still have a (yet another) WIP branch that adds context to error messages, the rest will probably get picked up whenever I get around to finished it up.

niner:
Rebased the commits and merged manually. Thanks for the reviews!

Issue #1623, opened 2021-12-19
We should consider bundling+using mimalloc (and maybe even getting rid of the FSA)


MasterDuke17:
I just ran some benchmarks of two different conditions, and each of those conditions at stock and with mimalloc 2.0.3 (https://github.com/microsoft/mimalloc) LD_PRELOADed. For each combination I ran five times and included all the numbers in the table below. These weren't done in a 100% rigorous manner, but I tried to keep the execution environment pretty similar across the various runs. For the `*_fsa_debug_*` runs I used this patch https://gist.github.com/MasterDuke17/e91e9d6d37ec09d110972d39e8c41545 to roughly see what would happen if we got rid of the FSA altogether and just relied on mimalloc. A quick tl;dr is that mimalloc pretty much beats stock everywhere, and in fact, it looks like mimalloc+fsa_debug (i.e., the FSA just punting to the regular MMV_* functions) might even be the fastest.

  | CORE.c stage parse | m-test | m-spectest | MVM_SPESH_BLOCKING=1 raku -e ''
-- | -- | -- | -- | --
stock | 40.786<br>40.499<br>40.747<br>40.696<br>40.575 | 21<br>19<br>19<br>19<br>19 | 172<br>165<br>165<br>165<br>165 | 9572921783 instructions (10 iterations)<br>1833 (100 iterations)
stock+fsa_debug | 43.452<br>44.879<br>44.019<br>44.043<br>44.212 | 22<br>19<br>19<br>20<br>20 | 179<br>173<br>175<br>173<br>174 | 9597148610 instructions (10 iterations)<br>1775 (100 iterations)
mimalloc | 39.413<br>38.799<br>38.787<br>39.159<br>38.787 | 20<br>18<br>18<br>18<br>18 | 160<br>154<br>153<br>154<br>154 | 8677884655 instructions (10 iterations)<br>1657 (100 iterations)
mimalloc+fsa_debug | 38.746<br>38.542<br>39.499<br>39.251<br>38.943 | 20<br>17<br>17<br>18<br>17 | 159<br>154<br>153<br>154<br>154 | 8604038353 instructions (10 iterations)<br>1616 (100 iterations)

jnthn:
This looks promising. If you have time and interest, a benchmark of how this does in producer/consumer situations where memory is allocated on one thread and freed on another (anything that heavily uses the `ThreadPool` probably does) could be interesting too.

[...]

PR #1634, opened 2022-01-04
Add mimalloc submodule

MasterDuke17:
See https://github.com/MoarVM/MoarVM/issues/1623 for the motivation.

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -258,6 +258,14 @@ $config{cincludes} = '' unless defined $config{cincludes};
 $config{moar_cincludes} = '' unless defined $config{moar_cincludes};
 $config{lincludes} = '' unless defined $config{lincludes};
 $config{install}   = '' unless defined $config{install};
+
+# mimalloc configuration
+$config{moar_cincludes} .= ' ' . $defaults{ccinc} . '3rdparty/mimalloc/include'
+                         . ' ' . $defaults{ccinc} . '3rdparty/mimalloc/src';
+$config{install}   .= "\t\$(MKPATH) \"\$(DESTDIR)\$(PREFIX)/include/mimalloc\"\n"
+                    . "\t\$(CP) 3rdparty/mimalloc/include/*.h \"\$(DESTDIR)\$(PREFIX)/include/mimalloc\"\n";
+push @hllincludes, 'mimalloc';
+
 if ($args{'has-libuv'}) {
     $defaults{-thirdparty}->{uv} = undef;
     unshift @{$config{usrlibs}}, 'uv';
@@ -461,7 +469,7 @@ if (not $args{static} and $config{prefix} ne '/usr') {
     push @ldflags, $config{ldrpath}             if !$args{relocatable};
 }
 push @ldflags, '-fsanitize=address'  if $args{asan};
-push @ldflags, '-fsanitize=thread'  if $args{tsan};
+push @ldflags, '-fsanitize=thread'   if $args{tsan};
 push @ldflags, $ENV{LDFLAGS}         if $ENV{LDFLAGS};
 $config{ldflags} = join ' ', @ldflags;
 
--- a/Configure.pl
+++ b/Configure.pl
@@ -398,6 +398,7 @@ $config{ldlibs} = ' -lasan ' . $config{ldlibs} if $args{asan} && $^O ne 'darwin'
 $config{ldlibs} = ' -lubsan ' . $config{ldlibs} if $args{ubsan} and $^O ne 'darwin';
 $config{ldlibs} = ' -ltsan ' . $config{ldlibs} if $args{tsan} and $^O ne 'darwin';
 $config{ldlibs} = $config{ldlibs} . ' -lzstd ' if $config{heapsnapformat} == 3;
+$config{ldlibs} = $config{ldlibs} . ' -latomic ' if $config{archname} =~ /mips/; # needed for mimalloc
 # macro defs
 $config{ccdefflags} = join ' ', map { $config{ccdef} . $_ } @{$config{defs}};
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -20,3 +20,7 @@
 	path = 3rdparty/ryu
 	url = https://github.com/MoarVM/ryu
 	ignore = untracked
+[submodule "3rdparty/mimalloc"]
+	path = 3rdparty/mimalloc
+	url = https://github.com/microsoft/mimalloc.git
+	ignore = untracked
--- a/LICENSE
+++ b/LICENSE
@@ -30,3 +30,4 @@ Unofficial summary of the intended application of the Artistic License 2.0:
 - sha1       Public Domain
 - tinymt     MIT
 - freebsd    MIT
+- mimalloc   MIT
--- a/build/Makefile.in
+++ b/build/Makefile.in
@@ -40,6 +40,7 @@ PKGCONFIGDIR = @prefix@/share/pkgconfig
 CFLAGS    = @cflags@ @ccdef@MVM_TRACING=$(TRACING) @ccdef@MVM_CGOTO=$(CGOTO)
 CINCLUDES = @moar_cincludes@ \
             @ccinc@@shaincludedir@ \
+            @ccincsystem@3rdparty/mimalloc \
             @ccincsystem@3rdparty/tinymt \
             @ccincsystem@3rdparty/ryu \
             @ccincsystem@3rdparty/dynasm \
@@ -253,6 +254,7 @@ OBJECTS2 = src/6model/reprs/MVMDLLSym@obj@ \
           src/platform/memmem32@obj@ \
           3rdparty/freebsd/memmem@obj@ \
           3rdparty/ryu/ryu/d2s@obj@ \
+          3rdparty/mimalloc/src/static@obj@ \
           src/platform/malloc_trim@obj@ \
           src/moar@obj@ \
           @platform@ \
--- a/build/setup.pm
+++ b/build/setup.pm
@@ -293,7 +293,7 @@ my %TC_MINGW32 = (
 
     libdir                   => '@bindir@',
     ccshared                 => '',
-    ldshared                 => '-shared -Wl,--out-implib,lib$(notdir $@).a',
+    ldshared                 => '-shared -Wl,--out-implib,lib$(notdir $@).a -lbcrypt',
     moarshared_norelocatable => '',
     moarshared_relocatable   => '',
     ldrpath                  => '',
@@ -479,7 +479,7 @@ our %COMPILERS = (
 my %OS_WIN32 = (
     exe      => '.exe',
     defs     => [ qw( WIN32 AO_ASSUME_WINDOWS98 ) ],
-    syslibs  => [ qw( shell32 ws2_32 mswsock rpcrt4 advapi32 psapi iphlpapi userenv user32 ) ],
+    syslibs  => [ qw( shell32 ws2_32 mswsock rpcrt4 advapi32 psapi iphlpapi userenv user32 bcrypt ) ],
     platform => '$(PLATFORM_WIN32)',
 
     translate_newline_output => 1,
--- a/src/6model/bootstrap.c
+++ b/src/6model/bootstrap.c
@@ -339,7 +339,7 @@ static void add_meta_object(MVMThreadContext *tc, MVMObject *type_obj, char *nam
         /* Set name. */
         name_str = MVM_string_ascii_decode_nt(tc, tc->instance->VMString, name);
         MVM_ASSIGN_REF(tc, &(meta_obj->header), ((MVMKnowHOWREPR *)meta_obj)->body.name, name_str);
-        type_obj->st->debug_name = strdup(name);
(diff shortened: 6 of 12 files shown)


nwc10:
(So far)

MoarVM build fails on a Raspberry Pi with

```
gcc -o moar -O2 -DNDEBUG -g3 -Wl,-rpath,"//home/nick/Sandpit/moar-O2/lib"  src/main.o -L. -lmoar
/usr/bin/ld: ./libmoar.so: undefined reference to `__atomic_store_8'
/usr/bin/ld: ./libmoar.so: undefined reference to `__atomic_load_8'
/usr/bin/ld: ./libmoar.so: undefined reference to `__atomic_compare_exchange_8'
/usr/bin/ld: ./libmoar.so: undefined reference to `__atomic_fetch_add_8'
collect2: error: ld returned 1 exit status
make: *** [Makefile:652: moar] Error 1
```

(have not investigated this yes, but as that's `gcc (Raspbian 8.3.0-6+rpi1) 8.3.0` so relatively current gcc on what I thought was a sane Debian derivative Linux, I'm a bit troubled that more esoteric systems are going to be pain. Annoyingly, 2 of the best machines in the gcc compile farm are offline as of 3 days ago)

MasterDuke17:
Huh, there are people commenting on issues (e.g., https://github.com/microsoft/mimalloc/issues/495) saying they've successfully built on Raspberry Pi, but with some build flags added. I'll see if I can figure out how to add them.

nwc10:
The (newish) big ending PPC system is one that is down. An older PPC system running CentOS:

```
In file included from 3rdparty/mimalloc/include/mimalloc-types.h:13:0,
                 from 3rdparty/mimalloc/include/mimalloc-internal.h:11,
                 from 3rdparty/mimalloc/src/static.c:17:
3rdparty/mimalloc/include/mimalloc-atomic.h:34:23: fatal error: stdatomic.h: No such file or directory
 #include <stdatomic.h>
                       ^
```

This is `gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)`

(Meanwhile Sparc linux is somewhere in NQP)

MasterDuke17:
> An older PPC system running CentOS:
> 
> ```
> In file included from 3rdparty/mimalloc/include/mimalloc-types.h:13:0,
>                  from 3rdparty/mimalloc/include/mimalloc-internal.h:11,
>                  from 3rdparty/mimalloc/src/static.c:17:
> 3rdparty/mimalloc/include/mimalloc-atomic.h:34:23: fatal error: stdatomic.h: No such file or directory
>  #include <stdatomic.h>
>                        ^
> ```
> 
> This is `gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44)`

Does the suggestion in https://github.com/cockroachdb/c-jemalloc/issues/8#issuecomment-810813668 work?

nwc10:
> Does the suggestion in [cockroachdb/c-jemalloc#8 (comment)](https://github.com/cockroachdb/c-jemalloc/issues/8#issuecomment-810813668) work?

I am not root. I can't install a newer gcc (at least not without building from source, and any dependencies from source)

(The very slow MIPS systems also fails due to lack of `stdatomic.h` - it has `gcc (Debian 4.6.3-14) 4.6.3`)

Configure.pl
```diff
@@ -398,6 +398,7 @@ sub uniq {
 $config{ldlibs} = ' -lubsan ' . $config{ldlibs} if $args{ubsan} and $^O ne 'darwin';
 $config{ldlibs} = ' -ltsan ' . $config{ldlibs} if $args{tsan} and $^O ne 'darwin';
 $config{ldlibs} = $config{ldlibs} . ' -lzstd ' if $config{heapsnapformat} == 3;
+$config{ldlibs} = $config{ldlibs} . ' -latomic ' if $config{archname} =~ /mips/; # needed for mimalloc
 # macro defs
```
nwc10, line 402: This isn't correct, as we also need this on (at least) arm32. The PPC machine capable of building 32 bit is down, so i can't check it, but it might need it too. x86 doesn't (I build `-m32` on x86_64). It also looks like Solaris needs a support library for C11 atomics...

I'd prefer to go the way of probing. I've started on https://github.com/nwc10/MoarVM/pull/new/C11-mimalloc-swings-and-roundabouts but it's getting late.

nwc10:
Also I learn

1) `<stdatomic.h>` is an **optional** C11 feature - we can't strictly rely on it ever being present. Although unlike Variable Length Arrays, I'm not aware of a compiler that has refused to ever implement it
2) mimalloc relies on a *correction* to the C11 spec.

With a version of clang which I infer is strictly conformant with the original C11 spec:

```
In file included from 3rdparty/mimalloc/src/static.c:32:
3rdparty/mimalloc/src/alloc.c:488:24: error: address argument to atomic operation must be a pointer to non-const _Atomic type ('const _Atomic(mi_threadid_t) *' invalid)
  if (mi_likely(tid == mi_atomic_load_relaxed(&segment->thread_id) && page->flags.full_aligned == 0)) {  // the thread id matches and it is not a full page, nor has aligned blocks
                       ^                      ~~~~~~~~~~~~~~~~~~~
3rdparty/mimalloc/include/mimalloc-atomic.h:47:50: note: expanded from macro 'mi_atomic_load_relaxed'
#define mi_atomic_load_relaxed(p)                mi_atomic(load_explicit)(p,mi_memory_order(relaxed))
                                                 ^                        ~
3rdparty/mimalloc/include/mimalloc-atomic.h:35:33: note: expanded from macro 'mi_atomic'
#define  mi_atomic(name)        atomic_##name
                                ^
<scratch space>:43:1: note: expanded from here
atomic_load_explicit
^
/usr/lib/llvm-7/lib/clang/7.0.1/include/stdatomic.h:135:30: note: expanded from macro 'atomic_load_explicit'
#define atomic_load_explicit __c11_atomic_load
                             ^
3rdparty/mimalloc/include/mimalloc-internal.h:161:46: note: expanded from macro 'mi_likely'
#define mi_likely(x)       __builtin_expect((x),1)
                                             ^
1 error generated.
```

Given

1) how small the code change from (regular) malloc to mimalloc is
2) how many ways there are for this to break that we've already found

I feel we have to instead do this conditionally (on good enough C11 atomics support). (I've run out of time to do this tonight. Tomorrow is another day)

However, if (also) we figure out that mimalloc is faster than the FSA, then I think that the correct speed/complexity trade off is to remove the code for the FSA, so that we effectively have

1) has good enough C11 atomics support -- fast and clean
2) doesn't have C11 atomics support -- slower and clean

and not prioritise eeking out extra speed on "Less Than Awesome" platforms, on the assumption that most of our userbase is currentish x86_64 or aarch64.

ie optimise for platforms that support mimalloc.

src/io/fileops.c
```diff
@@ -423,7 +424,7 @@ MVMString * MVM_file_readlink(MVMThreadContext *tc, MVMString *path) {
 
     MVM_free(path_s);
     result = MVM_string_utf8_c8_decode(tc, tc->instance->VMString, req.ptr, strlen(req.ptr));
-    MVM_free(req.ptr);
+    uv__free(req.ptr);
```
nwc10, line 427: The change is better than what we had, but is it strictly correct?

The documentation at http://docs.libuv.org/en/v1.x/fs.html seems to suggest that the "correct" call would be `uv_fs_req_cleanup`

however

1) The documentation seems to suggest that one ought to be calling that function after every UV call that filled in a `uv_fs_t`
2) Looking at the implementation, there's a lot of C in it that we know will always end up as simply a call to `uv__free`

so if it's "correct" to call `uv_fs_req_cleanup` it sure looks like a lot of work to be seen to be doing the right thing. So I'm not sure if anything needs to change in this commit.

MasterDuke17: Nice investigating. I suspect this isn't terribly hot code, so the extra work to be "correct" probably wouldn't hurt us too much. However, I'll leave it as is for now because I'm contemplating moving some of these not-directly-related-to-using-mimalloc changes out into their own PR(s).

MasterDuke17:
Many thanks for the testing and additional work on the probing, this is an area far outside my expertise, so any help is very much appreciated.

src/io/fileops.c
nwc10: > I'm contemplating moving some of these not-directly-related-to-using-mimalloc changes out into their own PR(s).

I like this plan. I *almost* wrote something like "it would be useful to figure out a commit message that didn't mention mimalloc directly" because I thought it would be useful to cherry-pick just this commit. But, as you can sort of see from the previous sentence, I couldn't find a good way to phrase the suggestion, let alone the commit message...

Yes, I think that the "fixes for bugs you discovered" commits (at least) belong logically at the start of the chain of commits in this branch, but they would probably be better in their own PR(s) because we're going to apply them, however we end up implementing mimalloc (or getting delayed by problems)

MasterDuke17: Just pulled a couple out, rebased, and force pushed.

nwc10:
I've created #1638 which makes mimalloc optional (and the default)

It's >90% your hard work. I didn't fiddle with git (or github extensions) to try to attribute commits to multiple people - I left things with what CLI git default to. I think that commit 672750e966013982a52dfec377616cc289b785dd is mostly your code, but an edited commit message. Commit c496e887c5b0e716ba2037325a6041465bff9e7b is half things you hit and fixed, and half similar things I found by inspection, inspired by what you'd done.

I (obviously) think that we should go with the #1638 approach instead of this PR. Hence - we can close this PR?

MasterDuke17:
Awesome! Nice catches in c496e88
<https://github.com/MoarVM/MoarVM/commit/c496e887c5b0e716ba2037325a6041465bff9e7b>,
however I'm not 100% sure that the
https://github.com/MoarVM/MoarVM/pull/1638/commits/c496e887c5b0e716ba2037325a6041465bff9e7b#diff-a3fdef47716adb53ef443f57bb5c40a93088821f0bf8da1e191d39925a32d738R264
change is correct. Since that is only ever populated with string literals
(e.g., via `MVM_telemetry_interval_start(tc, "spesh worker consuming a
log");`), and we're not dynamically overriding the allocator, I think it
needs to stay as plain `free()`.

Definitely let's close this and go with #1638
<https://github.com/MoarVM/MoarVM/pull/1638>.

MasterDuke17:
My question about the free in the telemeh code makes me think that the
strndup in the same code (
https://github.com/MoarVM/MoarVM/pull/1638/files#diff-a3fdef47716adb53ef443f57bb5c40a93088821f0bf8da1e191d39925a32d738R206)
should also be unchanged. The other uses of strdup were on MVM_* allocated
strings, but I think the telemeh stuff will still use the system allocator.

nwc10:
> My question about the free in the telemeh code makes me think that the strndup in the same code ( https://github.com/MoarVM/MoarVM/pull/1638/files#diff-a3fdef47716adb53ef443f57bb5c40a93088821f0bf8da1e191d39925a32d738R206) should also be unchanged. The other uses of strdup were on MVM_* allocated strings, but I think the telemeh stuff will still use the system allocator.

From looking at the code again I come to the same conclusion (hopefully correct) that the memory in question is only ever allocated in telemeh.c by `strndup` and only ever released by the corresponding `free`.

So at first I thought that the better choice was to leave these two unchanged.

But subsequently I think instead that, **if the above is correct** and the allocation/release pair is exclusively the `strndup`/`free` only, then it **is** actually better to change (both) to be `MVM_*`, even though it looks like code churn, because if we make this change, we eliminate the only remaining call to `free` in "our" codebase. (ie not `3rdparty`). And I think that this is good, because it makes it easier to audit that we are consistent in our memory allocation/release.

If we **don't** change `strndup`/`free` then I think we ought to annotate at least one with a comment, mentioning the other. And doing that makes the code less readable than just changing it to `MVM_*` and being done with it.

I think. It's all a bit subjective.

MasterDuke17:
> From looking at the code again I come to the same conclusion (hopefully correct) that the memory in question is only ever allocated in telemeh.c by `strndup` and only ever released by the corresponding `free`.

Ah! I missed that all the frees were for memory that came from the strndup. So yes, I 100% agree with you that we should change both and your PR is correct as-is.

MasterDuke17:
Closing in favor of https://github.com/MoarVM/MoarVM/pull/1638

niner:
On Freitag, 7. Jänner 2022 20:52:49 CET Nicholas Clark wrote:
> because if we make this change, we eliminate the only
> remaining call to `free` in "our" codebase. (ie not `3rdparty`).

I don't see how we will ever be able to do that. NativeCall will always have 
cases where we need to malloc and/or free. E.g. when we take ownership over 
memory that a library allocated. Even if that's currently not implemented, it 
will have to be at some point.

nwc10:
> I don't see how we will ever be able to do that. NativeCall will always have cases where we need to malloc and/or free. E.g. when we take ownership over memory that a library allocated. Even if that's currently not implemented, it will have to be at some point.

Whoops yes, good point.

Having slept on this, I still think on balance it would be slightly better to change the code in `telemeh.c` so that we minimise the number of "places where it's correct to use bare `malloc`/`free`. Mostly because everywhere outside of 3rdparty "ought" to be commenting about why this memory is `malloc`/`free` and not our wrappers.

PR #1635, opened 2022-01-06
Correctly free memory allocated by libuv

MasterDuke17:
Use the correct libuv function to do it. According to
http://docs.libuv.org/en/v1.x/fs.html#c.uv_fs_req_cleanup "Cleanup request.
Must be called after a request is finished to deallocate any memory libuv
might have allocated."


niner (approved):

MasterDuke17:
Forgot to mention it in the commit message, but I checked the documentation and of the functions that need `uv_fs_req_cleanup` to be called after them, `uv_fs_readlink` is the only one we use.

PR #1638, opened 2022-01-07
Add mimalloc submodule, and default to mimalloc where possible

nwc10:
Permit building with mimalloc instead of libc malloc

Implement this by statically include mimalloc as a single source. Default to using mimalloc if `Configure.pl's` probe detects that the compiler provides C11 atomics, otherwise continue to use the C library's malloc as before. Add `--no-mimalloc` and `--mimalloc` options to override the automatic behaviour.

Add CI tests for `--no-mimalloc`. The CI systems have compilers that support atomics, so they now default to mimalloc.

See #1623 for motivation and #1634 for the previous iteration

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -49,7 +49,7 @@ GetOptions(\%args, qw(
     prefix=s bindir=s libdir=s mastdir=s
     relocatable make-install asan ubsan tsan
     valgrind telemeh! dtrace show-autovect git-cache-dir=s
-    show-autovect-failed:s),
+    show-autovect-failed:s mimalloc!),
 
     'no-optimize|nooptimize' => sub { $args{optimize} = 0 },
     'no-debug|nodebug' => sub { $args{debug} = 0 },
@@ -549,6 +549,32 @@ build::probe::has_isinf_and_isnan(\%config, \%defaults);
 build::probe::unaligned_access(\%config, \%defaults);
 build::probe::ptr_size(\%config, \%defaults);
 build::probe::stdatomic(\%config, \%defaults);
+
+$config{use_mimalloc} = $args{mimalloc};
+if (!defined $config{use_mimalloc}) {
+    if ($config{has_stdatomic}) {
+        print "Defaulting to mimalloc because you have <stdatomic.h>\n";
+        $config{use_mimalloc} = 1;
+    }
+    else {
+        print "Defaulting to libc malloc because <stdatomic.h> was not found.\n";
+        $config{use_mimalloc} = 0;
+    }
+}
+
+if ($config{use_mimalloc}) {
+    $config{moar_cincludes} .= ' ' . $defaults{ccinc} . '3rdparty/mimalloc/include'
+                             . ' ' . $defaults{ccinc} . '3rdparty/mimalloc/src';
+    $config{install}   .= "\t\$(MKPATH) \"\$(DESTDIR)\$(PREFIX)/include/mimalloc\"\n"
+                       . "\t\$(CP) 3rdparty/mimalloc/include/*.h \"\$(DESTDIR)\$(PREFIX)/include/mimalloc\"\n";
+    push @hllincludes, 'mimalloc';
+    $config{mimalloc_include} = '@ccincsystem@3rdparty/mimalloc';
+    $config{mimalloc_object} = '3rdparty/mimalloc/src/static@obj@';
+}
+else {
+    $config{mimalloc_include} = "";
+    $config{mimalloc_object} = "";
+}
 
 my $archname = $Config{archname};
 if ($args{'jit'}) {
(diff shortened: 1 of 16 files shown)


MasterDuke17:
Huh, looks like the probe doesn't find stdatomic.h on Windows?

MasterDuke17:
Noticed two typos in the description of the new args in Configure.pl.

"we default is to" -> "we default to"

"Specify C<--no-mimalloc> to force use of mimalloc" -> "Specify
C<--mimalloc> to force use of mimalloc"

nwc10:
> Huh, looks like the probe doesn't find stdatomic.h on Windows?

I don't have access to Win32, so the CI run is the first time this code has encountered it. You spotted a problem faster than I did.

So, .../include/mimalloc-atomic.h

```
#if defined(__cplusplus)
// Use C++ atomics
#include <atomic>
#define  _Atomic(tp)            std::atomic<tp>
#define  mi_atomic(name)        std::atomic_##name
#define  mi_memory_order(name)  std::memory_order_##name
#elif defined(_MSC_VER)
// Use MSVC C wrapper for C11 atomics
#define  _Atomic(tp)            tp
#define  ATOMIC_VAR_INIT(x)     x
#define  mi_atomic(name)        mi_atomic_##name
#define  mi_memory_order(name)  mi_memory_order_##name
#else
// Use C11 atomics
#include <stdatomic.h>
#define  mi_atomic(name)        atomic_##name
#define  mi_memory_order(name)  memory_order_##name
#endif
```

Yet again, looks like "there are standards, and there's what MSVC provides". So I guess I need to ignore the probe if the compiler is MSVC, and just "go for it" with mimalloc.

nwc10:
I rebased this, and changed the key in `%Config` from `mimalloc` to `use_mimalloc`

I thought that this key is clearer (long term), and I thought it cleaner to have a revisionist history where it was "correct" from the first commit, rather than an extra commit to have to read and think about.

I haven't dug into the CI tests - need to go AFK.

Configure.pl
```diff
@@ -549,6 +548,37 @@ sub uniq {
 build::probe::has_isinf_and_isnan(\%config, \%defaults);
 build::probe::unaligned_access(\%config, \%defaults);
 build::probe::ptr_size(\%config, \%defaults);
+build::probe::stdatomic(\%config, \%defaults);
+
+$config{mimalloc} = $args{mimalloc};
```
MasterDuke17, line 553: Looks like a mixture of `mimalloc` and `use_mimalloc` might be the problem.

MasterDuke17:
> I rebased this, and changed the key in %Config from mimalloc to use_mimalloc

I don't care strongly either way, but to play devil's advocate, we have existing `dtrace` and `telemeh` flags (i.e., not `use_dtrace` and `use_telemeh`).

src/core/alloc.h
```diff
@@ -26,7 +38,11 @@ MVM_STATIC_INLINE void * MVM_realloc(void *p, size_t size) {
 }
 
 MVM_STATIC_INLINE void * MVM_recalloc(void *p, size_t old_size, size_t size) {
+#ifdef MVM_USE_MIMALLOC
+    void *ptr = mi_realloc(p, size);
```
MasterDuke17, line 42: There is an `mi_recalloc` https://github.com/microsoft/mimalloc/blob/master/include/mimalloc.h#L227, but it expects a new count and size (like regular calloc), not old size and new size. There are only a couple uses of `MVM_recalloc`, maybe a followup commit/PR could switch `MVM_recalloc` to use `mi_recalloc` and fix all the uses.

Configure.pl
```diff
@@ -549,6 +548,37 @@ sub uniq {
 build::probe::has_isinf_and_isnan(\%config, \%defaults);
 build::probe::unaligned_access(\%config, \%defaults);
 build::probe::ptr_size(\%config, \%defaults);
+build::probe::stdatomic(\%config, \%defaults);
+
+$config{use_mimalloc} = $args{mimalloc};
+if (!defined $config{mimalloc}) {
```
MasterDuke17, line 554: Whoops, like one was missed.

src/core/alloc.h
MasterDuke17: https://gist.github.com/MasterDuke17/24f85769c96d78166138d0c18c792020 has a diff against master that I believe changes all the uses of `MVM_recalloc` and just needs a tweak in the implementation of `MVM_recalloc` to use `mi_recalloc`.

Configure.pl
nwc10: Bother yes, well spotted. Clearly the order in https://colabti.org/irclogger/irclogger_log/moarvm?date=2022-01-09#l62 was wrong. I guess next time it ought be

1. make coffee
2. re-read commit again
3. only then (if good) push

MasterDuke17 (approved):

nwc10:
> I don't care strongly either way, but to play devil's advocate, we have existing `dtrace` and `telemeh` flags (i.e., not `use_dtrace` and `use_telemeh`).

Or `--use-dtrace` ...

The context was that I was working further on what became #1640 and realised that in `build/config.h.in` (nearly) everything was consistent *verb_noun*, where *verb* was pretty descriptive - basically *can* or *has*. So "mimalloc" is "yes, right, what about it?" We have it? (always true - we ship it) We can use it? (no, not necessarily true if we could but chose to disable it) So this configuration variable was "we have enabled it" hence "use". Readability seemed important here.

I didn't change the comand-line flag - which (somewhat) then increases inconsistency. Except that it would want to be `--use-mimalloc` not `--use_mimalloc` and I don't think that our configuration keys or (C macros) can use `-` in place of `_`. But I somewhat feel that the command line flags are already a lost cause.

MasterDuke17:
> I didn't change the comand-line flag - which (somewhat) then increases inconsistency. Except that it would want to be `--use-mimalloc` not `--use_mimalloc` and I don't think that our configuration keys or (C macros) can use `-` in place of `_`. But I somewhat feel that the command line flags are already a lost cause.

I wouldn't object to changing `--dtrace`, `--telemeh`, and `--mimalloc` to `--use-*`.

PR #1639, opened 2022-01-09
More preparations for fixing unsigned handling

niner:
This PR includes the final round of fixes for unsigned handling that can be split off from the main branch. These commits are independent of changes in the other layers, i.e. they are still compatible with the master branches of the other repositories.

The fixes are mostly about using unsigned data types when handling unsigned data in MAST code, a bit of spesh and preparation for giving unsigneds their own objprimspec (10).


src/spesh/args.c
```diff
@@ -157,6 +157,12 @@ static void insert_getarg_and_box(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpe
                 hlltype_op = MVM_OP_hllboxtype_i;
                 fetch_op = MVM_OP_sp_getarg_i;
                 break;
+            case MVM_CALLSITE_ARG_UINT:
+                unboxed_temp = MVM_spesh_manipulate_get_temp_reg(tc, g, MVM_reg_int64);
```
MasterDuke17, line 161: This doesn't need to be `MVM_reg_uint64`?

nwc10:
Works on "my" machine.

src/spesh/args.c
niner: You've got an eye for these things! Fixed it in the commit

MasterDuke17 (approved):

PR #1643, opened 2022-01-13
Allocations for blocks of MVM registers needs to be 8 byte aligned

nwc10:
Various MVM_callstack_allocate_* functions allocate a single block of memory, where the last section is used as an array of MVMRegister structs.  This section needs to be 8 byte aligned, as on some platforms the (real) CPU's 64 bit loads from it need to be 8 byte aligned.

Most of these were correct already, but building on sparc32 reveals a few that were not.

Also in `MVM_callstack_allocate_dispatch_record`, the allocation only needs to be sizeof(MVMCallStackDispatchRun), as this is what is used, not sizeof(MVMCallStackDispatchRecord). This hasn't been noticed previously, because the latter struct is twice the size, so the bug simply resulting in a bit of extra memory being allocated.

These changes, along with the C11 atomics and a bug fix to `tommath.h` are sufficient to get Rakudo building and passing tests on sparc32 (apart from nativecall, for which dynlib "crashes and burns" without a valid stack trace), and `libffi` wasn't available locally). Hopefully (also) this means that with this our code is sufficiently precise on endianness and alignment that it will build on any CPU architecture, meaning that the only porting blockers are our third party libraries.

### DIFF
--- a/src/core/callstack.c
+++ b/src/core/callstack.c
@@ -207,11 +207,12 @@ MVMCallStackFrame * MVM_callstack_allocate_frame(MVMThreadContext *tc, MVMuint32
 MVMCallStackHeapFrame * MVM_callstack_allocate_heap_frame(MVMThreadContext *tc,
         MVMuint32 work_size) {
     MVMFrame *frame = MVM_gc_allocate_frame(tc);
+    size_t frame_size_aligned = to_8_bytes(sizeof(MVMCallStackHeapFrame));
     tc->stack_top = allocate_record(tc, MVM_CALLSTACK_RECORD_HEAP_FRAME,
-            sizeof(MVMCallStackHeapFrame) + work_size);
+            frame_size_aligned + work_size);
     MVMCallStackHeapFrame *allocated = (MVMCallStackHeapFrame *)tc->stack_top;
     allocated->frame = frame;
-    frame->work = (MVMRegister *)((char *)allocated + sizeof(MVMCallStackHeapFrame));
+    frame->work = (MVMRegister *)((char *)allocated + frame_size_aligned);
     frame->allocd_work = work_size;
     return allocated;
 }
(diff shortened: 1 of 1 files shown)


src/core/callstack.c
```diff
@@ -326,7 +328,7 @@ MVMCallStackFlattening * MVM_callstack_allocate_flattening(MVMThreadContext *tc,
     /* Allocate. */
     size_t record_size = to_8_bytes(sizeof(MVMCallStackFlattening));
     size_t flags_size = to_8_bytes(num_args);
-    size_t nameds_size = (num_args - num_pos) * sizeof(MVMString *);
+    size_t nameds_size = to_8_bytes((num_args - num_pos) * sizeof(MVMString *));
```
niner, line 331: So we need alignment stricter than pointer size in some cases?

niner (approved):

src/core/callstack.c
nwc10: > So we need alignment stricter than pointer size in some cases?

We always did - `double` is 8 bytes, and on (at least) some platforms it is (or should be) 8 byte aligned. It's just that

- sparc seems to be really uppity about alignment - the CPU doesn't do misaligned reads, and the kernel chooses not to "fake it" with a CPU exception handler to emulate the read (slowly) in software. (which some other CPUs do)
- sparc is also a 32 bit *ABI* on a 64 bit CPU (not a 64 bit CPU emulating a 32 bit CPU), so the compiler knows
  * that it has 64 bit instructions if it needs them
  * if some other member in a `union` or `struct` means that that `struct` or `union` is 8-byte aligned, then it's free to use the 64 bit instructions to load or store members that naturally would only be 4-byte aligned, because it "knows' that they have to actually have 8-byte alignment. (other CPUs might also do this, but their misaligned read handling hides this)

PR #1645, opened 2022-01-14
Remove the fixed size allocator (FSA)

MasterDuke17:
With the recent change to make using the mimalloc allocator the default,
it was discovered that it does such a good job that the FSA doesn't
provide any extra benefit. Therefore we're removing it to simplify
things in general. While this will be a performance regression for the
few systems that can't use mimalloc, but those are very uncommon anyway.

Because the FSA had two jobs (fast allocations/frees for small amounts
of memory + an "at safepoint" mechanism to allow concurrent
modifications of things are aren't supposed to modifed concurrently to
be detected and an error thrown instead of just segfaulting) and
mimalloc only replaces the first, the `*_at_safepoint()` functions of
the FSA had to be implemented in terms of the normal alloc/free
capabilities. That code was essentially copied from the FSA's
implementations.


src/core/alloc.h
```diff
@@ -68,12 +68,51 @@ MVM_STATIC_INLINE void MVM_free(void *p) {
...
+    do {
+        orig = tc->instance->free_at_safepoint;
+        to_add->next = orig;
+    } while (!MVM_trycas(&(tc->instance->free_at_safepoint), orig, to_add));
 }
 
+MVM_STATIC_INLINE void * MVM_realloc_at_safepoint(MVMThreadContext *tc, void *p, size_t old_bytes, size_t new_bytes) {
+    void *allocd;
+#ifdef MVM_USE_MIMALLOC
+    allocd = mi_expand(p, new_bytes);
```
niner, line 93: A comment pointing out that mi_expand tries to expand in-place and therefore is safe to do right away (and not at a safepoint) would be nice.

niner (approved):
Excellent work!

src/6model/reprs/MVMStaticFrameSpesh.c
```diff
@@ -41,6 +41,7 @@ static void gc_free(MVMThreadContext *tc, MVMObject *obj) {
     MVM_spesh_stats_destroy(tc, sfs->body.spesh_stats);
     MVM_free(sfs->body.spesh_stats);
     MVM_spesh_arg_guard_destroy(tc, sfs->body.spesh_arg_guard, 0);
+    MVM_free(sfs->body.spesh_candidates);
```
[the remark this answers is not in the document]
MasterDuke17: I believe so. On master they're only ever freed when a new one is added to an existing list and the old list is freed.

src/core/alloc.h
MasterDuke17: Done, rebased (plus some typos fixed in the commit message), and force pushed.

niner (approved):

PR #1647, opened 2022-01-16
Fix handling of native unsigned integers

niner:
This is the final round of unsigned fixes that gets us into a stable situation again. It switches over to treating unsigneds as such by giving them their own objprimspec (10). This requires appropriate changes in NQP and Rakudo contained in similarily named branches in their respective repos.

### DIFF
--- a/src/6model/containers.c
+++ b/src/6model/containers.c
@@ -469,8 +469,8 @@ static void native_ref_fetch_i(MVMThreadContext *tc, MVMObject *cont, MVMRegiste
 
 static void native_ref_fetch_u(MVMThreadContext *tc, MVMObject *cont, MVMRegister *res) {
     MVMNativeRefREPRData *repr_data = (MVMNativeRefREPRData *)STABLE(cont)->REPR_data;
-    if (repr_data->primitive_type != MVM_STORAGE_SPEC_BP_INT)
-        MVM_exception_throw_adhoc(tc, "This container does not reference a native integer");
+    if (repr_data->primitive_type != MVM_STORAGE_SPEC_BP_UINT64)
+        MVM_exception_throw_adhoc(tc, "This container does not reference a native unsigned integer");
     switch (repr_data->ref_kind) {
         case MVM_NATIVEREF_LEX:
             res->u64 = MVM_nativeref_read_lex_i(tc, cont); /* covers unsigned as well */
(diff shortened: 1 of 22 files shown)


src/6model/containers.c
```diff
@@ -469,8 +469,8 @@ static void native_ref_fetch_i(MVMThreadContext *tc, MVMObject *cont, MVMRegiste
 
 static void native_ref_fetch_u(MVMThreadContext *tc, MVMObject *cont, MVMRegister *res) {
     MVMNativeRefREPRData *repr_data = (MVMNativeRefREPRData *)STABLE(cont)->REPR_data;
-    if (repr_data->primitive_type != MVM_STORAGE_SPEC_BP_INT)
-        MVM_exception_throw_adhoc(tc, "This container does not reference a native integer");
+    if (repr_data->primitive_type != MVM_STORAGE_SPEC_BP_UINT64)
```
MasterDuke17, line 472: Just a general question (i.e., not specifically related to this line), but why is it `MVM_STORAGE_SPEC_BP_UINT64` instead of `MVM_STORAGE_SPEC_BP_UINT`?

src/core/args.c
```diff
@@ -376,6 +380,10 @@ static MVMObject * decont_arg(MVMThreadContext *tc, MVMObject *arg) {
                             MVM_exception_throw_adhoc(tc, "Expected native int argument, but got num"); \
                         case MVM_CALLSITE_ARG_STR: \
                             MVM_exception_throw_adhoc(tc, "Expected native int argument, but got str"); \
+                        case MVM_CALLSITE_ARG_INT: \
+                        case MVM_CALLSITE_ARG_UINT: \
+                            /* Ignore signedness mismatch to facilitate rebootstrapping */ \
```
MasterDuke17, line 385: Should something different be done here after this and the NQP+Rakudo PRs are merged?

src/6model/containers.c
niner: I actually don't know. I think the better question is actually "why BP_INT and not BP_INT64"? I guess those constants have grown historically. There are also constants for other sizes both for unsigned and also for signed integers. BP_INT is the only of these constants without an explicit size.

src/core/args.c
niner: Yes, once we're up and running again, MoarVM should become more picky about these mismatches.

src/6model/containers.c
MasterDuke17: Yeah, maybe we should change all the `MVM_STORAGE_SPEC_BP_INT` (i.e., the ones without a size) to `MVM_STORAGE_SPEC_BP_INT64` after this is merged.

PR #1650, opened 2022-01-19
Use libc allocator functions for NativeCall

MasterDuke17:
Because the memory could be passed to a module and freed there, we can't
use the `MVM_*` functions because those could be using a different
allocator (e.g., mimalloc as is the new default build configuration). In
cases where we can't specifiy the function to use, we instead explicitly
allocate another chunk of memory with libc functions, copy what we
originally got into it, then free the original version.

### DIFF
--- a/src/core/alloc.h
+++ b/src/core/alloc.h
@@ -37,23 +37,6 @@ MVM_STATIC_INLINE void * MVM_realloc(void *p, size_t size) {
     return ptr;
 }
 
-/* This doesn't use any mimalloc functions because it's for use in NativeCall
- * code, where modules may be freeing or allocating memory and wouldn't be 
- * using mimalloc. */
-MVM_STATIC_INLINE void * recalloc(void *p, size_t old_size, size_t size) {
-    void *ptr = realloc(p, size);
-
-    if (size > 0) {
-        if (!ptr)
-            MVM_panic_allocation_failed(size);
-
-        if (size > old_size)
-            memset((char *)ptr + old_size, 0, size - old_size);
-    }
-
-    return ptr;
-}
-
 MVM_STATIC_INLINE void * MVM_recalloc(void *p, size_t old_size, size_t size) {
 #ifdef MVM_USE_MIMALLOC
     void *ptr = mi_realloc(p, size);
--- a/src/core/alloc.h
+++ b/src/core/alloc.h
@@ -37,6 +37,23 @@ MVM_STATIC_INLINE void * MVM_realloc(void *p, size_t size) {
     return ptr;
 }
 
+/* This doesn't use any mimalloc functions because it's for use in NativeCall
+ * code, where modules may be freeing or allocating memory and wouldn't be 
+ * using mimalloc. */
+MVM_STATIC_INLINE void * recalloc(void *p, size_t old_size, size_t size) {
+    void *ptr = realloc(p, size);
+
+    if (size > 0) {
+        if (!ptr)
+            MVM_panic_allocation_failed(size);
+
+        if (size > old_size)
+            memset((char *)ptr + old_size, 0, size - old_size);
+    }
+
+    return ptr;
+}
+
 MVM_STATIC_INLINE void * MVM_recalloc(void *p, size_t old_size, size_t size) {
 #ifdef MVM_USE_MIMALLOC
     void *ptr = mi_realloc(p, size);
--- a/src/core/nativecall.c
+++ b/src/core/nativecall.c
@@ -293,19 +293,19 @@ char * MVM_nativecall_unmarshal_string(MVMThreadContext *tc, MVMObject *value, M
                 str = MVM_string_ascii_encode_malloc(tc, value_str);
                 break;
             case MVM_NATIVECALL_ARG_UTF16STR:
-                str = MVM_string_utf16_encode(tc, value_str, 0);
-#ifdef MVM_USE_MIMALLOC
-                {
-                    // strlen is wrong for UTF-16. Clearly we don't test
-                    // this. But how do we find the length?
-                    size_t str_len = strlen(str) + 1;
-                    char *libc_str = malloc(str_len);
-                    memcpy(libc_str, str, str_len);
-                    MVM_free(str);
-                    str = libc_str;
-                }
-#endif
+            {
+                MVMuint64 output_size;
+                /* This adds a (two byte) U-0000 at the end, but doesn't include
+                 * that code point in the returned output_size. */
+                char *temp = MVM_string_utf16_encode_substr(tc, value_str, &output_size, 0, -1, NULL, 0);
+                /* output_size is in 2 byte units, and doesn't include the
+                 * terminating U-0000, which was written. */
+                size_t octet_len = 2 * (output_size + 1);
+                str = malloc(octet_len);
+                memcpy(str, temp, octet_len);
+                MVM_free(temp);
                 break;
+            }
             default:
                 str = MVM_string_utf8_encode_C_string_malloc(tc, value_str);
         }
--- a/src/core/nativecall.c
+++ b/src/core/nativecall.c
@@ -291,20 +291,34 @@ char * MVM_nativecall_unmarshal_string(MVMThreadContext *tc, MVMObject *value, M
         switch (type & MVM_NATIVECALL_ARG_TYPE_MASK) {
             case MVM_NATIVECALL_ARG_ASCIISTR:
                 str = MVM_string_ascii_encode_any(tc, value_str);
+#ifdef MVM_USE_MIMALLOC
+                {
+                    size_t str_len = strlen(str) + 1;
+                    char *libc_str = malloc(str_len);
+                    memcpy(libc_str, str, str_len);
+                    MVM_free(str);
+                    str = libc_str;
+                }
+#endif
                 break;
             case MVM_NATIVECALL_ARG_UTF16STR:
                 str = MVM_string_utf16_encode(tc, value_str, 0);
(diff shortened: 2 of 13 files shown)


MasterDuke17:
This passes NQP's and Rakudo's `make m-test`, but there are still invalid frees, etc in 
NativeHelpers-iovec.

src/core/nativecall.c
```diff
@@ -299,6 +299,12 @@ char * MVM_nativecall_unmarshal_string(MVMThreadContext *tc, MVMObject *value, M
                 str = MVM_string_utf8_encode_C_string(tc, value_str);
         }
 
+#ifdef MVM_USE_MIMALLOC
+        char *libc_cstr = malloc(strlen(str) + 1);
+        strcpy(libc_cstr, str);
+        MVM_free(str);
```
nwc10, line 305: Better to store the result of `strlen` and then use `memcpy` instead of `strcpy`

src/core/nativecall_libffi.c
```diff
@@ -777,7 +777,11 @@ MVMObject * MVM_nativecall_invoke(MVMThreadContext *tc, MVMObject *res_type,
     /* Free any memory that we need to. */
     if (free_strs)
         for (i = 0; i < num_strs; i++)
+#ifdef MVM_USE_MIMALLOC
+            free(ree_strs[i]);
+#else
```
nwc10, line 782: `ree_strs` - oops. (The compiler found this, not me. I happen to have been building libffi locally)

MasterDuke17: Doh. A bad copy/paste that I noticed in the dynasm version, but not this one. Thanks

src/core/alloc.h
```diff
@@ -37,6 +37,23 @@ MVM_STATIC_INLINE void * MVM_realloc(void *p, size_t size) {
     return ptr;
 }
 
+/* This doesn't use any mimalloc functions because it's for use in NativeCall
+ * code, where modules may be freeing or allocating memory and wouldn't be 
+ * using mimalloc. */
+MVM_STATIC_INLINE void * recalloc(void *p, size_t old_size, size_t size) {
+    void *ptr = realloc(p, size);
```
nwc10, line 44: I think this code is only used in one place, and I think that we're unlikely to need it in a second place. And it's a non-standard "API" where `mimalloc` and `MVM` disagree about the parameters. So I'd be tempted to inline this into its callsite.

**but**

I'd be happy to do that as a later commit. I don't want to delay fixing the real regression. I was only adding this comment on the basis that I was trying to write a proper review, and I didn't want to let the perfect be the enemy of the good.

src/core/nativecall.c
```diff
@@ -299,6 +299,13 @@ char * MVM_nativecall_unmarshal_string(MVMThreadContext *tc, MVMObject *value, M
                 str = MVM_string_utf8_encode_C_string(tc, value_str);
         }
 
+#ifdef MVM_USE_MIMALLOC
+        size_t str_len = strlen(str) + 1;
+        char *libc_str = malloc(str_len);
+        memcpy(libc_str, str, str_len);
+        MVM_free(str);
+#endif
```
nwc10, line 307: The changes in this file are the only part of the entire commit that is really bugging me, because the `strlen` is wrong for the case of `MVM_NATIVECALL_ARG_UTF16STR`

(It's also clear that we don't have any tests for this, but that's a fix needed in the rakudo testsuite, I guess, and shouldn't distract us here...)

What's also bothering me is that I'd like not to introduce a (new) bug. The only sensible way I can see to get the length of the UTF-16 string is to call `MVM_string_utf16_encode_substr` and pass in `&output_size`. But once I start down that path, it feels silly to use that approach for UTF-16, but `strlen` for the 8-bit variants. Which ultimately ends up with my minimal change being "less than tiny":

```
$ git diff --stat src/core/nativecall.c
 src/core/nativecall.c | 57 +++++++++++++++++++++++++++++++++------------------
 1 file changed, 37 insertions(+), 20 deletions(-)
```

and it's then easier to just paste the entire (revised) function:

```
char * MVM_nativecall_unmarshal_string(MVMThreadContext *tc, MVMObject *value, MVMint16 type, MVMint16 *free, MVMint16 unmarshal_kind) {
    if (IS_CONCRETE(value)) {
        /* Set whether to free the encoded string. */
        if (free) {
            if (REPR(value)->ID == MVM_REPR_ID_MVMCStr)
                *free = 0; /* Manually managed. */
            else if (free && type & MVM_NATIVECALL_ARG_FREE_STR_MASK)
                *free = 1;
            else
                *free = 0;
        }

        /* Encode string. */
        MVMString *value_str = MVM_repr_get_str(tc, value);
        MVMuint64 output_size;
        char *str;
        type &= MVM_NATIVECALL_ARG_TYPE_MASK;
        switch (type) {
            case MVM_NATIVECALL_ARG_ASCIISTR:
                str = MVM_string_ascii_encode(tc, value_str, &output_size, 0);
                break;
            case MVM_NATIVECALL_ARG_UTF16STR:
                str = MVM_string_utf16_encode_substr(tc, value_str, &output_size, 0, -1, NULL, 0);
                break;
            default:
#ifdef MVM_USE_MIMALLOC
                /* This *doesn't* write a NUL */
                str = MVM_string_utf8_encode(tc, value_str, &output_size, 0);
#else
                /* whilst this wrapper that calls it reallocs and adds a NUL */
                str = MVM_string_utf8_encode_C_string(tc, value_str);
#endif
        }

#ifdef MVM_USE_MIMALLOC
        char *libc_str;
        if (type == MVM_NATIVECALL_ARG_UTF16STR) {
            /* output_size is in 2 byte units, and doesn't include the
             * terminating U-0000, which was written. */
            size_t octet_len = 2 * (output_size + 1);
            libc_str = malloc(octet_len);
            memcpy(libc_str, str, octet_len);
        }
        else {
            /* This code is simpler if we ignore the detail that ...ascii_encode
             * wrote a NUL, but ...utf8_encode did not. */
            size_t octet_len = output_size + 1;
            libc_str = malloc(octet_len);
            memcpy(libc_str, str, output_size);
            libc_str[output_size] = 0;
        }
        MVM_free(str);
        return libc_str;
#else
        return str;
#endif
    }
    else {
        return NULL;
    }
}
```

"Ultimately" (ie next week if not sooner) I'd like to go with something like the functions I proposed in my branch to remove the repeated malloc, but I'd like to get this fix merged, as it fixes a regression from the previous release.

So (I think) on the basis that I don't like commits that knowingly break stuff (only to be fixed in the next commit) (because `git bisect` has no idea about this), I'd like **something** merged into this commit that avoids the new bug, but I'm not sure that I'm recommending my own suggestion in this comment.

nwc10 (changes requested):

src/core/alloc.h
MasterDuke17: > I think this code is only used in one place, and I think that we're unlikely to need it in a second place. And it's a
> non-standard "API" where mimalloc and MVM disagree about the parameters. So I'd be tempted to inline this into
> its callsite.

Yeah, I hit the need for this pretty soon and just assumed I'd need it again. I'm 50/50 on this. Anyone else have an opinion?

src/strings/ascii.c
```diff
@@ -213,3 +213,42 @@ char * MVM_string_ascii_encode(MVMThreadContext *tc, MVMString *str, MVMuint64 *
...
+    else {
+        MVMuint32 i = 0;
+        MVMCodepointIter ci;
+        MVM_string_ci_init(tc, &ci, str, 0, 0);
+        while (MVM_string_ci_has_more(tc, &ci)) {
+            MVMCodepoint ord = MVM_string_ci_get_codepoint(tc, &ci);
+            if (i == result_alloc) {
+                result_alloc += 8;
+                result = MVM_realloc(result, result_alloc + 1);
+            }
```
nwc10, line 239: Pants! That should be `realloc`, not `MVM_realloc`

Yes, I've spotted a bug in my own code. This stuff is fiddly.

src/core/nativecall.c
```diff
@@ -293,19 +293,17 @@ char * MVM_nativecall_unmarshal_string(MVMThreadContext *tc, MVMObject *value, M
...
-                    str = libc_str;
-                }
-#endif
+            {
+                MVMuint64 output_size;
+                /* This adds a (two byte) U-0000 at the end, but doesn't include
+                 * that code point in the returned output_size. */
+                char *temp = MVM_string_utf16_encode_substr(tc, value_str, &output_size, 0, -1, NULL, 0);
+                output_size += 2;
+                str = malloc(output_size);
```
nwc10, line 302: Pants again. There is actually a bug in this code (that I think you spotted and commented on, but I never acknowledged, so it got forgotten). `MVM_string_utf16_encode_substr` returns a size in whatever-the-word-is-for-16bit-units. The correct(ed) code was in the comment I made at the top level of this ticket:

            /* output_size is in 2 byte units, and doesn't include the
             * terminating U-0000, which was written. */
            size_t octet_len = 2 * (output_size + 1);
            libc_str = malloc(octet_len);
            memcpy(libc_str, str, octet_len);


So the only mistakes/problems I can spot in your branch were caused by me. Sorry.

I guess amend the two commits, which probably fits with a rebase.

(The previous rebase was useful. As I had NQP and Rakudo checked out at `master` I needed to privately merge the old branch head with master to get the version checks to work.)

nwc10:
Looks good to me. The flappy failure of ` t/02-rakudo/reproducible-builds.t` makes no sense to me.

nwc10:
This was merged with 780667e0ca2c117f0aaa122b6cf0745123a2f345

PR #1655, opened 2022-01-27
Manually implement simplified strtoll for coerce_si

MasterDuke17:
This way we don't have to allocate a new C string just to call strtoll
on it. Additionally, because we're always converting to base 10, we can
remove some of the checks that strtoll does.

There are two code paths, both use the same algorithm, but when the
MVMString is just storing ascii characters we can use its storage
directly, otherwise we have to go through a codepoint iterator.

NQP builds ok and passes `make m-test` and Rakudo builds ok and passes `make m-test m-spectest`

This example `my str $s = (100_000_000..1_000_000_000).pick.Str; my int $i; my $n = now; $i = $s for ^10_000_000; say now - $n; say $i; say $s` took ~0.46s before and made 10_000_000 temporary allocations. With this PR it takes ~0.22s and doesn't make any temporary allocations.

### DIFF
--- a/src/core/coerce.c
+++ b/src/core/coerce.c
@@ -1,5 +1,6 @@
 #include "moar.h"
 #include "ryu/ryu.h"
+#include <ctype.h>
 
 #if defined(_MSC_VER)
 #define strtoll _strtoi64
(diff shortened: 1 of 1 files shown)


src/core/coerce.c
```diff
@@ -332,11 +333,131 @@ MVMString * MVM_coerce_n_s(MVMThreadContext *tc, MVMnum64 n) {
...
+ * OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF
+ * SUCH DAMAGE.
+ */
+    if (str->body.storage_type == MVM_STRING_GRAPHEME_ASCII) {
+        const MVMGraphemeASCII *s = str->body.storage.blob_ascii;
+        MVMStringIndex i = 0;
+        MVMGraphemeASCII c;
+        do {
+            c = *s++;
+        } while (i++ < strgraphs && isspace(c));
```
nwc10, line 383: I don't think we want to use `isspace` (or `isdigit` below). Or anything from `<ctype.h>` as all the functions are sensitive to C locales. IIRC the obscure trap here is that even if (well, "as") we don't call  `setlocale` anywhere, we're (still) subject to the "C" locale, which is only defined for ASCII. *Most* everywhere ends up treating the octets 128-255 as false for all `is*` functions, but (at least) VMS manages to have some return true, because (I inferred, remotely) that it was treating those octets as ISO-8859-1 (or whichever DEC character set forms the basis for ISO-8859-1). So presumably other platforms could do this (and have `isspace(0xA0)` return true, or any other implementation defined behaviour they care to torture us with)

So, I think we need to implement our own here, which really are hard coded for ASCII. In turn, I suspect that for speed, `isgraph` likely needs to be a lookup table, but `isdigit` I guess is comparisons `... >= '0' && ... <= '9'`.

nwc10 (commented):
I think that this is a good idea.

src/core/coerce.c
```diff
@@ -332,11 +333,131 @@ MVMString * MVM_coerce_n_s(MVMThreadContext *tc, MVMnum64 n) {
...
+    signed long long cutoff;
+    MVMint32  cutlim;
+
+    if (!strgraphs)
+        return result;
+
+/*-
+ * Copyright (c) 1990 The Regents of the University of California.
+ * with minor modifications to simplify and work in MoarVM
+ * All rights reserved.
```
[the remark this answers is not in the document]
MasterDuke17: https://github.com/gcc-mirror/gcc/blob/master/libiberty/strtoll.c yeah, I'll add it to the comment.

src/core/coerce.c
MasterDuke17: I don't have strong feelings either way, but since we're currently just using `strtoll()` from stdlib.h, aren't we already locale sensitive?

nwc10: Yes, that's a good point. We are.

I might view that as a bug :-)

Which would mean that it isn't troubling me that this is a (theoretical) "regression" and we leave the code as-is. For now, and explore whether replacing them in a future commit with something hard-coded to ASCII gives slightly better compiled code.

PR #1661, opened 2022-02-01
Fix things found by lgtm.com

MasterDuke17:

### DIFF
--- a/src/6model/reprs/P6opaque.c
+++ b/src/6model/reprs/P6opaque.c
@@ -34,7 +34,7 @@ static MVMint64 try_get_slot(MVMThreadContext *tc, MVMP6opaqueREPRData *repr_dat
         MVMP6opaqueNameMap *cur_map_entry = repr_data->name_to_index_mapping;
         while (cur_map_entry->class_key != NULL) {
             if (cur_map_entry->class_key == class_key) {
-                MVMuint16 i;
+                MVMuint32 i;
                 for (i = 0; i < cur_map_entry->num_attrs; i++) {
                     if (MVM_string_equal(tc, cur_map_entry->names[i], name)) {
                         return cur_map_entry->slots[i];
@@ -171,7 +171,7 @@ static void gc_mark_repr_data(MVMThreadContext *tc, MVMSTable *st, MVMGCWorklist
     if (repr_data->name_to_index_mapping) {
         MVMP6opaqueNameMap *cur_map_entry = repr_data->name_to_index_mapping;
         while (cur_map_entry->class_key != NULL) {
-            MVMuint16 i;
+            MVMuint32 i;
             for (i = 0; i < cur_map_entry->num_attrs; i++) {
                 MVM_gc_worklist_add(tc, worklist, &cur_map_entry->names[i]);
             }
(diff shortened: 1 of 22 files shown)


src/instrument/line_coverage.c
```diff
@@ -47,7 +47,7 @@ static void instrument_graph_with_breakpoints(MVMThreadContext *tc, MVMSpeshGrap
             continue;
         }
 
-        if (line_number >= 0 && line_number != last_line_number) {
+        if (line_number != last_line_number) {
             breakpoint_ins = MVM_spesh_alloc(tc, g, sizeof(MVMSpeshIns));
```
nwc10, line 51: I think that it's wrong here.

    MVMint64 line_number = -1;

That's a signed value. What does it think that `line_number` can't be less than zero? To me it looks possible to get to the `if` statement without `line_number` being changed.

src/6model/reprs/P6opaque.c
```diff
@@ -1186,12 +1186,12 @@ static void deserialize_repr_data(MVMThreadContext *tc, MVMSTable *st, MVMSerial
...
     for (i = 0; i < num_classes; i++) {
-        MVMint32 num_attrs = 0;
+        MVMuint16 num_attrs = 0;
 
         MVM_ASSIGN_REF(tc, &(st->header), repr_data->name_to_index_mapping[i].class_key,
             MVM_serialization_read_ref(tc, reader));
 
-        num_attrs = MVM_serialization_read_int(tc, reader);
+        num_attrs = (MVMuint16)MVM_serialization_read_int(tc, reader);
         repr_data->name_to_index_mapping[i].names = (MVMString **)MVM_malloc(P6OMAX(num_attrs, 1) * sizeof(MVMString *));
```
nwc10, line 1195: I think that these two are wrong. Instead `j` should be MVMuint32. ie split:

    MVMuint16 i, j, num_classes;

nwc10 (changes requested):
Everything else "looked good to me". I admit that I didn't double check the changes such as

```
-    MVMuint16 handle_index = 0;
+    MVMuint32 handle_index = 0;
 ```

where the type got larger. I also didn't double check the format string changes. But as they didn't warn on arm32 and x86_64 (and stuff still built on ppc64, which is big endian) I assumed that they were sane.

File: src/instrument/line_coverage.c (hunk 1 above)
Comment (MasterDuke17) [reply src/instrument/line_coverage.c:51]: To me it looks like control only gets here if the `if` block starting on line 25 is run (since its `else` has a `continue`), and line 26 is `line_number = bbba->line_number;` and `bbba->line_number` is an `MVMuint32` https://github.com/MoarVM/MoarVM/blob/master/src/core/bytecode.h#L5.

File: src/6model/reprs/P6opaque.c (hunk 2 above)
Comment (MasterDuke17) [reply src/6model/reprs/P6opaque.c]: Yep, fixed and force pushed.

File: src/instrument/line_coverage.c (hunk 1 above)
Comment (nwc10) [reply src/instrument/line_coverage.c:51]: Yes, I think that you're correct. Thanks for that analysis.

Looks like the signed/unsigned oversight has been present since that code was first added in commit 24af1679de

I'd not considered this directly before, but I'm thinking that having `continue` in an `else` block is something to be avoided. So (in a future commit) we change

```
        if (bbba) {
            line_number = bbba->line_number;
            filename_string_index = bbba->filename_string_heap_index;
            MVM_free(bbba);
        } else {
            bb = bb->linear_next;
            continue;
        }
```

to this:

```
        if (!bbba) {
            bb = bb->linear_next;
            continue;
        }

        line_number = bbba->line_number;
        filename_string_index = bbba->filename_string_heap_index;
        MVM_free(bbba);
```

(and more generally I'm going to do this in every language in every codebase, unless there's a better reason not to. For all control flow keywords.)

nwc10 (approved):

PR #1665, opened 2022-02-03
Make it possible turn dispatcher debugger with compiler options

vrurg:
Sometimes it is better than editing the source.

### DIFF
--- a/src/disp/program.c
+++ b/src/disp/program.c
@@ -2,10 +2,14 @@
 
 /* Debug dumping, to figure out what we're recording and what programs we are
  * inferring from those recordings. */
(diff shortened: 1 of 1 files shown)


jnthn (approved):

vrurg:
And sometimes I wish there be a way to turn debugging on/off at run-time for different subsystems. Like, say, `nqp::mvmdebug(MVM_DISP_RECORDINGS)` in NQP/Raku. And then in MoarVM code:

```C
if MVM_debug_on(tc, MVM_DISP_RECORDINGS) {
    /* produce some output */
}
```

PR #1669, opened 2022-02-06
Define -D__USE_MINGW_ANSI_STDIO=1 for MinGW

MasterDuke17:
This will fix it complaining about `%h` and `%z` being invalid sprintf
directives.


MasterDuke17:
This is not better. Now there are tons of
```
src/core/nativecall.h: In function 'MVM_nativecall_find_thread_context':
src/core/nativecall.h:145:22: warning: unknown conversion type character 'l' in format [-Wformat=]
         MVM_panic(1, "native callback ran on thread (%"PRId64") unknown to MoarVM",
```

PR #1673, opened 2022-02-22
Fix potential invalid free in nativecall

MasterDuke17:
When using mimalloc this particular memory comes from libc's malloc(), so
we have to use libc's free().

### DIFF
--- a/src/core/nativecall.c
+++ b/src/core/nativecall.c
@@ -282,50 +282,47 @@ double MVM_nativecall_unmarshal_double(MVMThreadContext *tc, MVMObject *value) {
     return (double)MVM_repr_get_num(tc, value);
 }
 
+char * MVM_nativecall_encode_string(MVMThreadContext *tc, MVMString *value_str, MVMint16 type, MVMint16 *free, MVMint16 unmarshal_kind, const MVMREPROps *repr) {
+    /* Encode string. */
+    char *str;
+    switch (type & MVM_NATIVECALL_ARG_TYPE_MASK) {
+        case MVM_NATIVECALL_ARG_ASCIISTR:
+            str = MVM_string_ascii_encode_malloc(tc, value_str);
+            break;
+        case MVM_NATIVECALL_ARG_UTF16STR:
+        {
+            MVMuint64 output_size;
+            /* This adds a (two byte) U-0000 at the end, but doesn't include
+             * that code point in the returned output_size. */
+            char *temp = MVM_string_utf16_encode_substr(tc, value_str, &output_size, 0, -1, NULL, 0);
+            /* output_size is in 2 byte units, and doesn't include the
+             * terminating U-0000, which was written. */
+            size_t octet_len = 2 * (output_size + 1);
+            str = malloc(octet_len);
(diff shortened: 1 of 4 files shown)


src/core/nativecall.c
```diff
@@ -282,50 +282,47 @@ double MVM_nativecall_unmarshal_double(MVMThreadContext *tc, MVMObject *value) {
...
-                 * terminating U-0000, which was written. */
-                size_t octet_len = 2 * (output_size + 1);
-                str = malloc(octet_len);
-                memcpy(str, temp, octet_len);
-                MVM_free(temp);
-                break;
-            }
-            default:
-                str = MVM_string_utf8_encode_C_string_malloc(tc, value_str);
+char * MVM_nativecall_encode_string(MVMThreadContext *tc, MVMString *value_str, MVMint16 type, MVMint16 *free, MVMint16 unmarshal_kind, const MVMREPROps *repr) {
```
niner, line 285: Since you already get the value and all callers pass `REPR(value)` as repr, you could just do that within the function and get rid of one argument.

src/core/nativecall.c
```diff
@@ -282,50 +282,47 @@ double MVM_nativecall_unmarshal_double(MVMThreadContext *tc, MVMObject *value) {
...
+            *free = 1;
+        else
+            *free = 0;
     }
+
+    return str;
+}
+
+char * MVM_nativecall_unmarshal_string(MVMThreadContext *tc, MVMObject *value, MVMint16 type, MVMint16 *free, MVMint16 unmarshal_kind) {
+    return IS_CONCRETE(value) ? MVM_nativecall_encode_string(tc, MVM_repr_get_str(tc, value), type, free, unmarshal_kind, REPR(value)) : NULL;
```
niner, line 325: Can you please wrap this line? Maybe like this:
```
return IS_CONCRETE(value)
    ? MVM_nativecall_encode_string(tc, MVM_repr_get_str(tc, value), type, free, unmarshal_kind)
    : NULL;
```

niner (changes requested):

src/core/nativecall.c
MasterDuke17: I tried that before (and just confirmed again), but it causes a fail in https://github.com/rakudo/rakudo/blob/master/t/04-nativecall/02-simple-args.t#L45
```
not ok 11 - checked previously passed string
# Failed test 'checked previously passed string'
# at t/04-nativecall/02-simple-args.t line 45
# expected: '7'
#      got: '0'
```

niner: Oooh....yes, now I see it. MVM_nativecall_unmarshal_string passes the unboxed string but the REPR of the original object.

MasterDuke17: Yeah, it's needed for https://github.com/MoarVM/MoarVM/pull/1673/files#diff-fcf886c5757a17d2e2429daadd4b01976cbc2a886a64f80b73d0412c2748a6b0R313-R314, but maybe there's a better/other way to accomplish that?

niner (approved):

PR #1674, opened 2022-02-22
The result of a getenv() call should not be freed

MasterDuke17:
"The string pointed to by the return value of getenv() may be statically
allocated, and can be modified by a subsequent call to getenv(), putenv(3),
setenv(3), or unsetenv(3)." - man 3 getenv


patrickbkr:
Out of interest: Was there a misbehavior that led to this finding or did you just happpen to notice?

MasterDuke17:
@[Coke] has been experiencing some invalid frees on windows that go away
when MoarVM is compiled with `--no-mimalloc`, so I was investigating
anywhere we free stuff inside `#ifdef WIN32`s.

PR #1678, opened 2022-03-03
Some nativeref optimizations and fixes

MasterDuke17:

I struggled to find a benchmark, but `my uint $a = 0; my uint $b; my uint $c; my uint $d; while $a++ < 10_000_000 { $b = $a; $c = $b; $d = $c; }; say now - INIT now; say($a); say($b); say($c); say($d)` before was ~1.25s and after was ~1.19s and callgrind reported ~10m fewer instructions. A spesh log confirmed that `decont_u` was being devirtualized before and double-devirtualized after.

### DIFF
--- a/src/6model/containers.c
+++ b/src/6model/containers.c
@@ -592,7 +592,7 @@ static void native_ref_store_u(MVMThreadContext *tc, MVMObject *cont, MVMuint64 
         MVM_exception_throw_adhoc(tc, "This container does not reference a native integer");
     switch (repr_data->ref_kind) {
         case MVM_NATIVEREF_LEX:
-            MVM_nativeref_write_lex_i(tc, cont, value); /* FIXME need a MVM_nativeref_write_lex_u */
+            MVM_nativeref_write_lex_u(tc, cont, value);
             break;
         case MVM_NATIVEREF_ATTRIBUTE:
             MVM_nativeref_write_attribute_u(tc, cont, value);
--- a/src/6model/containers.c
+++ b/src/6model/containers.c
@@ -720,14 +720,13 @@ static void native_ref_configure_container_spec(MVMThreadContext *tc, MVMSTable 
 }
 
 void *MVM_container_devirtualize_fetch_for_jit(MVMThreadContext *tc, MVMSTable *st, MVMuint16 type) {
-    if (type != MVM_reg_int64)
-        return NULL;
     if (st->container_spec == &native_ref_spec) {
         MVMNativeRefREPRData *repr_data = (MVMNativeRefREPRData *)st->REPR_data;
         switch (repr_data->ref_kind) {
             case MVM_NATIVEREF_LEX:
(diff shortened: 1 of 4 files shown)


src/6model/containers.c
```diff
@@ -720,35 +720,37 @@ static void native_ref_configure_container_spec(MVMThreadContext *tc, MVMSTable
...
 void *MVM_container_devirtualize_fetch_for_jit(MVMThreadContext *tc, MVMSTable *st, MVMuint16 type) {
-    if (type != MVM_reg_int64)
-        return NULL;
     if (st->container_spec == &native_ref_spec) {
         MVMNativeRefREPRData *repr_data = (MVMNativeRefREPRData *)st->REPR_data;
         switch (repr_data->ref_kind) {
             case MVM_NATIVEREF_LEX:
                 switch (type) {
                     case MVM_reg_int64: return MVM_nativeref_read_lex_i;
+                    case MVM_reg_uint64: return MVM_nativeref_read_lex_i;
```
zhuomingliang, line 729: MVM_nativeref_read_lex_u?

MasterDuke17: Based on https://github.com/MoarVM/MoarVM/blob/abb918cdbfcbead689dbcb1dad8833223c42c610/src/6model/containers.c#L476 I assumed I wouldn't need to add a MVM_nativeref_read_lex_u.

src/6model/reprs/NativeRef.c
```diff
@@ -597,6 +597,36 @@ void MVM_nativeref_write_lex_i(MVMThreadContext *tc, MVMObject *ref_obj, MVMint6
...
+        case MVM_reg_uint16:
+            var->u16 = (MVMuint16)value;
+            break;
+        case MVM_reg_uint32:
+            var->u32 = (MVMuint32)value;
+            break;
+        case MVM_reg_uint64:
+            var->u64 = value;
+            break;
+        case MVM_reg_int8:
```
niner, line 616: I think MVM_nativeref_write_lex_i need only handle signed int regs and MVM_nativeref_write_lex_u only unsigned int regs. native_ref_store_i and native_ref_store_u already check the ref type accordingly. By throwing in case of an unexpected reg type we can check whether my understanding is wrong or if we maybe take a wrong turn somewhere.

niner (changes requested):

niner:
My change request isn't super important, but if you have the time, it'd be nice. Otherwise I'm glad those FIXMEs slowly disappear :)

src/6model/reprs/NativeRef.c
MasterDuke17: Done in [fed7bb8](https://github.com/MoarVM/MoarVM/pull/1678/commits/fed7bb86ce28630083e48162d9574ed98212718e), no problems in NQP build+test or Rakudo build+test.

niner (approved):

PR #1680, opened 2022-03-04
Fix all returned native integers getting treated as signed

niner:


src/core/args.c
```diff
@@ -769,13 +778,83 @@ void MVM_args_set_result_int(MVMThreadContext *tc, MVMint64 result, MVMint32 fra
...
+        else if (MVM_spesh_log_is_logging(tc))
+            MVM_spesh_log_return_to_unlogged(tc);
+        target = tc->cur_frame->caller;
+    }
+    if (target) {
+        switch (target->return_type) {
+            case MVM_RETURN_VOID:
+                if (tc->cur_frame->static_info->body.has_exit_handler)
+                    save_for_exit_handler(tc,
+                        MVM_repr_box_int(tc, MVM_hll_current(tc)->int_box_type, result));
```
MasterDuke17, line 798: Why isn't this `uint_box_type`?

src/core/frame.c
```diff
@@ -906,6 +906,9 @@ MVMuint64 MVM_frame_try_return(MVMThreadContext *tc) {
                     case MVM_RETURN_INT:
                         result = MVM_repr_box_int(tc, hll->int_box_type, caller->return_value->i64);
                         break;
+                    case MVM_RETURN_UINT:
+                        result = MVM_repr_box_uint(tc, hll->int_box_type, caller->return_value->u64);
```
MasterDuke17, line 910: Same question here re `uint_box_type`.

src/spesh/optimize.c
```diff
@@ -2279,6 +2279,7 @@ static void optimize_bb_switch(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpeshB
         case MVM_OP_dispatch_n:
         case MVM_OP_dispatch_s:
         case MVM_OP_dispatch_i:
+        case MVM_OP_dispatch_u:
             MVM_spesh_disp_optimize(tc, g, bb, p, ins, &next_ins);
             break;
         case MVM_OP_sp_guard:
```
MasterDuke17, line 2285: Shouldn't there also be a case for `MVM_OP_sp_runbytecode_u` added a couple lines below?

src/core/args.c
niner: That's something I kinda hate about my uint changes, but also something that I don't really know a way around. uint_box_type is UInt which is not a proper type, but a subset. We need to know UInt so we can determine properties like most importantly signedness. But we cannot use it to box a value. I guess an alternative could be to split it into uint_box_type and uint_unbox_type.

src/spesh/optimize.c
niner: Oh, yes, indeed! Nice catch!

src/core/args.c
MasterDuke17: Oh right, ISTR some chat with lizmat about that a little while ago. Now I wonder if that's what was causing the problems I saw when I added an `hllboxtype_u` op and had some of the arg handling use that...

MasterDuke17 (approved):

PR #1681, opened 2022-03-06
Fix spesh of named uint parameters

MasterDuke17:
Handle the newly added `*_u` versions of named parameters ops. Now
something like `sub a(uint :$a!) { $a ?? $a + 1 !! 0 }; my uint $a = 1;
for ^1_000_000 { $a = a(:$a); }; say $a;` no longer dies with `Unexpected
named argument 'a' passed`, even though it finishes fine with a smaller
number of iterations in the loop.

I had a version of this that added `hllboxtype_u` and `sp_getargs_u` ops, but those caused sigkills in the NQP build and I'm not sure why. This does fix the problematic example given above, but I'm not sure if `hllboxtype_u` and `sp_getargs_u` really should be added and used.


niner (approved):
Excellent work!

PR #1685, opened 2022-03-13
Use JFS64 instead of TinyMT64 as our PRNG

MasterDuke17:
The Mersenne Twister is now a bit outdated and there are alternatives
that are both faster and have better statistical properties.

`my num $r; my int $i := 0; my int $s := nqp::time; $r := nqp::rand_n(1e8) while $i++ < 100_000_000; say(nqp::div_n(nqp::time - $s, 1000000000e0)); say($r)` drops from ~0.58s to ~0.48s (with `MVM_SPESH_BLOCKING=1`).


MasterDuke17:
@nwc10 it looks like the fails are all on Windows (MSVC + MinGW). I think the only Windows-specific bit is https://github.com/MoarVM/MoarVM/pull/1685/files#diff-e4b5afce4611608fd85a5a5579c1a585e34a2b43573725201841976980f84f6dR9-R17, any idea why we'd have a problem with it?

src/io/procops.c
```diff
@@ -1094,22 +1094,21 @@ MVMint64 MVM_proc_getppid(MVMThreadContext *tc) {
...
 
 /* generates a number between 0 and 1 */
 MVMnum64 MVM_proc_rand_n(MVMThreadContext *tc) {
-    return tinymt64_generate_double(tc->rand_state);
+    return ((jfs64_generate_uint64(tc->rand_state) >> 11) / (double)(1L << 53));
 }
 
 MVMnum64 MVM_proc_randscale_n(MVMThreadContext *tc, MVMnum64 scale) {
-    return tinymt64_generate_double(tc->rand_state) * scale;
+    return ((jfs64_generate_uint64(tc->rand_state) >> 11) / (double)(1L << 53)) * scale;
```
niner: Is there a way to get the scaled number without losing so much accuracy? Feels like a shame to throw away those random bits.

MasterDuke17: No idea. Maybe something like shifting by ‘11 - magnitude(scale)’ and then dividing by ‘max_representable_int * scale’ would work, but I don’t know the math well enough to be sure that’s safe and correct.

MasterDuke17: Based on https://stackoverflow.com/questions/46901022/how-to-convert-a-uint64-t-to-a-double-float-between-0-and-1-with-maximum-accurac (and some other things I've read) and our discussion, I think there isn't actually any more precision to be gained.

PR #1686, opened 2022-03-17
Correctly allocate/free CStrs when using mimalloc

MasterDuke17:


coke:
This fixes the difficult to golf issue I had on windows that occurred using mimalloc.

niner (approved):

PR #1690, opened 2022-03-22
Mimalloc should only release memory to the OS at exit if --full-cleanup was used

MasterDuke17:


src/moar.c
```diff
@@ -756,6 +756,11 @@ void MVM_vm_destroy_instance(MVMInstance *instance) {
 
     /* Clear up VM instance memory. */
     MVM_free(instance);
+
+#ifdef MVM_USE_MIMALLOC
+    /* Ask mimalloc to free any memory it's holding onto. */
+    mi_collect(true);
+#endif
```
jnthn, line 763: The idea seems sensible, but what if somebody creates more than one `MVMInstance` in the same process? Perhaps this should not be here, but rather in the caller(s) of `MVM_vm_destroy_instance` when we're doing a full cleanup?

moon-chilled: FWIW last time I tried to create multiple MVMInstances in one process, I got a segfault creating the second one.  (This was probably around 2-3 years ago.)

MasterDuke17: @jnthn If I understand your concern correctly (if one instance is destroyed, we don't want to `free()` memory another instance might be using and cause a use-after-free), I think I can resolve it by saying that "free" is/was a bad word to use in that comment since it could cause confusion with `free()`. AIUI, `mi_collect()` is like `malloc_trim()`, it's about returning to the OS memory that's *already* been `free()`ed https://microsoft.github.io/mimalloc/group__extended.html#ga421430e2226d7d468529cec457396756, so should be safe to call at any time. So it might not even be necessary to call it at all in this case, but without knowing how valgrind/heaptrack/etc might interact with mimalloc, I wanted to minimize any chance of false positives (i.e., valgrind/heaptrack/etc incorrectly thinking there was a leak when using `--full-cleanup`).

jnthn: Ah, got it, thanks.

jnthn (approved):

PR #1695, opened 2022-03-29
Fix misspesh of native calls

niner:


src/spesh/disp.c
```diff
@@ -1250,6 +1252,27 @@ static int translate_dispatch_program(MVMThreadContext *tc, MVMSpeshGraph *g,
...
+                                case MVM_NATIVECALL_ARG_USHORT:
+                                case MVM_NATIVECALL_ARG_UINT:
+                                case MVM_NATIVECALL_ARG_ULONG:
+                                case MVM_NATIVECALL_ARG_ULONGLONG: {
+                                    MVMSpeshOperand var = skip_args >= 0 ? args[skip_args + j] : temporaries[dp->first_args_temporary + j];
+                                    is_rw_operand[j] = 1;
+                                    rw_operands[j] = MVM_spesh_manipulate_get_temp_reg(tc, g, MVM_reg_uint64);
+                                    emit_bi_op(tc, g, bb, &insert_after, MVM_OP_decont_u, rw_operands[j], var);
+
+                                    callsite = MVM_callsite_drop_positional(tc, callsite, j);
```
MasterDuke17, line 1265: Couldn't this pair of drop+insert be an `MVM_callsite_replace_positional` instead?

niner: Indeed it can! MVM_callsite_replace_positional just didn't exist yet when the original code was written.

PR #1699, opened 2022-04-10
Take stdatomic.h probe into account in decision whether or not to use c11 atomics 

MasterDuke17:
Had to move the probe earlier, and then also some of the ldlib setup code.


niner (approved):
Looking good! Please sqash before merging

Issue #1704, opened 2022-05-23
src/io/syncsocket.c:287:38: error: 'AI_NUMERICSERV' undeclared (first use in this function)


barracuda156:
`MoarVM` fails to update to 2022.04 on 10.6 (build for `ppc`):
```
src/io/syncsocket.c: In function 'MVM_io_resolve_host_name':
src/io/syncsocket.c:287:38: error: 'AI_NUMERICSERV' undeclared (first use in this function); did you mean 'NI_NUMERICSERV'?
  287 |     hints.ai_flags = AI_ADDRCONFIG | AI_NUMERICSERV | AI_PASSIVE;
      |                                      ^~~~~~~~~~~~~~
      |                                      NI_NUMERICSERV
src/io/syncsocket.c:287:38: note: each undeclared identifier is reported only once for each function it appears in
src/io/syncsocket.c: At top level:
cc1: note: unrecognized command-line option '-Wno-logical-op-parentheses' may have been intended to silence earlier diagnostics
make: *** [src/io/syncsocket.o] Error 1
```
Config line:
```
Executing:  cd "/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_lang_MoarVM/MoarVM/work/MoarVM-2022.04" && /opt/local/bin/perl Configure.pl --prefix=/opt/local --cc=/opt/local/bin/gcc-mp-11 --has-dyncall --has-libatomic_ops --has-libtommath --has-libuv --no-mimalloc
```

niner:
Is this about Mac OS X 10.6? According to Wikipedia that has been out of support from Apple for more than 8 years already.

barracuda156:
Yes, but this error is probably fixable for 10.6. The previous version of MoarVM has built fine.
```
36-111% port -v installed moarvm
The following ports are currently installed:
  MoarVM @2022.02_0 requested_variants='' platform='darwin 10' archs='ppc' date='2022-03-22T08:10:05+0800'
  MoarVM @2022.02_1 (active) requested_variants='' platform='darwin 10' archs='ppc' date='2022-04-16T16:01:15+0800'
```
P. S. I was going to update the port in Macports, but ran into this problem. We need to fix it for PPC systems, they are supported by Macports.

niner:
We have used AI_NUMERICSERV since 2019, so it looks like a build tool chain issue. Maybe it's using the wrong compiler or libc or something like that?

barracuda156:
[...]

PR #1705, opened 2022-05-24
Add definition of AI_NUMERICSERV, fixes build on MacOS 10.6 for ppc

barracuda156:
This fixes a build on 10.6 for `ppc` and closes my own ticket: https://github.com/MoarVM/MoarVM/issues/1704

The fix is borrowed from `graphviz` code: https://gitlab.com/graphviz/graphviz/-/blob/main/cmd/lefty/os/unix/io.c
See also: https://trac.macports.org/ticket/41916

### DIFF
--- a/src/io/syncsocket.c
+++ b/src/io/syncsocket.c
@@ -5,6 +5,10 @@
 #define snprintf _snprintf
 #endif
 
+#ifndef AI_NUMERICSERV
+#define AI_NUMERICSERV 0
+#endif
+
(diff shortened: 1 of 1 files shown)


barracuda156:
@patrickbkr @lizmat Could someone review this please? We have this fix in Macports now, but it is preferable to have it in upstream, that gonna benefit those building without Macports.

barracuda156:
P. S. Rebased commit onto the current master.

jnthn (approved):

PR #1708, opened 2022-06-14
Attempt at implenting a "bytecode-size" dispatch

lizmat:
This is a WIP.

For some reason, this does not actually compile, and my C is too rusty

### DIFF
--- a/src/disp/syscall.c
+++ b/src/disp/syscall.c
@@ -1194,6 +1194,22 @@ static MVMDispSysCall set_cur_hll_config_key = {
     .expected_concrete = { 1, 0 },
 };
 
+/* code-bytecode-size */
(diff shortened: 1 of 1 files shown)


MasterDuke17:
I think you want `MVMuint32 bytecode_size = ((MVMCode *)obj)->body.sf->body.bytecode_size;`

lizmat:
Yup, that did the trick.  But I fail to understand why that extra `->body` is needed?

lizmat:
Aha!  I completely missed the "Body" part at https://github.com/MoarVM/MoarVM/blob/master/src/6model/reprs/MVMStaticFrame.h#L3

Completely understand now.  I should have looked closer  :-)

PR #1712, opened 2022-07-12
Fallback resolver

niner:
When we run code at BEGIN time the actual lexical environment the code
is supposed to run in hasn't been set up yet. Previously we got around
that by compiling a wrapper frame, containing all lexicals visible from
the code we run. This frame is rather expensive and we want to replace
it with an on-demand resolver that's part of the compiler (i.e. HLL
code)


jnthn (commented):
Makes sense to me. Is there any change needed in `oplist`, so indicate that the various lexical lookup ops might now end up invoking?

src/core/oplist
```diff
@@ -89,17 +89,17 @@ DEPRECATED_63       r(obj) ins
...
 getlex              w(`1) rl(`1) :pure :deoptonepoint :logged
 bindlex             wl(`1) r(`1)
-getlex_ni           w(int64) str :pure :noinline
-getlex_nn           w(num64) str :pure :noinline
-getlex_ns           w(str) str :pure :noinline
-getlex_no           w(obj) str :pure :deoptonepoint :logged :noinline
+getlex_ni           w(int64) str :pure :noinline :invokish :throwish
+getlex_nn           w(num64) str :pure :noinline :invokish :throwish
+getlex_ns           w(str) str :pure :noinline :invokish :throwish
+getlex_no           w(obj) str :pure :deoptonepoint :logged :noinline :invokish :throwish
```
jnthn, line 95: If they can call now, I wonder if also `deoptallpoint` applies?

niner:
> If they can call now, I wonder if also `deoptallpoint` applies?

Wouldn't the same be true then for assign and decont which can also call?

jnthn:
> Wouldn't the same be true then for assign and decont which can also call?

Yes. Seems `istype` carires it, at least.

jnthn (approved):

PR #1714, opened 2022-07-20
[WIP] Make some enhancements to the Semaphore REPR

Kaiepi:
- The gist of this PR is the first commit message. tl;dr libuv => atomics and the niceties we can gain for the `Semaphore` REPR by them.
- JIT is TODO.
- More testing needed throughout.


src/6model/reprs/Semaphore.c
```diff
@@ -113,25 +160,60 @@ static const MVMREPROps Semaphore_this_repr = {
...
     interval_id = MVM_telemetry_interval_start(tc, "Semaphore.acquire");
     MVMROOT(tc, sem, {
+        MVMuint64 count;
         MVM_gc_mark_thread_blocked(tc);
-        uv_sem_wait(sem->body.sem);
+#ifdef MVM_USE_C11_ATOMICS
+        while (atomic_flag_test_and_set_explicit(&sem->body.waits, memory_order_acquire))
+            MVM_thread_yield(tc);
+        while (!(count = atomic_load_explicit(&sem->body.count, memory_order_acquire)))
+            MVM_thread_yield(tc);
```
jnthn, line 189: I did this in the MoarVM GC, before replacing it with condition variables. In theory it should be fine; in practice it wasn't, because yield can result on the thread being put back on the CPU in very short order, and so it's not that much better than busy-waiting.

Kaiepi: Do you think `uv_rwlock_t` might apply here?

Kaiepi:
I've given a handful of other approaches a go, but other synchronization primitives don't appear to play nice with the nice thing we should have going with the acquire/release ordering:
- `uv_mutex_t` alone is prone to deadlocks
- `uv_mutex_t` + `uv_cond_t` ditto
- `uv_rwlock_t` ditto

To my knowledge, this implementation doesn't terribly differ from [OpenBSD's](https://github.com/openbsd/src/blob/master/lib/librthread/rthread_sem.c) given the operations we need to support at the moment. The key difference is the `yield` is eschewed in favor of a `futex` call with no wait. Regardless, `sleep(0e0)` is racey for whatever reason.

Kaiepi:
Others have had success with [splitting the counter](https://travisdowns.github.io/blog/2020/07/06/concurrency-costs.html) to reduce the cost of waiting.

Kaiepi:
Semaphore serialization went missing at some point, but I figure it's easier to deal with once the rest of the REPR's sorted out.

Kaiepi:
Closing in favor of https://github.com/rakudo/rakudo/pull/5000 etc. as per @jnthn's recommendation.

PR #1715, opened 2022-07-24
Replace the MVMROOT's block argument with varargs

Kaiepi:
Legitimate usage of a comma within the block, e.g. `MVMuint64 a, b;` results in an avoidable compiler error otherwise.

A change from my `semaphoverhaul` branch that I figure stands on its own.


lizmat:
I think this PR has bitrotted, so closing.  Please re-open if disagree.

PR #1721, opened 2022-09-28
Add unsigned comparison ops (eq, ne, (l|g)(e|t))

MasterDuke17:
Copied from their signed versions, but with unsigned operands in the interpreter and unsigned ops in the jit.


src/core/interp.c
```diff
@@ -355,9 +355,6 @@ void MVM_interp_run(MVMThreadContext *tc, void (*initial_invoke)(MVMThreadContex
                 GC_SYNC_POINT(tc);
                 goto NEXT;
             }
-            OP(DEPRECATED_62):
```
lizmat, line 358: I understood it was bad to remove ops?

lizmat: Ah, I see, the _u candidates are using previously deprecated slots.  Won't that affect being able to execute older binaries?

MasterDuke17: You mean old precomp files? I believe it's been done before, the reason being that it means the moarvm files don't increase in size quite as much as if we always added new ops in new slots. However, as I'm typing this I realized that I just used the next deprecated slot down the oplist, but maybe I should instead have used the oldest? Ping @niner @jnthn @timo

jnthn: Deprecated here means "any use of the op will throw an exception", not "you can use it but should stop soon". So anything still referencing them will have long been broken.

niner:
On Wednesday, 28 September 2022 10:10:34 CEST Elizabeth Mattijsen wrote:

> Ah, I see, the _u candidates are using previously deprecated slots.  Won't
> that affect being able to execute older binaries?

Ostensibly yes, but those ops have been deprecated so long ago that it doesn't 
matter. The only realistic scenario would be trying a very old nqp bootstrap 
with a current MoarVM and that would likely break for a whole lot of other 
reasons.

PR #1724, opened 2022-10-14
Fix nqp::objprimunsigned handling of unsigned or sized integers

Kaiepi:
This would compare the `boxed_primitive` flag against `MVM_STORAGE_SPEC_BP_INT` before returning the `is_unsigned` flag, but this excludes the other sizes of `int`, as well as `uint` altogether. Give this the same check as `nqp::objprimbits`; encourage stringy types to default to zero themselves.

Required in order for https://github.com/Raku/nqp/pull/781 to pass its tests.


Kaiepi:
Forgot the JIT... sec.

niner (approved):

PR #1731, opened 2022-12-19
Speedup creating sha1 digest string

MasterDuke17:
Instead of using sprintf to convert the byte array into the hex string, just implement the conversion manually.

`my $a; my int $i := 0; my int $s := nqp::time; while $i++ < 1_000_000 { $a := nqp::sha1("a") }; say(nqp::div_n(nqp::time - $s, 1000000000)); say($a)` goes from ~1.44s to ~0.48s. However, real-life impact will be much less, since we usually hash much larger strings and then the main cost is the UTF8 encoding and actual hashing.

### DIFF
--- a/3rdparty/sha1/sha1.c
+++ b/3rdparty/sha1/sha1.c
@@ -225,15 +225,16 @@ void SHA1_Digest(SHA1Context* context, unsigned char digest[SHA1_DIGEST_SIZE])
 void SHA1Final(SHA1Context *context, char *output)
 {
     unsigned char digest[20];
-    int i,j;
+    int i;
     char *c = output;
 
     SHA1_Digest(context, digest);
     
-    for (i = 0; i < SHA1_DIGEST_SIZE/4; i++) {
-        for (j = 0; j < 4; j++) {
-            sprintf(c,"%02X", digest[i*4+j]);
-            c += 2;
-        }
+    unsigned char *d   = digest;
+    const char    *hex = "0123456789ABCDEF";
+    for (i = 0; i < 20; i++) {
+        *c++ = hex[(*d >> 4) & 0xF];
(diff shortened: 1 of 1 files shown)


lizmat:
Shouldn't we always do SHAs on bytes, rather than codepoints?

2colours:
> Shouldn't we always do SHAs on bytes, rather than codepoints?

What do you mean? Mind you, this is vanilla C so "char" is guaranteed to be really just a byte, and the string literal is just a (zero-terminaed) array of bytes.

MasterDuke17:
https://github.com/MoarVM/MoarVM/blob/42a3a2c0578a2df959791686a3e2f55584bcc367/src/6model/serialization.c#L3200 is what I was referring to. I guess the point is to get what’s hashed in a normalized representation? But maybe that’s not strictly necessary?

lizmat:
I don't think it's necessary.

I think in the past the issue was raised why you weren't able to create a SHA1 on a `Buf`.

OTOH, do we want a string with decomposed codepoints generate the same SHA1 as a string with composed codepoints?   I guess that's what was one of the reasons for creating a SHA1 from a string.

OYTOH, I think if someone would like to create a SHA1 from a Buf, they should be able to do so.

MasterDuke17:
> OYTOH, I think if someone would like to create a SHA1 from a Buf, they should be able to do so.

I agree, and adding that capability could be done, but is unrelated to this PR.

lizmat:
If it would be possible to get support for buf, or even better, support for creating a SHA1 from a PIO, then a lot of module loading would become faster, especially in the `-I` case.

MasterDuke17:
> If it would be possible to get support for buf, or even better, support for creating a SHA1 from a PIO, then a lot of module loading would become faster, especially in the `-I` case.

I can experiment with creating those.

PR #1737, opened 2023-01-05
Update to Unicode 15.0

thundergnat:
tools/ucd2c.pl is hand edited.

Everything else is automatically generated by running tools/ucd2c.pl

Note: Rakudo built on this MoarVM/nqp will fail testing until https://github.com/rakudo/rakudo/pull/5150 is applied.

This builds & tests clean MoarVM and nqp on all of the azure pipeline CI tests. **All** of the CI build failures below are because a Rakudo built with this will fail until it is updated as well.

The CodeFactor issue is a complaint about a "complex method". The method in question is  auto-generated by tools/ucd2c.pl. It _is_ complex, because it has to be. The change is that line 38 was added encompassing a new CJK block.

https://github.com/MoarVM/MoarVM/blob/3d21ef2e37df1df72b81ec93c3752cc249b9d427/src/strings/unicode_uca.c#L21-L40


lizmat:
Unless someone objects, I will merge this later today

lizmat:
That I do not know:  @thundergnat ?

thundergnat:
As far as I could tell, there are no _new_ rules for grapheme composition or handling, just lots of additions to existing rule classes.

lizmat:
Ok, I will then do the merge and the necessary bumps and other merges tomorrow.

Unless @jnthn stops me.

PR #1738, opened 2023-02-08
Speedup VMArray's deserialize()...

MasterDuke17:
by moving a switch outside of a loop. This means we have to duplicate the loop inside each case statement, but the compiler generates better code. For `raku -e ''`, we deserialize almost 90k VMArrays. With `MVM_SPESH_DISABLE=1` we go from ~626k to ~623k instructions and with `MVM_SPESH_BLOCKING=1` from ~1.135b to ~1.132b instructions.

### DIFF
--- a/src/6model/reprs/VMArray.c
+++ b/src/6model/reprs/VMArray.c
@@ -1275,47 +1275,57 @@ static void deserialize(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, vo
     if (body->ssize)
         body->slots.any = MVM_malloc(body->ssize * repr_data->elem_size);
 
-    for (i = 0; i < body->elems; i++) {
-        switch (repr_data->slot_type) {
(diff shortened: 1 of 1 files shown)


niner (approved):

lizmat:
Do we have an idea which types of arrays are deserialized most?  Maybe those cases should be tried first?  I'm thinking U32 specifically?

MasterDuke17:
> Do we have an idea which types of arrays are deserialized most? Maybe those cases should be tried first? I'm thinking U32 specifically?

For `raku -e ''`, there are 89,928 MVM_ARRAY_OBJs , 8 MVM_ARRAY_STRs, 2 MVM_ARRAY_I64s, and 1 MVM_ARRAY_N64. I'll admit I wasn't exactly expecting those results, but since MVM_ARRAY_OBJ is already the first case in the switch I'll just leave well enough alone.

PR #1740, opened 2023-02-10
Update GitHub CI pipelines to track branch renames and OS deprecations

MasterDuke17:


patrickbkr (commented):
This is the Azure pipeline it has nothing to do with GitHub actions.

MasterDuke17:
> Also the respective pipelines in NQP and rakudo should be changed as well, so that they match up.

Yep, once this is good I'll try and sync NQP and Rakudo.

Issue #1743, opened 2023-02-20
Avoid MAX_PATH issue on windows


coke:
Code in src/io/dirops.c is explicitly checking for MAX_PATH on windows, which means that even if a user has disabled that check in the windows registry, raku will still fail to deal with paths greater than MAX_PATH length.

* `MVM_dir_open`
* `MVM_dir_mkdir`

Should be updated to use the corresponding libuv functions, which will hopefully hide any OS specific code in the library, and allow windows users to use longer path names.

MasterDuke17:
FYI, I started on this last night. I think I'm mostly done, but I don't
find the libuv documentation all that clear and so I still need to get the
synchronization between closing dir handles and our gc process right.

coke:
I'd be happy to try out a work in progress, btw.

coke:
Build error on windows:

```
linking 3rdparty\libuv\uv.lib
linking moar.dll
   Creating library moar.dll.lib and object moar.dll.exp
dirops.obj : error LNK2001: unresolved external symbol S_ISDIR
moar.dll : fatal error LNK1120: 1 unresolved externals
```

which only appears here:
```
$ grep -r S_ISDIR
3rdparty/libuv/src/unix/fs.c:    if (rc == 0 && S_ISDIR(buf.st_mode)) {
```

coke:
Found https://www.davekb.com/browse_programming_tips:windows_isreg_isdir:txt

coke:
Maybe the issue is: why am I linking against unix/fs.c and not win/fs.c ?

coke:
> Maybe the issue is: why am I linking against unix/fs.c and not win/fs.c ?

Oh - S_ISDIR is also in the new code, that's why I'm getting the error. I had a "moarvm-md" and a "moar-md" folder next to each other and was looking in one on git bash and the other on shell and confused myself.

So, yes, the link with suggested implementation of S_ISDIR on windows is probably the way to go.

coke:
One the one hand, I'm able to build, but the rakudo install fails with:

```
+++ Preparing installation
NMAKE : fatal error U1077: 'C:\sandbox\rakudo-md\rakudo-m.exe' : return code '0xc0000005'
Stop.
```

On the OTHER hand, rakudo-m.exe was installed, and I can run it.

On the GRIPPING hand, when I run rakudo-m.exe on my failing test script, I get a segfault.

coke:
[...]

PR #1745, opened 2023-02-24
Convert all dir functions to just use libuv...

MasterDuke17:
instead of doing our own `#define`s for Windows support.

Almost all the logic is the same. However, libuv removed '.' and '..' from readdir results, and since Rakudo (and roast) expects them, this will require a minor change in Rakudo. The JVM also doesn't return those for its readdir implementation, so there's already code that deals with that in Rakudo in `#?if jmv`s, which will just need to be adjusted to be `#?if !js`.

NQP builds ok and passes `make m-test` and Rakudo (with the aforementioned changes) builds ok and passes `make m-test m-spectest`. Might be a fix for https://github.com/MoarVM/MoarVM/issues/1743.


coke:
nmake failure:
```
src\io\dirops.c(59): error C2065: 'rmdir_erro': undeclared identifier
NMAKE : fatal error U1077: '"C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.29.30133\bin\HostX64\x64\cl.EXE"' : return code '0x2'
```

coke:
If I fix the typo, I can build moarvm/nqp/rakudo, but the rakudo install fails:

```
+++ Creating installation directories
+++ Removing old files
+++ Installing files
+++ Preparing installation
NMAKE : fatal error U1077: 'C:\sandbox\rakudo-md\rakudo-m.exe' : return code '0xc0000005'
Stop.
```

and if I try to get a directory listing with the built/not installed version, I get nothing, whereas the installed "off the shelf" raku generates output.

```
C:\sandbox\rakudo-md>.\rakudo-m.exe -e "dd 'C:/sandbox/'.IO.dir"

C:\sandbox\rakudo-md>raku -e "dd 'C:/sandbox/'.IO.dir"
<a screen of elided output>
```

coke:
> but the rakudo install fails:

I did a re-rebuild of this moarvm with --debug to track down the install error; I reinstalled moarvm, went immediately back to rakudo... and the install worked fine. I suspect user error on the previous failure somehow.

MasterDuke17:
Hm. I definitely had all passing tests, I can see them in my terminal scrollback. However, I realize now it may have been when I accidentally forgot to remove `--optimize=0` from my MoarVM configure.

ugexe:
```
PS C:\rakubrew\versions\moar-2023.02\foo> raku -e 'my $path = q|a| x 245; mkdir $path; say $path.IO.e'
Failed to create directory 'C:\rakubrew\versions\moar-2023.02\foo\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' with mode '0o777': Failed to mkdir: Unknown error
  in block  at 'SETTING::'src/core.c/IO/Path.pm6 line 539
  in any  at 'SETTING::'src/core.c/IO/Path.pm6 line 538
  in method mkdir at 'SETTING::'src/core.c/IO/Path.pm6 line 541
  in sub mkdir at 'SETTING::'src/core.c/io_operators.pm6 line 312
  in block <unit> at -e line 1
 
Actually thrown at:
  in method throw at 'SETTING::'src/core.c/Exception.pm6 line 64
  in method throw at 'SETTING::'src/core.c/Failure.pm6 line 65
  in method sink at 'SETTING::'src/core.c/Failure.pm6 line 120
  in block <unit> at -e line 1
```
The above is the output with this PR and the rakudo PR removing `./..`. Without those changes that same one-liner outputs `False` (and if I change the path length to something much shorter, its `True`).

I have long paths enabled fwiw

ugexe:
Note that running `node -e "var fs = require('fs'); fs.mkdirSync('a'.repeat(245));"` in the same directory as my previous example works. So theoretically long paths should be able to work.

MasterDuke17:
Huh. Memory management problems aside, I thought this would solve the large path issue. Maybe I need to find node’s implementation to see how it uses the libuv functions.Sent from my iPhoneOn Feb 24, 2023, at 10:47 PM, Nick Logan ***@***.***> wrote:﻿
Note that running node -e "var fs = require('fs'); fs.mkdirSync('a'.repeat(245));" in the same directory as my previous example works. So theoretically long paths should be able to work.

ugexe:
Hmm, so more details I missed previous: running the one-liner on 2022.12 prints `False` (i.e. `$path.IO.e` is false) but it *does* create the directory. With the raku/moarvm PRs it does not create the directory. The node variant creates the directory, and `fs.existsSync(...)` is true.

ugexe:
```
PS C:\rakubrew\versions\moar-2022.02\foo> raku -e 'my $path = q|a| x 245; mkdir $path; say $path.IO.e; my $file = $path.IO.add(q|myfile.txt|); $file.open(:create); say $file.IO.e;'
False
Failed to open file C:\rakubrew\versions\moar-2023.02\foo\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\myfile.txt: No such file or directory
  in block <unit> at -e line 1
 
PS C:\rakubrew\versions\moar-2023.02\foo> ls
 
 
    Directory: C:\rakubrew\versions\moar-2023.02\foo
 
 
Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         2/24/2023  12:02 PM                aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
```
This is the current raku behavior (no PRs applied).

ugexe:
```
PS C:\rakubrew\versions\moar-2023.02\foo> node -e "var fs = require('fs'); const mydir = 'a'.repeat(245); fs.mkdirSync(mydir); console.log(fs.existsSync(mydir)); const myfile = mydir + '/' + 'myfile.txt'; fs.openSync(myfile, 'w'); console.log(fs.existsSync(myfile));"
true
true

PS C:\rakubrew\versions\moar-2023.02\foo> ls


    Directory: C:\rakubrew\versions\moar-2023.02\foo


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d-----         2/24/2023  12:07 PM                aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
                                                  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

PS C:\rakubrew\versions\moar-2023.02\foo> ls .\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\


    Directory: C:\rakubrew\versions\moar-2023.02\foo\aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----         2/24/2023  12:07 PM              0 myfile.txt
```
This is the node output

src/io/dirops.c
```diff
@@ -13,55 +13,25 @@
...
+                           && S_ISDIR(req.statbuf.st_mode)));
             if (!(*p = ch)) break;
         }
 
-#ifndef _WIN32
-    errno = mkdir_error; /* error reporting should use errno of latest mkdir */
-#endif
+    uv_fs_req_cleanup(&req);
+
     if (!created) return -1;
```
ugexe, line 35: Currently this `-1` gets passed to `strerror(...)` when `mkdir_p` is called later on, which says "Unknown error". I think it needs to be something like:
```diff
diff --git a/src/io/dirops.c b/src/io/dirops.c
index fba7112fe..e97c6ddb0 100644
--- a/src/io/dirops.c
+++ b/src/io/dirops.c
@@ -32,7 +32,7 @@ static int mkdir_p(MVMThreadContext *tc, char *pathname, MVMint64 mode) {

     uv_fs_req_cleanup(&req);

-    if (!created) return -1;
+    if (!created) return mkdir_error;

     return 0;
 }
@@ -44,7 +44,7 @@ void MVM_dir_mkdir(MVMThreadContext *tc, MVMString *path, MVMint64 mode) {

     if ((mkdir_error = mkdir_p(tc, pathname, mode)) != 0) {
         MVM_free(pathname);
-        MVM_exception_throw_adhoc(tc, "Failed to mkdir: %s", strerror(mkdir_error));
+        MVM_exception_throw_adhoc(tc, "Failed to mkdir: %s", uv_strerror(mkdir_error));
     }

     MVM_free(pathname);
```

ugexe: Alternatively it could probably just `errno = mkdir_error`

MasterDuke17: I think I changed it to `if (!created) return mkdir_error;` in an earlier version of this work, but that caused a spectest fail. I'll admit I didn't investigate whether it would have made sense to modify NQP and/or Rakudo to get the spectest to pass.

MasterDuke17: > Currently this `-1` gets passed to `strerror(...)` when `mkdir_p` is called later on, which says "Unknown error". I think it needs to be something like:
> 
> ```diff
> diff --git a/src/io/dirops.c b/src/io/dirops.c
> index fba7112fe..e97c6ddb0 100644
> --- a/src/io/dirops.c
> +++ b/src/io/dirops.c
> @@ -32,7 +32,7 @@ static int mkdir_p(MVMThreadContext *tc, char *pathname, MVMint64 mode) {
> 
>      uv_fs_req_cleanup(&req);
> 
> -    if (!created) return -1;
> +    if (!created) return mkdir_error;
> 
>      return 0;
>  }
> @@ -44,7 +44,7 @@ void MVM_dir_mkdir(MVMThreadContext *tc, MVMString *path, MVMint64 mode) {
> 
>      if ((mkdir_error = mkdir_p(tc, pathname, mode)) != 0) {
>          MVM_free(pathname);
> -        MVM_exception_throw_adhoc(tc, "Failed to mkdir: %s", strerror(mkdir_error));
> +        MVM_exception_throw_adhoc(tc, "Failed to mkdir: %s", uv_strerror(mkdir_error));
>      }
> 
>      MVM_free(pathname);
> ```

If you change it to return `mkdir_error` and then `uv_strerror(mkdir_error)`, what does it print? Of maybe just a `fprintf(stderr, "mkdir_error = %d, error string = '%s'\n", mkdir_error, uv_strerror(mkdir_error));` in the loop after the `uv_fs_mkdir()` call would be helpful.

ugexe: I had edited that post; it since includes the error message: `No such file or directory`

src/io/dirops.c
```diff
@@ -13,55 +13,25 @@
...
             *p  = '\0';
-#ifdef _WIN32
-            created = CreateDirectoryW(pathname, NULL);
-#else
-            created = (mkdir(pathname, mode) == 0
-                       || ( (mkdir_error = errno) == EEXIST
-                       && uv_fs_stat(NULL, &req, pathname, NULL) == 0
-                       && S_ISDIR(req.statbuf.st_mode)));
-#endif
+            created = (mkdir_error = uv_fs_mkdir(NULL, &req, pathname, mode, NULL) == 0
```
ugexe, line 26: Does this need an extra set of parens? `(mkdir_error = uv_fs_mkdir(NULL, &req, pathname, mode, NULL)) == 0`

ugexe:
So [nodejs:fs.mkdir](https://github.com/nodejs/node/blob/0597f1b673e247af40759b8778022e7a9f4d8121/lib/fs.js#L1383-L1405) does stuff like `binding.mkdir(pathModule.toNamespacedPath(path)`. It actually calls `toNamespacedPath` all over the place, which does the following:

```
// Import the path module
const path = require('path');
  
let originalPath = "C:\\Windows\\users";
console.log("Original Path:", originalPath);
  
let nameSpacedPath = path.toNamespacedPath(originalPath);
console.log("Namespaced Path:", nameSpacedPath);
```
```
Original Path: C:\Windows\users
Namespaced Path: \\?\C:\Windows\users
```

This is notable because I think the `\\?\` prefix is required for long paths with various file functions on Windows.

src/io/dirops.c
```diff
@@ -13,55 +13,25 @@
...
+                       || (mkdir_error == EEXIST
```
ugexe, line 27: This should check for `UV_EEXISTS`

patrickbkr:

PR #1746, opened 2023-03-06
Add a bunch of `*_u` stuff

MasterDuke17:
This was all in an attempt to make `raku -e 'my uint $sum = 0; for ^100_000_000 -> uint $i { $sum += $i }; say now - INIT now; say $sum;'` be as quick as when using `int` (5.6s vs 0.6s), but it doesn't actually make a difference. Creating the PR so more people can get eyes on and see if there's anything obvious I'm missing.


lizmat:
I think you would need to look at the static optimizer code in src/Perl6/Optimizer.nqp ?

MasterDuke17:
You know, I had been testing with just such a patch, but I bet it got
dropped when I pulled in a bunch of NQP and Rakudo updates. Will have to
check that later.

MasterDuke17:
> You know, I had been testing with just such a patch, but I bet it got dropped when I pulled in a bunch of NQP and Rakudo updates.

Nope, still there:
```patch
diff --git src/Perl6/Optimizer.nqp src/Perl6/Optimizer.nqp
index 8c14282da..7fc6178be 100644
--- src/Perl6/Optimizer.nqp
+++ src/Perl6/Optimizer.nqp
@@ -2521,6 +2521,18 @@ class Perl6::Optimizer {
             ]
         }
 
+        # my uint $foo
+        elsif nqp::istype($node, QAST::Var)
+          && nqp::objprimspec($node.returns) == 10 {
+            return [$extra
+              ?? QAST::Op.new( :op<add_u>, :returns($node.returns),
+                   $node,
+                   QAST::IVal.new( :value($extra) )
+                 )
+              !! $node
+            ]
+        }
+
         # No way we can make this faster
         []
     }
```

src/6model/reprs/P6opaque.c
```diff
@@ -2021,6 +2111,68 @@ static void spesh(MVMThreadContext *tc, MVMSTable *st, MVMSpeshGraph *g, MVMSpes
...
+                     * range integer, whereas we need to handle that case here.
+                     */
+                    void * data = OBJECT_BODY(facts->value.o);
+                    data = MVM_p6opaque_real_data(tc, data);
+                    MVMP6bigintBody *body = (MVMP6bigintBody *)((char *)data + repr_data->attribute_offsets[repr_data->unbox_uint_slot]);
+
+                    if (MVM_BIGINT_IS_BIG(body)) {
+                        mp_int *i = body->u.bigint;
+                        const int bits = mp_count_bits(i);
+                        if (bits <= 63) {
```
niner, line 2137: Should probably be `bits <= 64` for uints?

src/6model/reprs/P6opaque.c
```diff
@@ -2021,6 +2111,68 @@ static void spesh(MVMThreadContext *tc, MVMSTable *st, MVMSpeshGraph *g, MVMSpes
...
+                            value = mp_get_mag_ull(i);
+                            have_value = 1;
+                        }
+                        //else if (bits == 64
```
niner, line 2141: Don't need the commented code anymore

MasterDuke17:
It looks like maybe the fails are unrelated to this PR?

MasterDuke17:
@niner ping

niner (approved):

lizmat:
Am about to merge this: @MasterDuke17 should I do a merge commit or squash it into a single commit?

MasterDuke17:
> Am about to merge this: @MasterDuke17 should I do a merge commit or squash it into a single commit?

Hmm. I’d prefer a merge for bisectibility. I wouldn’t mind cleaning up some of the commits, but I’m away from my computer for a week and might not get a chance before then. If you want to merge go ahead, otherwise I’ll clean things up and merge when I get a chance.

PR #1753, opened 2023-05-26
callercode no longer prevents inlining

timo:
assuming that an inliner's curcode is what an inlinee's callercode
is supposed to return, we can just turn callercode into
curcode when inlining. There is already a guard that prevents
an inlining block from immediately inlining another so we can't for
example by accident skip over one level of
code objects AFAICT.

here's a simple script that tries to provoke a mistake from this:

```
use nqp;

my $idval = 0;
my $other-idval = 0;

my $number-of-changes = 0;

sub update-and-say($new-idval) {
    if $idval != $new-idval {
        $idval = $new-idval;
        say "outer callercode results in: $idval";
        $number-of-changes++;
    }
}

sub update-and-say-second($new-idval) {
    if $other-idval != $new-idval {
        $other-idval = $new-idval;
        say "inner callercode results in: $other-idval";
        $number-of-changes++;
    }
}

sub second-level-inline() {
    update-and-say-second(nqp::objectid(nqp::callercode()));
}

sub to-inline() {
    my $new-idval = nqp::objectid(nqp::callercode());
    update-and-say($new-idval);
    second-level-inline();
}

sub inliner() {
    to-inline()
}

for ^100000 {
    inliner;
}

if $number-of-changes != 2 {
    die "there should have been exactly two values in total for the callercode results";
}
```

### DIFF
--- a/src/core/oplist
+++ b/src/core/oplist
@@ -476,7 +476,7 @@ ctxouter            w(obj) r(obj) :pure
 ctxcaller           w(obj) r(obj) :pure
 ctxlexpad           w(obj) r(obj) :pure
 curcode             w(obj) :pure
-callercode          w(obj) :pure :noinline
+callercode          w(obj) :pure
 add_I               w(obj) r(obj) r(obj) r(obj) :pure
 sub_I               w(obj) r(obj) r(obj) r(obj) :pure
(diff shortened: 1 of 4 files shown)


niner (approved):
Looks reasonable. Can you please add some reference to this situation to the code that prevents multi-level inlining, so we know why we can't remove it?

PR #1757, opened 2023-06-04
Fix Windows build

patrickbkr:
In the libuv bump to 1.45.0 there was a change introducing more libs to link to:

https://github.com/libuv/libuv/commit/748d894e82abcdfff7429cf745003e182c47f163

The libs that commit introduced into the libuv build system are:

dbghelp, ole32, uuid

Looking at the commit it seems no symbol in uuid is used. So I left it out in this commit. MoarVM builds, so I guess it's fine.


lizmat:
Should we release a 2023.05.01 with this fix?

patrickbkr:
MasterDuke is working on getting MinGW working as well. We could wait a bit longer I guess.

PR #1774, opened 2023-11-11
Enable c11 atomics on msvc 17.8+

ugexe:
Starting with MSVC 17.8 Preview 2, c11 atomics can be used to build with a msvc toolchain. To get c11 atomics to work we need to do a few things:

* Pass the /experimental:c11atomics flag until msvc implements locking atomics
* Use c17 instead of c++latest, as the microsoft C11 atomics ops implementation doesn't work if it is in _cplusplus mode
* Fix a couple spots where the compiler complained, presumably because of the change from c++latest to c17

I don't know how well this works as an unrelated failure prevents me from testing it. But it does compile now  with --c11-atomics when using the latest preview build of MSVC.

See:
https://devblogs.microsoft.com/cppblog/c11-threads-in-visual-studio-2022-version-17-8-preview-2/

### DIFF
--- a/build/setup.pm
+++ b/build/setup.pm
@@ -419,7 +419,7 @@ our %COMPILERS = (
         ld => 'link',
         as => 'ml64',
 
-        ccmiscflags  => '/nologo /MT /std:c++latest',
+        ccmiscflags  => '/nologo /MT /std:c17 /experimental:c11atomics',
         ccwarnflags  => '',
         ccoptiflags  => '/Ox /GL /DNDEBUG',
         ccdebugflags => '/Zi',
--- a/build/setup.pm
+++ b/build/setup.pm
@@ -419,7 +419,10 @@ our %COMPILERS = (
         ld => 'link',
         as => 'ml64',
 
-        ccmiscflags  => '/nologo /MT /std:c17 /experimental:c11atomics',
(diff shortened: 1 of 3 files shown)


niner (approved):

build/setup.pm
```diff
@@ -419,7 +419,7 @@ our %COMPILERS = (
         ld => 'link',
         as => 'ml64',
 
-        ccmiscflags  => '/nologo /MT /std:c++latest',
+        ccmiscflags  => '/nologo /MT /std:c17 /experimental:c11atomics',
```
ugexe, line 422: It looks like `/experimental:c11atomics` causes warnings for older MSVC about not recognizing that flag. One option would be to use `/wd9002` to disable the warning, although this approach will suppress the warning for all unrecognized flags (not just `/experimental:c11atomics`)

ugexe: Hmm, looking at the CI for https://github.com/MoarVM/MoarVM/pull/1774/commits/7c2f842adbef354ca619955fc52a798c2c3853d1 I guess `/wd9002` isn't recognized either and just results in an additional warning

ugexe: I'm closing this in favor of https://github.com/MoarVM/MoarVM/pull/1777 which leaves out the `/experimental:c11atomics` flag, but makes all the other changes required to use c11 atomics on the msvc toolchain.

PR #1776, opened 2023-11-12
Support long paths on Windows

ugexe:
On Windows if you want path parts longer than MAX_PATH then you need to use DOS paths, i.e. absolute paths prefixed with `\\?\`. Additionally, using DOS paths does not work with / so those slashes need to be normalized when prefixing paths with `\\?\`. This adds the aforementioned prefix and normalization on Windows, allowing long paths to be used on that OS.


src/io/dirops.c
```diff
@@ -33,79 +33,33 @@ static char * UnicodeToUTF8(const wchar_t *str)
...
         if (!*p || IS_SLASH(*p)) {
             ch = *p;
             *p  = '\0';
-#ifdef _WIN32
-            created = CreateDirectoryW(pathname, NULL);
-#else
             created = (mkdir(pathname, mode) == 0
                        || ( (mkdir_error = errno) == EEXIST
                        && uv_fs_stat(NULL, &req, pathname, NULL) == 0
-                       && S_ISDIR(req.statbuf.st_mode)));
```
ugexe, line 57: I deleted the `S_ISDIR` line because it doesn't seem to be available on Windows as-is. Maybe it needs a header included?

src/io/dirops.c
```diff
@@ -33,79 +33,33 @@ static char * UnicodeToUTF8(const wchar_t *str)
...
     if (!created) return -1;
 
     return 0;
 }
 
 /* Create a directory recursively. */
 void MVM_dir_mkdir(MVMThreadContext *tc, MVMString *path, MVMint64 mode) {
-    char * const pathname = MVM_string_utf8_c8_encode_C_string(tc, path);
-
-#ifdef _WIN32
```
ugexe, line 74: We delete all of this because it is trying to do its own path prefix stuff with `\\?\`, and doesn't work correctly when `str_len > MAX_PATH` and the path is already prefixed with `\\?\`

src/io/fileops.c
```diff
@@ -350,7 +350,7 @@ MVMString * MVM_file_in_libpath(MVMThreadContext *tc, MVMString *orig) {
     if (lib_path) {
         /* We actually have a lib_path to consider. See if the filename is
          * absolute (XXX wants a platform abstraction, and doing better). */
-        char *orig_cstr = MVM_string_utf8_c8_encode_C_string(tc, orig);
+        char *orig_cstr = MVM_process_path(tc, orig);
         int  absolute   = orig_cstr[0] == '/' || orig_cstr[0] == '\\' ||
```
ugexe, line 354: We could reuse the `is_absolute_path` utility function here

src/io/io.c
```diff
@@ -434,3 +434,58 @@ int MVM_set_std_handle_to_nul(FILE *file, int fd, BOOL read, int std_handle_type
...
+        // Free the original path string and use new path.
+        MVM_free(original_path);
+        original_path = new_path;
+    }
+#endif
+
+    return original_path;
+}
+
+int is_absolute_path(const char *path) {
```
ugexe, line 474: Hmm, actually this only detects if its an absolute path on Windows, and would need to be adjusted to be reused as mentioned in one of my earlier comments. As-is it is only used on Windows, so is technically ok.

src/io/io.c
```diff
@@ -434,3 +434,57 @@ int MVM_set_std_handle_to_nul(FILE *file, int fd, BOOL read, int std_handle_type
     return 1;
 }
 #endif
+
+char* MVM_process_path(MVMThreadContext *tc, MVMString *path) {
```
ugexe, line 438: I assume I should rename this. Any suggestions?

MasterDuke17: Heh, I was thinking the same thing. Don't have any great ideas yet.

niner: MVM_platform_path?

MasterDuke17:
Any idea why the Win_MVM_reloc test keeps failing?

ugexe:
Hmmm, I’m not sure. A guess would be the absolute paths coming in from the build system aren’t cleaned (like they contain .. or ., which may not be allowed when using \\\\?\\)

ugexe:
@patrickbkr I imagine I can figure this out on my own, but off the top of your head do you have any direction regarding a relative path (or a path containing `.` or `..`) gets passed to MoarVM from the build system for relocatable builds?

patrickbkr:
> @patrickbkr I imagine I can figure this out on my own, but off the top of your head do you have any direction regarding a relative path (or a path containing `.` or `..`) gets passed to MoarVM from the build system for relocatable builds?

First thing I thought that could be related: https://github.com/rakudo/rakudo/tree/main/src/core.c/CompUnit/RepositoryRegistry.rakumod#L148

Looking at the error in the build, it seems the Rakudo Configure.pl fails calling nqp.exe to retrieve the NQP build config. That in turn fails because the compiled nqp.exe fails to find the ModuleLoader. At the time Rakudo is compiled, NQP really is a native executable (instead of a call to moar.exe + loads of args). Following that logic, this might be the place you are looking for:

https://github.com/Raku/nqp/tree/main/src/vm/moar/runner/main.c#L164

It's not the build system, but the C code of the native executable that determines the NQP_HOME path at runtime. If that turns out to be the issue (it looks like it is), then the respective Rakudo runner will have the same issue:

https://github.com/rakudo/rakudo/tree/main/src/vm/moar/runner/main.c

src/io/io.c
```diff
@@ -440,8 +440,11 @@ char* MVM_platform_path(MVMThreadContext *tc, MVMString *path) {
     char *original_path = MVM_string_utf8_c8_encode_C_string(tc, path);
 
 #ifdef _WIN32
-    // Check if the path already starts with "\\?\"
-    if (is_absolute_path(original_path) == 1 && strncmp(original_path, "\\\\?\\", 4) != 0) {
+    // Add prefix if:
+    // * It is an absolute path
+    // * It doesn't already start with \\?\
+    // * It is at least MAX_PATH in length
+    if (is_absolute_path(original_path) == 1 && strlen(original_path) > MAX_PATH && strncmp(original_path, "\\\\?\\", 4) != 0) {
```
ugexe, line 447: I suspect this `strlen(original_path) > MAX_PATH` will get CI passing. I think more generally something could check if the path contains a `.` or `..` and either not apply the prefix or to somehow resolve the path to not contain those relative path parts.

ugexe: Indeed it seems to have worked.

ugexe:
Hmmm, mingw doesn't like the switch to using `mkdir`:
```
compiling src\io\procops.o
src\io\dirops.c: In function 'mkdir_p':
src\io\dirops.c:62:24: error: too many arguments to function 'mkdir'
             created = (mkdir(pathname, mode) == 0
                        ^~~~~
In file included from C:/mingw64/x86_64-w64-mingw32/include/fcntl.h:8,
                 from 3rdparty\libuv\include/uv/win.h:60,
                 from 3rdparty\libuv\include/uv.h:69,
                 from src/moar.h:41,
                 from src\io\dirops.c:1:
C:/mingw64/x86_64-w64-mingw32/include/io.h:280:15: note: declared here
   int __cdecl mkdir (const char *) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
               ^~~~~
gmake: *** [Makefile:676: src\io\dirops.o] Error 1
gmake: *** Waiting for unfinished jobs....
##[error]PowerShell exited with code '1'.
```
We could do something [like this](https://stackoverflow.com/a/14199731/1772220) but using [uv_fs_mkdir](http://docs.libuv.org/en/v1.x/fs.html#c.uv_fs_mkdir) is probably the better solution. https://github.com/MoarVM/MoarVM/pull/1745 makes that change, although it looks like there was still an issue with it since the CI for windows are red.

MasterDuke17:
Hasn’t mingw been broken for a while?Sent from my iPhoneOn Nov 23, 2023, at 10:49 AM, Nick Logan ***@***.***> wrote:﻿
Hmmm, mingw doesn't like the switch to using mkdir:
compiling src\io\procops.o
src\io\dirops.c: In function 'mkdir_p':
src\io\dirops.c:62:24: error: too many arguments to function 'mkdir'
             created = (mkdir(pathname, mode) == 0
                        ^~~~~
In file included from C:/mingw64/x86_64-w64-mingw32/include/fcntl.h:8,
                 from 3rdparty\libuv\include/uv/win.h:60,
                 from 3rdparty\libuv\include/uv.h:69,
                 from src/moar.h:41,
                 from src\io\dirops.c:1:
C:/mingw64/x86_64-w64-mingw32/include/io.h:280:15: note: declared here
   int __cdecl mkdir (const char *) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
               ^~~~~
gmake: *** [Makefile:676: src\io\dirops.o] Error 1
gmake: *** Waiting for unfinished jobs....
##[error]PowerShell exited with code '1'.

We could do something like this but using uv_fs_mkdir is probably the better solution. #1745 makes that change, although it looks like there was still an issue with it since the CI for windows are red.

MasterDuke17:
I kind of forgot about #1745. Maybe it should be combine with this one? Or rebased on top of this to see how it does then?Sent from my iPhoneOn Nov 23, 2023, at 10:49 AM, Nick Logan ***@***.***> wrote:﻿
Hmmm, mingw doesn't like the switch to using mkdir:
compiling src\io\procops.o
src\io\dirops.c: In function 'mkdir_p':
src\io\dirops.c:62:24: error: too many arguments to function 'mkdir'
             created = (mkdir(pathname, mode) == 0
                        ^~~~~
In file included from C:/mingw64/x86_64-w64-mingw32/include/fcntl.h:8,
                 from 3rdparty\libuv\include/uv/win.h:60,
                 from 3rdparty\libuv\include/uv.h:69,
                 from src/moar.h:41,
                 from src\io\dirops.c:1:
C:/mingw64/x86_64-w64-mingw32/include/io.h:280:15: note: declared here
   int __cdecl mkdir (const char *) __MINGW_ATTRIB_DEPRECATED_MSVC2005;
               ^~~~~
gmake: *** [Makefile:676: src\io\dirops.o] Error 1
gmake: *** Waiting for unfinished jobs....
##[error]PowerShell exited with code '1'.

We could do something like this but using uv_fs_mkdir is probably the better solution. #1745 makes that change, although it looks like there was still an issue with it since the CI for windows are red.

lizmat:
Perhaps this should be merged despite its mingw failures?

lizmat:
ping @niner

lizmat:
> Perhaps this should be merged despite its mingw failures?

Will merge after the 2024.10 release unless someone stops me.

raiph:
@lizmat 

I'm guessing you spotted the comment below by @MasterDuke17 and/or concluded that they would follow up before you merged, but just in case:

> I kind of forgot about [#1745](https://github.com/MoarVM/MoarVM/pull/1745). Maybe it should be combine with this one? Or rebased on top of this to see how it does then?

> We could do something like **this** but using `uv_fs_mkdir` is probably the better solution. [#1745](https://github.com/MoarVM/MoarVM/pull/1745) makes that change

(I presume "this" meant this PR, and that their last guidance is that the other PR is "probably the better solution".)

lizmat:
@raiph in that light, closing this PR

ugexe:
#1745 has little to nothing to do with this PR other than being additive. I'm confused at why this would be closed - did anyone actually look at the changes between these two PRs?

lizmat:
@ugexe sorry!

patrickbkr:

PR #1777, opened 2023-11-12
Prepare for msvc toolchain c11 atomics

ugexe:
Visual Studio 2022 version 17.8 Preview 2 supports c11 atomics when using the `/experimental:c11atomics` flag. However, the microsoft c11 atomics implementation does not work in _cplusplus mode. This updates the build system to use `c17` instead of `c++latest`, which allows MoarVM to compile with microsoft c11 atomics (assuming it is available). This also updates a Windows macro that worked using `c++latest` but not `c17` to use the same macro as everything else.

### DIFF
--- a/build/setup.pm
+++ b/build/setup.pm
@@ -419,7 +419,7 @@ our %COMPILERS = (
         ld => 'link',
         as => 'ml64',
 
-        ccmiscflags  => '/nologo /MT /std:c++latest',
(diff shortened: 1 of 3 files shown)


niner (approved):

ugexe:
fwiw I don't have merge permissions to the MoarVM repository

PR #1780, opened 2023-11-27
Convert all dir functions to just use libuv with windows long path support

MasterDuke17:
This is #1745 + #1776 to see what the combination looks like in CI.


ugexe:
The path doesn't start with `\\?\` because of https://github.com/MoarVM/MoarVM/pull/1780/commits/ee63bb3b67600b23a62c16c14b4d8ab1df707ca4

ugexe:
> Failed to remove the directory 'C:\Users\VSSADM~1\AppData\Local\Temp\rakudo-precompTXX99GWZUNWDWHT3Z9CPGG99\34310A559C756907FABDDC38457C1B3DCE353841': Failed to rmdir: directory not empty

Hmm I'm not sure what that error is about. My initial guess would be that a file or directory isn't getting deleted (maybe because a handle is left open?) but I'm sure you already suspected that.

ugexe:
It seems to be caused by https://github.com/rakudo/rakudo/blob/c3e68caa1b84c825b9ba4aff91602168ab8645be/t/02-rakudo/reproducible-builds.t#L35 (i.e. the code in the test itself, not something in rakudo). I wonder if the test would pass if it was rerun a couple of times... and if that works I wonder if we need to add another 'retry filesystem operation on windows' thing we added elsewhere in the past.

MasterDuke17:
> edit: hmmm, but the reloc build has the same error which suggest retrying the test probably wouldn't help...

Well, maybe it needs a couple retries.... I'll restart those jobs to see if another run works.

MasterDuke17:
> I wanted to see debug output added to [rakudo/rakudo@`7347058`/src/core.c/CompUnit/PrecompilationStore/FileSystem.rakumod#L230C1-L237](https://github.com/rakudo/rakudo/blob/73470580c9286b8257be9175fd8bff57f3b14bb4/src/core.c/CompUnit/PrecompilationStore/FileSystem.rakumod#L230C1-L237) so we could see what is there initially, what actually does get deleted, etc. That would help pinpoint where in MoarVM to look at harder at least...

I think it might be possible to get CI to use a different MoarVM and/or Rakudo repo. So create a Rakudo PR that adds debugging info to that function, and then also changes the Rakudo CI workflow to use my MoarVM repo on this branch (or vice versa, add a Rakudo branch that adds the debugging and then add a change to this PR to point there).

src/io/dirops.c
```diff
@@ -207,163 +124,82 @@ static const MVMIOOps op_table = {
...
-            MVM_exception_throw_adhoc(tc, "Failed to open dir: %s", strerror(opendir_error));
-        }
+    char * const dir_name = MVM_platform_path(tc, dirname);
+    opendir_error = uv_fs_opendir(NULL, &req, dir_name, NULL);
+    MVM_free(dir_name);
 
-        data->dir_handle = dir_handle;
+    if (opendir_error) {
+        uv_fs_req_cleanup(&req);
+        MVM_exception_throw_adhoc(tc, "Failed to open dir: %s", strerror(opendir_error));
```
ugexe, line 152: Maybe `uv_strerror(...)` would give a different error message than `Unknown error` for this:
```
"let's just try the delete loop again"
IO::Path.new("C:\\Users\\VSSADM~1\\AppData\\Local\\Temp\\rakudo-precompEN2LSWRQH2TF6XK1L3YXLUJGO\\DD83F90085EDBC48B67BFCC7C5582499ED966F1C\\6B", :SPEC(IO::Spec::Win32), :CWD("D:\\a\\1\\rakudo"))
Failed to get the directory contents of 'C:\Users\VSSADM~1\AppData\Local\Temp\rakudo-precompEN2LSWRQH2TF6XK1L3YXLUJGO\DD83F90085EDBC48B67BFCC7C5582499ED966F1C\6B': Failed to open dir: Unknown error
```

MasterDuke17: The error with `uv_strerror` is 'operation not permitted'. https://dev.azure.com/Rakudo/rakudo/_build/results?buildId=5074&view=logs&j=70f6b8a1-f3cf-5727-bd79-d08ddb1caf67&t=f59a17df-92a9-5d7f-18bb-7f81f1c720a8&l=1948

src/io/dirops.c
```diff
@@ -66,6 +66,18 @@ void MVM_dir_rmdir(MVMThreadContext *tc, MVMString *path) {
     char * const pathname = MVM_platform_path(tc, path);
     uv_fs_t req;
     int rmdir_error = uv_fs_rmdir(NULL, &req, pathname, NULL);
+#ifdef _WIN32
+    if (rmdir_error == ENOTEMPTY) {
```
ugexe, line 70: It probably doesn't matter, but should the errors check for e.g. `UV_ENOTEMPTY`?

src/io/dirops.c
```diff
@@ -66,6 +66,18 @@ void MVM_dir_rmdir(MVMThreadContext *tc, MVMString *path) {
...
+        uv_fs_req_cleanup(&req);
+        sleep(0.1);
+        rmdir_error = uv_fs_rmdir(NULL, &req, pathname, NULL);
+        if (rmdir_error == ENOENT) {
+            // Did actually get deleted before we tried again.
+            uv_fs_req_cleanup(&req);
+            return;
```
ugexe, line 77: Should we add a `MVM_free(pathname);` here?

ugexe:
Hmm, it is definitely confusing that having the sleep in MoarVM seems to not fix the issue that putting the sleep in Rakudo does 🤔

MasterDuke17:
> Hmm, it is definitely confusing that having the sleep in MoarVM seems to not fix the issue that putting the sleep in Rakudo does 🤔

However, the first sleep added in rakudo was *before* the rmdir, not after the first call to it. I'll see if that changes anything.

ugexe:
One hail mary I can think of would be updating the deletion code to hopefully ensure that `.dir` isn't still open when deleting:
```
    method delete-by-compiler(CompUnit::PrecompilationId:D $compiler-id) {
         my $compiler-dir := self.prefix.add($compiler-id);
         for $compiler-dir.dir.list -> $subdir {
             .unlink for $subdir.dir.list;
             $subdir.rmdir;
         }
         $compiler-dir.rmdir;
    }
```
which changes `.dir` to `.dir.list`

ugexe:
fwiw I cut and paste all the changes one by one into my franken-VM on 2023.04 and the reproducible build test passes every time without the need to retry.

MasterDuke17:
Think it’s because of the libuv bump between then and now?Sent from my iPhoneOn Dec 6, 2023, at 7:17 PM, Nick Logan ***@***.***> wrote:﻿
fwiw I cut and paste all the changes one by one into my franken-VM on 2023.04 and the reproducible build test passes every time without the need to retry.

ugexe:
Im not sure. Maybe in the CI test we can cd to libuv, checkout an older version, and do nmake install to test that. Or maybe @coke can try running that test 😬

ugexe:
Just to be clear: my VM is running libuv v1.44.1, so just reverting v1.47.0 wouldn't put it at the same version. Looking at the changelog though I'm not sure the libuv version would be affecting this.

lizmat:
ping @niner

patrickbkr:

PR #1782, opened 2024-01-05
Don't lose return value with LEAVE phasers

patrickbkr:
Fixes https://github.com/rakudo/rakudo/issues/4521.

When using &return an exception is thrown and the return value saved in `tc->last_payload`. LEAVE phasers interupt the stack unwind and call the LEAVE code block. Code in there can and often does make use of `tc->last_payload` via &return or other means. This overwrites the original value which is then lost. Fix by saving / restoring the original value in the special return stack frame. That frame is added to the call stack before calling the LEAVE code block and the value can be restored once we are back to that stack frame.

I suspect more needs to be done to make the JIT also recognize this.
@MasterDuke17 Any pointers to where I could start looking into the JIT?

### DIFF
--- a/src/core/frame.c
+++ b/src/core/frame.c
@@ -953,6 +953,7 @@ typedef struct {
     MVMuint8  *abs_addr;
     MVMuint32  rel_addr;
     void      *jit_return_label;
+    MVMObject *payload;
 } MVMUnwindData;
 static void mark_unwind_data(MVMThreadContext *tc, void *sr_data, MVMGCWorklist *worklist) {
     MVMUnwindData *ud  = (MVMUnwindData *)sr_data;
@@ -964,6 +965,7 @@ static void continue_unwind(MVMThreadContext *tc, void *sr_data) {
     MVMuint8 *abs_addr = ud->abs_addr;
     MVMuint32 rel_addr = ud->rel_addr;
     void *jit_return_label = ud->jit_return_label;
+    tc->last_payload = ud->payload;
     MVM_frame_unwind_to(tc, frame, abs_addr, rel_addr, NULL, jit_return_label);
 }
 void MVM_frame_unwind_to(MVMThreadContext *tc, MVMFrame *frame, MVMuint8 *abs_addr,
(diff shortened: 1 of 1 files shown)


patrickbkr:
Grepping for `last_payload` gives only a single JITed opcode: `lastexpayload`. That should be unimpaired by the changes in this PR.

MasterDuke17:
> I suspect more needs to be done to make the JIT also recognize this.
> @MasterDuke17 Any pointers to where I could start looking into the JIT?

> Grepping for last_payload gives only a single JITed opcode: lastexpayload. That should be unimpaired by the changes in this PR.

I'm not particularly well-versed in the frame-related code, but a quick grep through the codebase seems to confirm your finding that the JIT doesn't really know/do much about this stuff, so nothing needs to be done for this PR.

niner:
What happens with this if the LEAVE phaser does an explicit `return`?

patrickbkr:
> What happens with this if the LEAVE phaser does an explicit `return`?

Interesting question. I'll explore this a bit. We do need tests for this anyways. Intuitively I'd say this should be a compile time exception.

2colours:
@vrurg @Voldenet (idk if that's the correct handle on Github)

Regarding semantics at least, we all seemed to agree that instead of banning explicit returns, there is a rather obvious useful behavior for them: they should overwrite the value set by the previous return (whether implicit or explicit) of the same scope. This means that the behavior should be as if the LEAVE phaser was the last thing in the sub, not the first thing after the sub.

2colours:
Actually, as much as I understand, this means that the issue being addressed might not be an issue in the first place. Surely the LEAVE phaser can't (and shouldn't) run before the `return` of the same sub, if it has one. This means either it has to be completely deprived from sending a return value - perhaps this is what you are aiming at - or only the value it returns can be the final return value.

For me it seems the behavior should be much like CATCH: hijacking the previous value and possibly "rereturning" it. But this is way into design territory actually.

patrickbkr:
Related: https://github.com/MoarVM/MoarVM/issues/1784

patrickbkr:
> > What happens with this if the LEAVE phaser does an explicit `return`?
> 
> Interesting question. I'll explore this a bit. We do need tests for this anyways. Intuitively I'd say this should be a compile time exception.

This is explored over in https://github.com/Raku/problem-solving/issues/417. I'd like to consider that a separate issue. Can this PR thus be merged?

Who's best to ping? @MasterDuke17 @niner

niner (approved):

PR #1785, opened 2024-01-16
Fix returning from LEAVE to surrounding scope

### LINKED ISSUE #1784: `return` in `LEAVE` misses handler in directly surounding scope
```
sub s() {
    LEAVE return 5;
    return 7;
}
s()
```
gives

> Attempt to return outside of any Routine
>   in sub s at leave-bug.raku line 2
>   in block <unit> at leave-bug.raku line 5

I've dug into this a bit.
From what I've understood, the `return 7` starts an unwind. The exit handler of `s()` is found and called via `MVM_frame_dispatch_from_c`. That in turn sets `cur_frame->return_address` to `*(tc->interp_cur_op)` which still points to someplace in `&return`. The `return 5` then searches for an applicable handler in the call stack, but misses the one of `s()` because `cur_frame->return_address` is then usually outside of the handlers corresponding area.

The best idea I could come up with to fix this is setting a sane `cur_frame->return_address` before calling the exit handler. But I don't know what a sane value could be. The value will not be used to return to (we already have a handler we are unwinding to), but needs to meet  `pc >= fh->start_offset && pc <= fh->end_offset` (that's in [exceptions.c search_frame_handlers_lex](https://github.com/MoarVM/MoarVM/blob/main/src/core/exceptions.c#L251)). I think the `LEAVE` doesn't  even have a representation in the bytecode.

Hm. It's hacky, but something like the following could do the trick:
```
if (cur_frame == fh->frame) {
    cur_frame->return_address = fh->start_offset;
}
```
I do hope for a better idea though.

patrickbkr:
When a LEAVE phaser is run during an unwind, there is no valid return_address set in the LEAVE phasers caller frame. (It return_address is actually set to cur_op which is still the one that started the unwind in some unrelated frame.) Thus when a `return` in a LEAVE happens, the handler of the outer frame is missed, because `search_frame_handlers_lex()` validates that the `return_address` lies in the handlers area. Luckily there is a flag set on the LEAVE's outer frame:
MVM_FRAME_FLAG_EXIT_HAND_RUN. We can simply check for that flag and ignore the frame handlers area. This is fine to do, because frames that have an exit handler attached can not be inlined. This fixes:

    sub s() {
        LEAVE return 5;
        return 7;
    }
    s()

which before this fix printed

    Attempt to return outside of any Routine
    in sub s at leave-bug.raku line 2
    in block at leave-bug.raku line 5

Fixes #1784

### DIFF
--- a/src/core/exceptions.c
+++ b/src/core/exceptions.c
@@ -175,13 +175,28 @@ static MVMint32 search_frame_handlers_lex(MVMThreadContext *tc, MVMFrame *f,
     MVMuint32 i;
     MVMuint32 skipping = *skip_first_inlinee;
     MVMFrameHandler *fhs = MVM_frame_effective_handlers(f);
+
+    /* When a LEAVE phaser is run during an unwind, there is no valid
+     * return_address set in the LEAVE phasers outer frame. (It's
+     * return_address is actually set to interp_cur_op, which is still the one
+     * that started the unwind in some unrelated frame.) Thus when a `return`
+     * in a LEAVE happens, the handler of the outer frame is missed, because
+     * `search_frame_handlers_lex()` validates that the `return_address` lies
+     * in the handlers area. Luckily there is a flag set on the LEAVE's outer
+     * frame: MVM_FRAME_FLAG_EXIT_HAND_RUN. We can simply check for that flag
+     * and ignore the frame handlers area. This is fine to do, because frames
+     * that have an exit handler attached can not be inlined.
+     */
+    MVMint32 fehr = f->flags == MVM_FRAME_FLAG_EXIT_HAND_RUN;
+
     if (f->spesh_cand && f->spesh_cand->body.jitcode && f->jit_entry_label) {
         MVMJitCode *jitcode = f->spesh_cand->body.jitcode;
         void *current_position = MVM_jit_code_get_current_position(tc, jitcode, f);
         MVMJitHandler    *jhs = jitcode->handlers;
-        for (i = MVM_jit_code_get_active_handlers(tc, jitcode, current_position, 0);
+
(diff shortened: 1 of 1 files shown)


patrickbkr:
I believe this works reliably. So I guess it's ready for review.

patrickbkr:
The general question of what `return` in `LEAVE` should do is explored over in https://github.com/Raku/problem-solving/issues/417.

Technically this PR is unreladed to that issue. Can this PR thus be merged?

Again, who's best to ping? @MasterDuke17 @niner

src/core/exceptions.c
```diff
@@ -248,8 +248,8 @@ static MVMint32 search_frame_handlers_lex(MVMThreadContext *tc, MVMFrame *f,
             }
             if (skipping || !handler_can_handle(f, fh, cat, payload))
                 continue;
-            if (pc >= fh->start_offset &&
-                    pc <= fh->end_offset &&
+            if (((pc >= fh->start_offset && pc <= fh->end_offset) ||
+                    f->flags == MVM_FRAME_FLAG_EXIT_HAND_RUN) &&
```
niner, line 252: While you described the issue well in the commit message, I think this warrants also a comment in the code as it's quite a bit surprising.

Also what about the JITed case? Why don't we need any special handling there?

patrickbkr: My memory started to get fuzzy already, I'll have to look at this again.

patrickbkr: I've read into this again. I see no reason, why this shouldn't also affect JIT compiled frames. So, while I didn't manage to reproduce the error with only the non-jitted code path fixed, I've now added the respective fix to the jitted code path as well.

I've also added a respective comment to the code.

PR #1786, opened 2024-01-21
Implement return prioritization

patrickbkr:
Calling &return triggers a stack unwind which can cause code (e.g. in `LEAVE` phasers) to run. That code can also call `&return`. In such situations one usually wants the return to win that unwinds more of the stack. This is the behavior implemented by this change. Previously the later return always won.

Related to Raku/problem-solving#417. As that issue hasn't reached a conclustion yet, this PR shouldn't be merged either.

### DIFF
--- a/src/core/exceptions.c
+++ b/src/core/exceptions.c
@@ -331,6 +331,13 @@ static LocatedHandler search_for_handler_from(MVMThreadContext *tc, MVMFrame *f,
  * an exception object already, it will be used; NULL can be passed if there
  * is not one, meaning it will be created if needed (based on the category
  * parameter; if ex_obj is passed, the category is not used). */
+static void cleanup_control_handler(MVMThreadContext *tc) {
+    if (tc->active_control_handler) {
+        MVM_free(tc->active_control_handler);
+        tc->active_control_handler = NULL;
+    }
+
+}
 static void unwind_after_handler(MVMThreadContext *tc, void *sr_data);
 static void cleanup_active_handler(MVMThreadContext *tc, void *sr_data);
 static void run_handler(MVMThreadContext *tc, LocatedHandler lh, MVMObject *ex_obj,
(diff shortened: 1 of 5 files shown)


src/core/exceptions.c
```diff
@@ -346,13 +353,54 @@ static void run_handler(MVMThreadContext *tc, LocatedHandler lh, MVMObject *ex_o
...
+                    if (cur_frame == lh.frame) {
+                        lh.frame = tc->active_control_handler->frame;
+                        lh.handler = tc->active_control_handler->handler;
+                        lh.jit_handler = tc->active_control_handler->jit_handler;
+                        tc->last_payload = tc->active_control_handler->ex_obj;
+                        frame_found = 1;
+                        break;
+                    }
+                    else if (cur_frame == tc->active_control_handler->frame) {
+                        MVM_free(tc->active_control_handler);
```
patrickbkr, line 380: Does tc->active_control_handler->ex_obj need to be freed separately? Is this a memory leak?

src/core/threadcontext.h
```diff
@@ -208,6 +208,9 @@ struct MVMThreadContext {
     /* Last payload made available in a payload-goto exception handler. */
     MVMObject *last_payload;
 
+    /* Handler the currently active CONTROL exception is unwinding to. */
+    MVMActiveHandler *active_control_handler;
```
patrickbkr, line 212: MVMActiveHandler isn't a perfect fit. We don't need MVMActiveHandler->next and we save the payload in MVMActiveHandler->ex_obj, which is a poor name. Should we just create an MVMActiveControlHandler solely for this?

src/core/exceptions.c
```diff
@@ -331,6 +331,13 @@ static LocatedHandler search_for_handler_from(MVMThreadContext *tc, MVMFrame *f,
  * an exception object already, it will be used; NULL can be passed if there
  * is not one, meaning it will be created if needed (based on the category
  * parameter; if ex_obj is passed, the category is not used). */
+static void cleanup_control_handler(MVMThreadContext *tc) {
```
patrickbkr, line 334: Is this guaranteed to be always called? What happens, when some other exception is thrown while the unwind is still in progress?

patrickbkr (commented):
This passes all spec tests. But there are some open questions that need clarification. I'm happy for someone knowledgeable to give answers.  Otherwise I'll just have to dig in more.

src/core/exceptions.c
jnthn: At least this question is an easy one: an `MVMObject *` is a pointer to an object managed by the GC, and those will be cleared up by the GC (in fact, to free it would be an error).

src/core/threadcontext.h
jnthn: Probably better to have a different structure. Also, one needs to GC mark the object.

patrickbkr:
This approach is fatally wrong. It basically entirely disables `&return` inside phasers which messes up all but the simplest code there can be.

PR #1788, opened 2024-01-27
Implement return prioritization (2nd try)

patrickbkr:
In general, if a &return was called in a routine, we don't want any other code in exit handlers to interfere with that return from returning.

It's possible for exit handlers to escape normal control flow (e.g. via a &return). In that case instead of reaching the handler the unwind targets, we would return to the frame's return_address. To make things right, we tune the unwind target frame's return address to point to the handler. Also set that frames RETURNING flag and preserve the return value, getlexpayload can then pick it up again.

JIT implementation of `lastexpayload` is TBD.

Related to Raku/problem-solving#417

### DIFF
--- a/src/core/frame.c
+++ b/src/core/frame.c
@@ -1012,6 +1012,23 @@ void MVM_frame_unwind_to(MVMThreadContext *tc, MVMFrame *frame, MVMuint8 *abs_ad
                 if (cur_frame == tc->thread_entry_frame)
                     MVM_exception_throw_adhoc(tc, "Thread entry point frame cannot have an exit handler");
 
+                /* It's possible for the exit handler to escape normal control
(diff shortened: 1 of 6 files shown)


patrickbkr:
I now have the JIT bits in place as well. I still fail to test it though (as in, I haven't managed to produce a test case that fails with the JIT bits left out).

patrickbkr:
I just found a case where this implementation fails. Thus this PR is draft again.
The return value of the original unwind is preserved, but the hijacking of the unwind via a `&return` inside a `LEAVE` still happens.
I have extended the tests in https://github.com/Raku/roast/pull/851 respectively, which this PR thus fails.

patrickbkr:
Superseded by https://github.com/MoarVM/MoarVM/pull/1812

PR #1790, opened 2024-02-11
Slightly optimize creating backtraces

MasterDuke17:
Instead of using snprintf just to convert a number into a C string and then converting that C string into an MVMString, just use MVM_coerce_u_s to convert the number directly an MVMString.

The number of instructions reported by `MVM_SPESH_BLOCKING=1 valgrind --tool=callgrind raku -e 'my $a; $a = Backtrace.new(1) for ^1_000; say $a'` decreases from ~1.375b to ~1.369b and the time reported by `raku -e 'my $a; $a = Backtrace.new(1) for ^100_000; say now - INIT now; say $a'` decreases from ~0.76s to ~0.67s.

### DIFF
--- a/src/core/exceptions.c
+++ b/src/core/exceptions.c
@@ -561,9 +561,7 @@ MVMObject * MVM_exception_backtrace(MVMThreadContext *tc, MVMObject *ex_obj) {
             MVMBytecodeAnnotation *annot = MVM_bytecode_resolve_annotation(tc, &cur_frame->static_info->body,
                                                 offset > 0 ? offset - 1 : 0);
             MVMuint32             fshi   = annot ? (MVMint32)annot->filename_string_heap_index : -1;
-            char         line_number[11] = {0}; // 11 == 10 (max # of decimal digits in an MVMuint32) + 1 (NULL terminator)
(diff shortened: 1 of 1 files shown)


MasterDuke17:
I can repro the problem in t/nqp/093-oo-ops.t, but it also happens on main, so it isn't caused by this PR.

FWIW, the problem is that `./nqp-m -e 'say(nqp::objprimunsigned(num))'` reports random values, instead of `0`, like it should (or `1` if you pass `uint`).

lizmat:
Seems to work for me on main?
```
$ ./nqp --version -e 'say(nqp::objprimunsigned(num))'
This is nqp version 2024.01-17-g6583e61c5 built on MoarVM version 2024.01
$ ./nqp -e 'say(nqp::objprimunsigned(num))' 
0
$ ./nqp -e 'say(nqp::objprimunsigned(uint))'
1
```

niner (approved):

niner:
The flaky NQP test is fixed with commit https://github.com/MoarVM/MoarVM/commit/a1a18e9842464fdc17ae85e9494fb75a002c496c

PR #1791, opened 2024-02-15
Update libuv to version 1.48.0

dogbert17:
Spectest clean under Linux Mint 20.3 (Una)


lizmat:
Changes: https://github.com/libuv/libuv/blob/master/ChangeLog

lizmat (approved):

Issue #1794, opened 2024-03-04
question / offer: build GitHub workflow based MoarVM releases


AntonOks:
@jdv @MasterDuke17 @lizmat 

Due to #1789, I wanted to assure MoarVM is still building successfully on all current [GitHub runners](https://github.com/actions/runner-images#available-images) or verify the same error as on my priv. Windows11 box shows up with one of the images... 

... so I forked MoarVM and set up a [GitHub "build MoarVM" workflow](https://github.com/AntonOks/MoarVM/blob/main/.github/workflows/build_release.yml) to test such builds and create [related release assets](https://github.com/AntonOks/MoarVM/releases).

I'm happy to see GitHub based builds do work great... and still confused why MoarVM is not building on my Win11 box...

Anyhow, I ask myself now if someone see's a benefit of this **build_release.yml** and would like to use it also here...? What do you think?

dontlaugh:
It looks good to me.

The release step is convenient, but if there's any hesitation about adopting a workflow for that in the short term, you could also just make a PR with the matrix build stuff.

PR #1796, opened 2024-03-15
build_release.yml: Add initial build workflow

AntonOks:
Add a GitHub release build workflow to assure MoarVM is successfully building:
- on all GitHub runners (ubuntu, macos, windows)
- based on the actual pushed "git tag"


AntonOks:
see also https://github.com/MoarVM/MoarVM/issues/1794

lizmat:
Thanks!

PR #1798, opened 2024-04-21
Address bug in twoway_memmem_uint32 use of memcmp (Fixes R#5550)

ab5tract:
tl;dr -- The `memcmp` call was not considering the 32-bit shape of the byte strings being passed.

More information available in the commit message.

Also included a performance improvement from the upstream FreeBSD/musl implementation of `memmem` ([further details ](https://reviews.freebsd.org/rG7dbcd06e63101d51e6a777f7315cfde794411e53)).


ab5tract:
Even though this fixes the original bug (and appears to be the correct thing to do), there are now test failures.

Investigating.

PR #1799, opened 2024-04-25
Just malloc instead of stack+malloc+memcpy when coercing int to string

MasterDuke17:
Instead of itoa() writing into a stack-allocated buffer of the max possible size and then MVM_malloc()ing a new buffer of the actual size and memcpy()ing from the stack buffer into the new buffer, alway MVM_malloc() a buffer of the max size and just use that. This means we don't use the stack as much and don't have to memcpy(). However, we are now probably wasting some memory with every coerce that doesn't hit the cache, since most coercions likely result in a smaller string than the max possible size. One obvious solution would be to MVM_realloc() down to the actual size, but that's actually slower than the original code. This does mean that unmanaged_size() will slightly underreport for the MVMStrings resulting from these coercions, but maybe the difference will be small enough not to really matter?

My test case, `MVM_SPESH_BLOCKING=1 nqp-m -e 'my str $s; my int $i := 0; my $n := nqp::time; while $i++ < 10_000_000 { $s := $i }; say(nqp::div_n(nqp::time - $n, 1000000000e0)); say($s)'` decreased from ~0.30s to ~0.27s and the number of instructions reported by callgrind (with only 1_000_000 iterations and not tracking the time) decreased from ~642.2m to ~608.7m.

I logged all coercions that didn't hit the cache during the Rakudo build and of the 8.2m, 7.8m were for length 13 (so wasting 7 bytes), and the rest were no more than 10 (so wasting 10-19 bytes).

### DIFF
--- a/src/core/coerce.c
+++ b/src/core/coerce.c
@@ -91,8 +91,6 @@ static char * i64toa_jeaiii(int64_t i, char* b) {
 /* End code */
 
 MVMString * MVM_coerce_i_s(MVMThreadContext *tc, MVMint64 i) {
-    char buffer[20];
-    int len;
     /* See if we can hit the cache. */
     int cache = 0 <= i && i < MVM_INT_TO_STR_CACHE_SIZE;
     if (cache) {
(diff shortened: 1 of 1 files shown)


lizmat:
> 7.8m were for length 13 (so wasting 7 bytes), and the rest were no more than 10 (so wasting 10-19 bytes).

So the Rakudo build would waste 7.8M * 7 = ~55MByte of memory?  Or am I missing something?

MasterDuke17:
Correct. But nine's suggestion would get rid of that, I just haven't gotten
around to implementing it yet (got nerd-sniped by in-situ-strings).

MasterDuke17:
With the most recent commit the time is the same, ~0.27s down from ~0.30s, but the instruction count doesn't decrease as much, ~642.2m down to ~620m. However, it now shouldn't waste any space.

niner (approved):

PR #1800, opened 2024-04-25
Update jeaiii's itoa() to the latest version

MasterDuke17:
With some slight modifications to convert it to C (from C++).

My test case, `MVM_SPESH_BLOCKING=1 nqp-m -e 'my str $s; my int $i := 0; my $n := nqp::time; while $i++ < 10_000_000 { $s := $i }; say(nqp::div_n(nqp::time - $n, 1000000000e0)); say($s)'` decreased from ~0.30s to ~0.29s and the number of instructions reported by callgrind (with only 1_000_000 iterations and not tracking the time) decreased from ~642.2m to ~640.3m.


lizmat:
Is this still good to go?

MasterDuke17:
> Is this still good to go?

Not quite. This seems to have uncovered a bug in GCC unsigned int handling. I think there's a workaround, but I'll need to do some more investigating.

PR #1802, opened 2024-05-09
Store short strings inside string's body

MasterDuke17:
Uses the 8 bytes that the pointer to a buffer usually takes and puts up to eight 8-bit or two 32-bit characters into it.

Started by @timo++.

The time reported by micro-benchmark `MVM_SPESH_BLOCKING=1 nqp-m -e 'my str $s; my int $i := 0; my $n := nqp::time; while $i++ < 10_000_000 { $s := $i }; say(nqp::div_n(nqp::time - $n, 1000000000e0)); say($s)'` drops from ~0.270s to ~0.145s, the number of instructions (for 1m iterations) drops from ~628m to ~555m, and the number of allocation functions called as reported by heaptrack drops from ~10m to ~252k.

However, another micro-benchmark `MVM_SPESH_BLOCKING=1 nqp-m -e 'my str $s := "ab.cd.ef.gh.ij.kl.mn.op.qr.st.uv.wx.yz"; my int $i := 0; my @a; nqp::setelems(@a, 30); my $n := nqp::time; while $i++ < 1_000_000 { @a := nqp::split(".", $s); }; say(nqp::div_n(nqp::time - $n, 1000000000e0)); say(@a[0])'` has no significant change in time reported, and the instruction count increased from ~10.2b to ~11.9b, even though the allocation functions called as reported by heaptrack drops from ~15.2m to ~2.2m.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -337,6 +337,15 @@ static void iterate_gi_into_string(MVMThreadContext *tc, MVMGraphemeIter *gi, MV
             MVM_string_gi_next_strand_rep(tc, gi);
         }
     }
+    if (result->body.num_graphs <= 8 && result->body.storage_type == MVM_STRING_GRAPHEME_8) {
+        MVMuint8 i;
+        MVMGrapheme8 *old = result->body.storage.blob_8;
+        for (i = 0; i < result->body.num_graphs; i++) {
+            result->body.storage.in_situ[i] = old[i];
+        }
+        result->body.storage_type    = MVM_STRING_IN_SITU;
+        MVM_free(old);
+    }
 }
 #define copy_strands_memcpy(BLOB_TYPE, SIZEOF_TYPE, STORAGE_TYPE) { \
     result->body.storage.BLOB_TYPE = MVM_malloc(sizeof(SIZEOF_TYPE) * MVM_string_graphs_nocheck(tc, orig)); \
@@ -763,7 +772,7 @@ MVMString * MVM_string_substring(MVMThreadContext *tc, MVMString *a, MVMint64 of
     MVMROOT(tc, a, {
         result = (MVMString *)MVM_repr_alloc_init(tc, tc->instance->VMString);
         result->body.num_graphs = end_pos - start_pos;
-        if (a->body.storage_type != MVM_STRING_STRAND) {
+        if (a->body.storage_type != MVM_STRING_STRAND && result->body.num_graphs > 8) {
             /* It's some kind of buffer. Construct a strand view into it. */
             result->body.storage_type    = MVM_STRING_STRAND;
             result->body.storage.strands = allocate_strands(tc, 1);
@@ -774,7 +783,7 @@ MVMString * MVM_string_substring(MVMThreadContext *tc, MVMString *a, MVMint64 of
             result->body.storage.strands[0].end         = end_pos;
             result->body.storage.strands[0].repetitions = 0;
         }
-        else if (a->body.num_strands == 1 && a->body.storage.strands[0].repetitions == 0) {
+        else if (a->body.num_strands == 1 && a->body.storage.strands[0].repetitions == 0 && result->body.num_graphs > 8) {
             /* Single strand string; quite possibly already a substring. We'll
              * just produce an updated view. */
             MVMStringStrand *orig_strand = &(a->body.storage.strands[0]);
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -199,8 +199,14 @@ static void turn_32bit_into_8bit_unchecked(MVMThreadContext *tc, MVMString *str)
     MVMStringIndex i;
     MVMGrapheme8 *dest_buf = NULL;
     MVMStringIndex num_graphs = MVM_string_graphs_nocheck(tc, str);
-    str->body.storage_type   = MVM_STRING_GRAPHEME_8;
-    dest_buf = str->body.storage.blob_8 = MVM_malloc(str->body.num_graphs * sizeof(MVMGrapheme8));
+    if (num_graphs <= 8) {
+        str->body.storage_type   = MVM_STRING_IN_SITU_8;
+        dest_buf = str->body.storage.in_situ_8;
+    }
+    else {
+        str->body.storage_type   = MVM_STRING_GRAPHEME_8;
+        dest_buf = str->body.storage.blob_8 = MVM_malloc(str->body.num_graphs * sizeof(MVMGrapheme8));
+    }
     MVM_VECTORIZE_LOOP
     for (i = 0; i < num_graphs; i++) {
         dest_buf[i] = old_buf[i];
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -2889,9 +2889,8 @@ MVMString * MVM_string_chr(MVMThreadContext *tc, MVMint64 cp) {
 
     s = (MVMString *)REPR(tc->instance->VMString)->allocate(tc, STABLE(tc->instance->VMString));
     if (can_fit_into_8bit(g)) {
-        s->body.storage_type       = MVM_STRING_GRAPHEME_8;
-        s->body.storage.blob_8     = MVM_malloc(sizeof(MVMGrapheme8));
-        s->body.storage.blob_8[0]  = g;
+        s->body.storage_type       = MVM_STRING_IN_SITU;
+        s->body.storage.in_situ[0] = g;
     } else {
         s->body.storage_type       = MVM_STRING_GRAPHEME_32;
         s->body.storage.blob_32    = MVM_malloc(sizeof(MVMGrapheme32));
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -2454,6 +2454,8 @@ MVMint64 MVM_string_compare(MVMThreadContext *tc, MVMString *a, MVMString *b) {
     MVMStringIndex alen, blen, i = 0, scanlen;
     MVMGraphemeIter gi_a, gi_b;
 
+    MVMuint8 a_is_eight, b_is_eight, a_is_in_situ, b_is_in_situ;
+
     MVM_string_check_arg(tc, a, "compare");
     MVM_string_check_arg(tc, b, "compare");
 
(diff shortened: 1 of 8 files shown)


MasterDuke17:
Not every single case where an MVMString is created do I check whether it could be in-situ. Some of them seemed a bit unlikely to hit in enough instances to be worth the extra checking. However, some might need metrics gathered on the resulting sizes to really know for sure whether it's worth it. And I'd appreciate any suggestions for places I've missed.

MasterDuke17:
The CI fails are all because of t/02-rakudo/reproducible-builds.t, I'm not quite sure what the problem is or how to fix it.

MasterDuke17:
> The CI fails are all because of t/02-rakudo/reproducible-builds.t, I'm not quite sure what the problem is or how to fix it.

Well, locally a re-bootstrap (and then even another just for good luck) didn't fix it.

lizmat:
FWIW, I've seen `t/02-rakudo/reproducible-builds.t` flap, but rarely.  Possibly this PR is exposing the cause of this flapper.

timo:
on the other hand, if we have such a very small substring, it's also possible to decide "we should store the string in-situ instead of a rope-based substring" ... ok i very vaguely recall that we might already be doing something similar to that?? maybe only for comb for example

src/strings/ops.c
```diff
@@ -337,6 +389,15 @@ static void iterate_gi_into_string(MVMThreadContext *tc, MVMGraphemeIter *gi, MV
...
         }
     }
+    if (result->body.num_graphs <= 2 && result->body.storage_type == MVM_STRING_GRAPHEME_32) {
+        MVMuint8 i;
+        MVMGrapheme32 *old = result->body.storage.blob_32;
+        for (i = 0; i < result->body.num_graphs; i++) {
+            result->body.storage.in_situ_32[i] = old[i];
+        }
+        result->body.storage_type    = MVM_STRING_IN_SITU_32;
+        MVM_free(old);
```
timo, line 399: it would be interesting to see how often this triggers; should we be able to know `result->body.num_graphs` near the start of the function, and could we then always never go with storage type string_grapheme_32 throughout the body of the function?

MasterDuke17: You're right, we could probably do it right around https://github.com/MoarVM/MoarVM/pull/1802/files#diff-bc76be5614df2473912916db269fc246c79efbc83a0a065ee8b785aa3c68d040R330-R333

MasterDuke17:
> for the second benchmark you posted, it's probably important to note that `@a := nqp::split(...)` will create a new array each go-around, which is probably a big chunk of allocations, and i'd wager that the split will create ropes for all the entries rather than use any real storage. it would be kind of fiddly, but possible, to have in-situ-storage for "non-repeated substring of just a single other string" ropes? with a specific storage type, the pointer can point at the other string, the spot for the strand count can be re-used for a start offset, the num_graphs would naturally double as the "how long is the substring" count. then the split example should be able to reduce its allocations.

> on the other hand, if we have such a very small substring, it's also possible to decide "we should store the string in-situ instead of a rope-based substring" ... ok i very vaguely recall that we might already be doing something similar to that?? maybe only for comb for example

The second thing is (should be) happening now, though kind of in a round-about way that might not be more efficient. Here https://github.com/MoarVM/MoarVM/pull/1802/files#diff-bc76be5614df2473912916db269fc246c79efbc83a0a065ee8b785aa3c68d040R827 in `MVM_string_substring` I added the `&& result->body.num_graphs > 8`, so if it's <= 8 then it goes into the branch that just does `iterate_gi_into_string()`, which should put stuff directly into an in-situ string if it can fit.

> another thing from your benchmark code that i can't determine whether it's useful or not is to have a variant of nqp::split that takes an existing array and puts its data into it, instead of creating a new array to return. this would want a survey of the whole nqp and rakudo code bases to find spots where we can benefit. i don't expect it's very many, though

ISTR some conversation with you and/or lizmat about this, maybe for storing profile information, or rusage? Ah, looks like it was just done directly in the nqp::getrusage op. https://github.com/Raku/nqp/blob/main/docs/ops.markdown#getrusage

timo:
```
From e215e74bba2853cde0c9ddd53b1a1a424d860674 Mon Sep 17 00:00:00 2001
From: Timo Paulssen <timonator@perpetuum-immobile.de>
Date: Fri, 5 Jul 2024 22:31:16 +0200
Subject: [PATCH] fix comparison of strings involving in situ strings

---
 src/strings/ops.c | 47 ++++++++++++++++++++++++++---------------------
 1 file changed, 26 insertions(+), 21 deletions(-)

diff --git a/src/strings/ops.c b/src/strings/ops.c
index a7b00465b..32e864152 100644
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -2454,6 +2454,8 @@ MVMint64 MVM_string_compare(MVMThreadContext *tc, MVMString *a, MVMString *b) {
     MVMStringIndex alen, blen, i = 0, scanlen;
     MVMGraphemeIter gi_a, gi_b;
 
+    MVMuint8 a_is_eight, b_is_eight, a_is_in_situ, b_is_in_situ;
+
     MVM_string_check_arg(tc, a, "compare");
     MVM_string_check_arg(tc, b, "compare");
 
@@ -2468,35 +2470,26 @@ MVMint64 MVM_string_compare(MVMThreadContext *tc, MVMString *a, MVMString *b) {
     /* Otherwise, need to scan them. */
     scanlen = blen < alen ? blen : alen;
 
+    a_is_eight = a->body.storage_type == MVM_STRING_GRAPHEME_8 || a->body.storage_type == MVM_STRING_GRAPHEME_ASCII || a->body.storage_type == MVM_STRING_IN_SITU_8;
+    b_is_eight = b->body.storage_type == MVM_STRING_GRAPHEME_8 || b->body.storage_type == MVM_STRING_GRAPHEME_ASCII || b->body.storage_type == MVM_STRING_IN_SITU_8;
+
+    a_is_in_situ = a->body.storage_type == MVM_STRING_IN_SITU_8 || a->body.storage_type == MVM_STRING_IN_SITU_32;
+    b_is_in_situ = b->body.storage_type == MVM_STRING_IN_SITU_8 || b->body.storage_type == MVM_STRING_IN_SITU_32;
+
     /* Short circuit a case where the other conditionals won't speed it up */
     if (a->body.storage_type == MVM_STRING_STRAND || b->body.storage_type == MVM_STRING_STRAND) {
         /* do nothing */
     }
-    else if ((a->body.storage_type == MVM_STRING_GRAPHEME_8 || a->body.storage_type == MVM_STRING_GRAPHEME_ASCII)
-          && (b->body.storage_type == MVM_STRING_GRAPHEME_8 || b->body.storage_type == MVM_STRING_GRAPHEME_ASCII)) {
-        MVMGrapheme8  *a_blob8 = a->body.storage.blob_8;
-        MVMGrapheme8  *b_blob8 = b->body.storage.blob_8;
+    else if (a_is_eight && b_is_eight) {
+        MVMGrapheme8  *a_blob8 = a_is_in_situ ? a->body.storage.in_situ_8 : a->body.storage.blob_8;
+        MVMGrapheme8  *b_blob8 = b_is_in_situ ? b->body.storage.in_situ_8 : b->body.storage.blob_8;
         while (i < scanlen && a_blob8[i] == b_blob8[i]) {
             i++;
         }
     }
-    else if (a->body.storage_type == MVM_STRING_IN_SITU_8 || b->body.storage_type == MVM_STRING_IN_SITU_8) {
-        MVMGrapheme8  *a_blob8 = a->body.storage.in_situ_8;
-        MVMGrapheme8  *b_blob8 = b->body.storage.in_situ_8;
-        while (i < scanlen && a_blob8[i] == b_blob8[i]) {
-            i++;
-        }
-    }
-    else if (a->body.storage_type == MVM_STRING_GRAPHEME_32 && b->body.storage_type == MVM_STRING_GRAPHEME_32) {
-        MVMGrapheme32  *a_blob32 = a->body.storage.blob_32;
-        MVMGrapheme32  *b_blob32 = b->body.storage.blob_32;
-        while (i < scanlen && a_blob32[i] == b_blob32[i]) {
-            i++;
-        }
-    }
-    else if (a->body.storage_type == MVM_STRING_IN_SITU_32 || b->body.storage_type == MVM_STRING_IN_SITU_32) {
-        MVMGrapheme32  *a_blob32 = a->body.storage.in_situ_32;
-        MVMGrapheme32  *b_blob32 = b->body.storage.in_situ_32;
+    else if (!a_is_eight && !b_is_eight) {
+        MVMGrapheme32  *a_blob32 = a_is_in_situ ? a->body.storage.in_situ_32 : a->body.storage.blob_32;
+        MVMGrapheme32  *b_blob32 = b_is_in_situ ? b->body.storage.in_situ_32 : b->body.storage.blob_32;
         while (i < scanlen && a_blob32[i] == b_blob32[i]) {
             i++;
         }
@@ -2509,9 +2502,15 @@ MVMint64 MVM_string_compare(MVMThreadContext *tc, MVMString *a, MVMString *b) {
             case MVM_STRING_GRAPHEME_ASCII:
                 blob8 = a->body.storage.blob_8;
                 break;
+            case MVM_STRING_IN_SITU_8:
+                blob8 = a->body.storage.in_situ_8;
+                break;
             case MVM_STRING_GRAPHEME_32:
                 blob32 = a->body.storage.blob_32;
                 break;
+            case MVM_STRING_IN_SITU_32:
+                blob32 = a->body.storage.in_situ_32;
+                break;
             default:
                 MVM_exception_throw_adhoc(tc,
                     "String corruption in string compare. Unknown string type.");
@@ -2521,9 +2520,15 @@ MVMint64 MVM_string_compare(MVMThreadContext *tc, MVMString *a, MVMString *b) {
             case MVM_STRING_GRAPHEME_ASCII:
                 blob8 = b->body.storage.blob_8;
                 break;
+            case MVM_STRING_IN_SITU_8:
+                blob8 = b->body.storage.in_situ_8;
+                break;
             case MVM_STRING_GRAPHEME_32:
                 blob32 = b->body.storage.blob_32;
                 break;
+            case MVM_STRING_IN_SITU_32:
+                blob32 = b->body.storage.in_situ_32;
+                break;
             default:
                 MVM_exception_throw_adhoc(tc,
                     "String corruption in string compare. Unknown string type.");
-- 
2.45.1
```

this fixes reproducible builds

timo (approved):
i think it's ready to merge

PR #1803, opened 2024-05-13
Provide some mechanisms for introspecting file mode

ab5tract:
In order to avoid adding yet-another member variable
to IO::Handle, we instead provide some new syscalls
for inquiring about the mode of the underlying
BOOTIO object.

Due to ambiguity created by O_RONLY being 0, we
need to set a second flag variable that only cares about
the r/w/rw distinction. We can do this directly in the
existing flag determination by adding a new parameter
and storing it at the right location.

We then expose our own values for simplified open file mode
(ro = 1, wo = 2, rw = 3) through a single syscall and let the
HLL check against it with new constants representing
the above mapping.

_EDITED: The above reflects the current commit message._

### DIFF
--- a/src/disp/syscall.c
+++ b/src/disp/syscall.c
@@ -1270,6 +1270,31 @@ static MVMDispSysCall async_unix_listen = {
     .expected_concrete = { 1, 1, 1, 1, 0 },
 };
 
+/* handle-open-mode */
+static void handle_open_mode_impl(MVMThreadContext *tc, MVMArgs arg_info) {
+    MVMOSHandle *handle  = (MVMOSHandle *)get_obj_arg(arg_info, 0);
+
+    if (handle->body.ops->introspection && handle->body.ops->introspection->mvm_open_mode) {
+        MVMint64 open_mode = handle->body.ops->introspection->mvm_open_mode(tc, handle);
+
+        if (open_mode < 1)
+            MVM_exception_throw_adhoc(tc, "Incomprehensible open mode: %lld (valid modes are 1=RO|2=WO|3=RW)", open_mode);
+
+        MVM_args_set_result_int(tc, (MVMint64) open_mode, MVM_RETURN_CURRENT_FRAME);
+    } else {
+        MVM_exception_throw_adhoc(tc, "Incompatible handle type supplied");
+    }
+}
+static MVMDispSysCall handle_open_mode = {
+    .c_name = "handle-open-mode",
+    .implementation = handle_open_mode_impl,
+    .min_args = 1,
+    .max_args = 1,
+    .expected_kinds = { MVM_CALLSITE_ARG_OBJ },
+    .expected_reprs = { MVM_REPR_ID_MVMOSHandle },
+    .expected_concrete = { 1 },
+};
+
 /* file-stat */
 static void file_stat_impl(MVMThreadContext *tc, MVMArgs arg_info) {
     MVMString *filename = get_str_arg(arg_info, 0);
@@ -1622,6 +1647,7 @@ void MVM_disp_syscall_setup(MVMThreadContext *tc) {
     add_to_hash(tc, &set_compunit_resolver);
     add_to_hash(tc, &async_unix_connect);
     add_to_hash(tc, &async_unix_listen);
+    add_to_hash(tc, &handle_open_mode);
     add_to_hash(tc, &file_stat);
     add_to_hash(tc, &stat_flags);
     add_to_hash(tc, &stat_time);
--- a/src/io/syncfile.c
+++ b/src/io/syncfile.c
@@ -44,6 +44,9 @@ typedef struct {
     /* Did read already report EOF? */
     int eof_reported;
 
+    /* Did the open mode specify O_RDONLY, O_WRONLY, or O_RDWR? (ignores other mode settings) */
+    short open_mode;
+
     /* Output buffer, for buffered output. */
     char *output_buffer;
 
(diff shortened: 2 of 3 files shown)


src/disp/syscall.c
```diff
@@ -1270,6 +1270,28 @@ static MVMDispSysCall async_unix_listen = {
...
 
+/* handle-open-mode */
+static void handle_open_mode_impl(MVMThreadContext *tc, MVMArgs arg_info) {
+    MVMOSHandle *oshandle  = (MVMOSHandle *)get_obj_arg(arg_info, 0);
+
+    MVMIOFileData *data = (MVMIOFileData *)oshandle->body.data;
+    MVMint64 open_mode  = (MVMint64) recover_open_mode_simple(data->open_mode_simple);
+
+    if (open_mode < 0)
+        MVM_exception_throw_adhoc(tc, "Incomprehensible open mode");
```
MasterDuke17, line 1281: Could this error message include some more context? E.g., something like “, %d is < 0” at the end.

src/io/syncfile.c
```diff
@@ -433,11 +405,26 @@ static int resolve_open_mode(int *flag, const char *cp) {
     }
 }
 
+/*
+    Returns only the part of the open mode relevant to checking readonly/writeonly/readwrite checks
+    (Because O_RDONLY is 0, we can't simply use bitwise logic).
+ */
+static void resolve_open_mode_simple(int *flag, const char *cp) {
```
niner, line 412: Why the rw parameter and not just return the result?

src/io/syncfile.c
```diff
@@ -433,11 +405,37 @@ static int resolve_open_mode(int *flag, const char *cp) {
...
```
MasterDuke17, line 412: Does this need to write to its argument? Couldn’t it simply return an int?

src/disp/syscall.c
```diff
@@ -1270,6 +1270,67 @@ static MVMDispSysCall async_unix_listen = {
     .expected_concrete = { 1, 1, 1, 1, 0 },
 };
 
+/* handle-mode-write-only */
+static void handle_mode_write_only_impl(MVMThreadContext *tc, MVMArgs arg_info) {
+    MVMOSHandle *oshandle  = (MVMOSHandle *)get_obj_arg(arg_info, 0);
+
+    if (! sizeof(oshandle->body.data) == sizeof(MVMIOFileData))
```
niner: That's not a good way for checking the handle type. In fact I'm quite surprised that it works at all. You are after all comparing the size of a `void *` with that of the MVMIOFileData struct.

I think that struct's definition should stay in syncfile.c and all access to it should go through proper accessor functions. Then the check here can simply be whether such a function exists or not.

src/io/syncfile.h
```diff
@@ -1,2 +1,39 @@
...
+    int fd;
+
+    /* Is it seekable? */
+    short seekable;
+
+    /* Is it known to be writable? */
+    short known_writable;
+
+    /* Was the mode specified readonly? */
+    short mode_readonly;
```
niner: We're probably not going to have millions of files open at the same time and if we do, the size of this struct would be the least of our problems. But even so, previously using shorts for flags was no waste, as one int + 2 shorts = 64 bits which they would have taken anyway due to byte_position's alignment requirements. However now that we have more flags here, we use 64 more bits just to store 3 flags. I think it's time to move them into a single 32 bit flags field.

niner (changes requested):

src/disp/syscall.c
ab5tract: Strange, I must not have pushed this fix, I know I dropped this check locally.

src/io/syncfile.c
ab5tract: I've since removed this extra function and now set the simplified flag value directly in `resolve_open_mode`. It's still an rw parameter, but at least it has a reason to be one now :)

src/disp/syscall.c
```diff
@@ -1270,6 +1270,28 @@ static MVMDispSysCall async_unix_listen = {
     .expected_concrete = { 1, 1, 1, 1, 0 },
 };
 
+/* handle-open-mode */
+static void handle_open_mode_impl(MVMThreadContext *tc, MVMArgs arg_info) {
+    MVMOSHandle *oshandle  = (MVMOSHandle *)get_obj_arg(arg_info, 0);
+
+    MVMIOFileData *data = (MVMIOFileData *)oshandle->body.data;
```
niner, line 1277: This is an unsafe pointer cast. We don't know that the handle is actually a file handle. At worst this can lead to a segfault.

I still say that rather than accessing the handle's data directly, the functionality should be hidden behind a proper IO op function pointer.

niner (changes requested):

src/disp/syscall.c
lizmat: @niner are you worried about a fuzzing injection vector?

ab5tract: Ah, that makes total sense. Applied.

src/io/syncfile.c
```diff
@@ -37,6 +37,9 @@ typedef struct {
     /* Is it known to be writable? */
     short known_writable;
 
+    /* Did the open mode specify O_RDONLY, O_WRONLY, or O_RDWR? (ignores other mode settings) */
+    short open_mode;
```
niner, line 41: If you just move open_mode to right after eof_reported, the whole struct shrinks by 8 bytes.

niner (approved):
Just the one change and this can go in.

src/io/syncfile.c
ab5tract: Done!

niner:
Good to go. Except, what's with CI on this PR?

ab5tract:
> Good to go. Except, what's with CI on this PR?

Honestly not sure. From the CI results of other PRs I was operating under the assumption that the pipeline is currently broken by default.

I guess we could test this theory by creating a PR that is a no-op mirror of main...

ab5tract:
So though #1807 demonstrates that there are indeed "default" failures in the CI, there are more failures happening in this PR than there should be.

The (non-defult) issue seems to stem from `t/09-moar/00-misc.t` failing a test:  

```
Subtest: no SPESH crashes with duplicate `use Test`
```

But to make things stranger, this test: 

1. doesn't fail locally on my Mac (though Mac_MVM pipeline fails
2. also fails for nqp main, but _not_ for Mac_MVM, only [Lin_MVM_reloc](https://dev.azure.com/jjmerelo0004/jjmerelo/_build/results?buildId=865&view=logs&j=07e75ff5-ce69-54f0-467d-376df9cd87cd&t=733f5fcc-37bb-5400-3254-3d19f26920c4&l=4500)

ab5tract:
@niner It looks like we are back to "only" failing the Windows tests thanks to [99b194bbd](https://github.com/Raku/roast/commit/99b194bbda6c0047d34cccdff883af348b1ec016) in `roast`.

Issue #1804, opened 2024-05-17
Memory Leaks in Termux


nitardus:
After installing Moar, npq and rakudo as shown [here](https://borg.moe/building-rakudo-perl-6-on-termux.html), termux crashed when I tried to install zef. It finally succeded installing zef with --/test, but it now crashes when I try to install anything: rakudo-gdb-m reveals that while zef searches for packages, there is MASSIVE forking / threading going on in the background, until it runs out of memory at  ~ 3 GiB memory usage. I recompiled moar with all sort of compiler options, and after using -asan, compilation of rakudo failed as follows:

`==13894==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 5120 byte(s) in 20 object(s) allocated from:
    #0 0xc56aecf21f14 in calloc out/lib/compiler-rt-aarch64/out/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:77:3
    #1 0xb91c4f838918 in mp_init /data/data/com.termux/files/home/src/MoarVM/3rdparty/libtommath/bn_mp_init.c:10:25
    #2 0xb91c4f7d7cd8 in MVM_bigint_radix /data/data/com.termux/files/home/src/MoarVM/src/math/bigintops.c:1630:16
    #3 0xb91c4f5b4a84 in MVM_interp_run /data/data/com.termux/files/home/src/MoarVM/src/core/interp.c:3336:40
    #4 0xb91c4f5406c4 in main /data/data/com.termux/files/home/src/MoarVM/src/main.c:307:10
    #5 0xc56ae905f068 in __libc_init (/apex/com.android.runtime/lib64/bionic/libc.so+0x5c068) (BuildId: 19a02d177e7ca67cb275d7ae9857394c)
    #6 0xb91c4f540024 in _start_main crtbegin.c

Direct leak of 4864 byte(s) in 19 object(s) allocated from:
    #0 0xc56aecf21f14 in calloc out/lib/compiler-rt-aarch64/out/llvm-project/compiler-rt/lib/asan/asan_malloc_linux.cpp:77:3
    #1 0xb91c4f838918 in mp_init /data/data/com.termux/files/home/src/MoarVM/3rdparty/libtommath/bn_mp_init.c:10:25
    #2 0xb91c4f7d17b4 in MVM_bigint_pow /data/data/com.termux/files/home/src/MoarVM/src/math/bigintops.c:650:24
[...]

PR #1805, opened 2024-05-18
Cleanup variables in bigint ops error cases

MasterDuke17:
In a bunch of error cases of bigint ops we weren't always cleaning up all the variables. These were noticed when investigating https://github.com/MoarVM/MoarVM/issues/1804, but I don't think they are the actual cause.


MasterDuke17:
It looks like t/02-rakudo/dd.t is failing in the all the CI runs, @lizmat @ab5tract related to some of your recent commits/discussion?

niner (approved):

PR #1808, opened 2024-05-19
Add missing __builtin_clz definition for MSVC

MasterDuke17:
It was missed when copying #defines from
https://github.com/llvm/llvm-project/blob/main/compiler-rt/lib/builtins/int_lib.h.

### DIFF
--- a/src/core/coerce.c
+++ b/src/core/coerce.c
@@ -17,14 +17,27 @@
 // This file is not part of the interface of this library.
 //
 //===----------------------------------------------------------------------===//
+#if !defined(__clang__)
+#include <intrin.h>
+
+static int __inline __builtin_clz(uint32_t value) {
+  unsigned long leading_zero = 0;
+  if (_BitScanReverse(&leading_zero, value))
+    return 31 - leading_zero;
+  return 32;
+}
+
 #if defined(_M_ARM) || defined(_M_X64)
+
 static int __inline __builtin_clzll(uint64_t value) {
   unsigned long leading_zero = 0;
   if (_BitScanReverse64(&leading_zero, value))
     return 63 - leading_zero;
-  return 63;
+  return 64;
 }
+
 #else
+
 static int __inline __builtin_clzll(uint64_t value) {
   uint32_t msh = (uint32_t)(value >> 32);
   uint32_t lsh = (uint32_t)(value & 0xFFFFFFFF);
@@ -32,6 +45,8 @@ static int __inline __builtin_clzll(uint64_t value) {
     return __builtin_clz(msh);
   return 32 + __builtin_clz(lsh);
 }
+
+#endif
 #endif
 #endif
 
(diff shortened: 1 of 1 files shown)


MasterDuke17:
Huh, anybody have an idea what's happening here? https://dev.azure.com/MoarVM/MoarVM/_build/results?buildId=1423&view=logs&j=97945c8c-7ab8-5ec2-f2c7-7c8541d3bf5c&t=439f2dd7-104a-517a-a6b1-3f392bb69fda&l=2266

timo:
do you know that the name exists in the dll for sure? maybe something changed somewhere. you should be able to use `dumpbin.exe /ALL D:\a\1\rakudo\13-cpp-mangling.dll` and `dumpbin.exe /ALL D:\a\1\rakudo\11-cpp.dll` and somewhere in there we should see the symbols we're not getting successfully

MasterDuke17:
> do you know that the name exists in the dll for sure? maybe something changed somewhere. you should be able to use `dumpbin.exe /ALL D:\a\1\rakudo\13-cpp-mangling.dll` and `dumpbin.exe /ALL D:\a\1\rakudo\11-cpp.dll` and somewhere in there we should see the symbols we're not getting successfully

I think the nativecall errors are unrelated to this change. But if you can explain https://dev.azure.com/MoarVM/MoarVM/_build/results?buildId=1436&view=logs&j=d36c29d8-9236-5cd5-d848-768a0b8a3eb3&t=d8ee4cfd-785a-54f9-bf7e-80664ee220b4&l=34 I would most grateful. If I pull our i64toa_jeaiii out into a separate file and run it with 1449334664 it's just fine. Is this a windows-only problem?

MasterDuke17:
> I think the nativecall errors are unrelated to this change. But if you can explain [dev.azure.com/MoarVM/MoarVM/_build/results?buildId=1436&view=logs&j=d36c29d8-9236-5cd5-d848-768a0b8a3eb3&t=d8ee4cfd-785a-54f9-bf7e-80664ee220b4&l=34](https://dev.azure.com/MoarVM/MoarVM/_build/results?buildId=1436&view=logs&j=d36c29d8-9236-5cd5-d848-768a0b8a3eb3&t=d8ee4cfd-785a-54f9-bf7e-80664ee220b4&l=34) I would most grateful. If I pull our i64toa_jeaiii out into a separate file and run it with 1449334664 it's just fine. Is this a windows-only problem?

And if I build this branch locally (with all the extra checks and debug fprintfs) it builds just fine without printing anything.

timo:
i don't have the brain power right now to actually look deeply at anything, but this looks kind of suspicious?

image

but probably just something you put in for debugging purposes?

MasterDuke17:
> i don't have the brain power right now to actually look deeply at anything, but this looks kind of suspicious?
> 
> image
> 
> but probably just something you put in for debugging purposes?

Well, that definitely needs to be clarified, but it *shouldn't* be the cause. That's just to figure out how big of a string to allocate, but even writing into a definitely-big-enough stack-allocated string initialized to all 0 ends up wrong.

timo:
not sure if this does anything, but try `%lld` instead of `%ld` in the printf, maybe that's causing a problem?

timo:
yeah at least some of the numbers in the debug output are suspiciously 31 bits long and then a much longer number for the output of m and m2

MasterDuke17:
> Maybe this could be related. After finding some test failures in `t/nqp/115-nums.t` on 32-bit Alpine Linux, i too placed an `fprintf` in `MVM_coerce_i_s` (but i put it before the call to `MVM_malloc`).
> 
> What i found was, the tests sometime cause `i` to take on a value of `-9223372036854775808` (`INT64_MIN`), which a few steps later causes `mag[msb]` to access past the end of `mag`.
> 
> I mentioned my observations with a bit more detail [here](https://github.com/MoarVM/MoarVM/pull/1799#discussion_r1623186473).

This is probably helpful information, thanks.

I wonder if we can easily add 32-bit Alpine Linux to our CI pipeline?

MasterDuke17:
@Cellesti the Windows CI jobs are passing now. Are you able to test with this branch? If so, and you tell me it's good, I'll cleanup all the commits and merge.

MasterDuke17:
Nope, I think the rest was for Windows. Thanks, seems like this is probably good to go.

PR #1809, opened 2024-05-19
Cheat to speedup using nqp::iterator+nqp::shift on a list

MasterDuke17:
We can assume the backing array is in a good state so reach in directly during iteration.

`time MVM_SPESH_BLOCKING=1 nqp -e 'my $l := nqp::list(); my int $p := 0; while $p++ < 10_000 { nqp::push($l, $p) }; my int $j := 10000; while $j-- { my $iter := nqp::iterator($l); nqp::while($iter, my $a := nqp::shift($iter)) }'` drops from ~1.46s to ~1.25s and `time MVM_SPESH_BLOCKING=1 nqp -e 'my $l := nqp::list_s(); my int $p := 0; while $p++ < 10_000 { nqp::push_s($l, $p) }; my int $j := 10000; while $j-- { my $iter := nqp::iterator($l); nqp::while($iter, my $a := nqp::shift_s($iter)) }'` drops from ~2.01s to ~1.83s.

### DIFF
--- a/src/6model/reprs/MVMIter.c
+++ b/src/6model/reprs/MVMIter.c
@@ -49,12 +49,15 @@ static const MVMStorageSpec * get_storage_spec(MVMThreadContext *tc, MVMSTable *
 static void shift(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *data, MVMRegister *value, MVMuint16 kind) {
     MVMIterBody *body = (MVMIterBody *)data;
     MVMObject *target = body->target;
+    MVMArrayBody *abody;
     switch (body->mode) {
         case MVM_ITER_MODE_ARRAY:
             body->array_state.index++;
             if (body->array_state.index >= body->array_state.limit)
                 MVM_exception_throw_adhoc(tc, "Iteration past end of iterator");
-            REPR(target)->pos_funcs.at_pos(tc, STABLE(target), target, OBJECT_BODY(target), body->array_state.index, value, kind);
+            abody = (MVMArrayBody *)OBJECT_BODY(target);
+            MVMObject *found = abody->slots.o[abody->start + body->array_state.index];
+            value->o = found ? found : tc->instance->VMNull;
             return;
         case MVM_ITER_MODE_ARRAY_INT:
             body->array_state.index++;
@@ -92,13 +95,13 @@ static void shift(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *da
             body->array_state.index++;
             if (body->array_state.index >= body->array_state.limit)
                 MVM_exception_throw_adhoc(tc, "Iteration past end of iterator");
+            abody = (MVMArrayBody *)OBJECT_BODY(target);
+            MVMString *str = abody->slots.s[abody->start + body->array_state.index];
             if (kind == MVM_reg_str) {
-                REPR(target)->pos_funcs.at_pos(tc, STABLE(target), target, OBJECT_BODY(target), body->array_state.index, value, kind);
+                value->s = str;
             }
             else if (kind == MVM_reg_obj) {
-                MVMRegister tmp;
-                REPR(target)->pos_funcs.at_pos(tc, STABLE(target), target, OBJECT_BODY(target), body->array_state.index, &tmp, MVM_reg_str);
-                value->o = MVM_repr_box_str(tc, MVM_hll_current(tc)->str_box_type, tmp.s);
+                value->o = MVM_repr_box_str(tc, MVM_hll_current(tc)->str_box_type, str);
             }
             else {
                 MVM_exception_throw_adhoc(tc, "Wrong register kind in iteration");


MasterDuke17:
I noticed this when looking at https://github.com/Raku/nqp/issues/685.

niner:
Why can we assume that the array is in a good state?

MasterDuke17:
> Why can we assume that the array is in a good state?

Well, I believe we’re allowed to *act* as if it’s in a good state. Modifying an array during iteration is not allowed AIUI. Also, the iterator keeps track of the index compared to the size of the array.

raiph:
> Modifying an array during iteration is not allowed AIUI.

I couldn't tell if this comment was purely about _nqp_ level commitments or also about _Raku_ level commitments.

For Raku, I'll say this. Shortly after `6.d` shipped, jnthn commented under [an SO answer by Vadim Belman](https://stackoverflow.com/questions/53856578/general-approach-to-modifying-an-array-while-looping-through-it). Based on my existing belief that modification during iteration was fine, plus Vadim's answer, plus jnthn's comment under it, I'd say it looks like it is "officially" allowed in Raku `6.d`.

For nqp, I'll say this. The [doc for `nqp::iterator`](https://github.com/Raku/nqp/blob/4604c6664ea4ba89285137b19dd6533f2ff1f29a/docs/ops.markdown#iterator) says (with my **added EMPHASIS**):

>You can also use `nqp::iterator()` to iterate over a list, which is functionally equivalent to cloning the list and not having to know which form of `shift` to have to use.

```
my $list := nqp::list('a', 'b', 'c');
my $iter := nqp::iterator($list);

while $iter {
    say(nqp::shift($iter));
}
```

> Using `iterator` this way proved to be not very well optimizable, so **this form will probably be DEPRECATED in the future**. Depending on circumstances one could use `clone` instead, or use an index variable...

timo:
This is the commit that seems to have introduced the verbiage

> functionally equivalent to cloning the list and not having to know which form of shift to have to use

https://github.com/Raku/nqp/commit/db6e30309c5c27c503903e972999752715502459

could have been by accident, or accidentally done two unrelated commits in one go?

In any case, I think this does not hold true whenever a list is shared between threads, and doesn't hold true whenever a list is modified while taking values from the iterator? I don't think NQP ever promised this kind of behaviour, and it's kind of tricky to do without literally cloning the list you're iterating over— remember that nqp doesn't have containers inside of lists.

I haven't looked at the diff of this PR yet

timo:
looked at the code now. directly going to the slots array inside the VMArray is not safe when just checking `(body->array_state.index >= body->array_state.limit)` before accessing it, because the `array_state.limit` is initialised and set when the iterator is created from the array. That means:

 * if the array is made smaller from the end, we can now access values in the slots array that are no longer valid. the GC doesn't update any entries beyond the end. we currently don't actually shrink the allocation I don't think, but that may change one day and that would cause the same problem
 * if the array is made smaller by actually shifting off entries from the start, the `start` index is incremented, which is another way to access outside the valid range of the array, with the same effect

What we could still do, however, is to use the fact that we know what `REPR(target)` is (the array iteration mode are only set if the repr of the target has `MVM_REPR_ID_VMArray`). There's shortcut functions in VMArray.c that let you refer to VMArray's different at_pos functions directly by its name, which is something the JIT benefits from: `MVM_VMArray_at_pos_o` and friends.

You can try if the performance gains are still good with only that change.

If you want to try that, and you have a patch ready, someone should also test it with the JIT on x86_64, I think.

Also, maybe looking at the spesh log or a C-level profile can be interesting here.

MasterDuke17:
> You can try if the performance gains are still good with only that change.
> 
> If you want to try that, and you have a patch ready, someone should also test it with the JIT on x86_64, I think.

I just tested this change to my branch (after rebasing to main) on my x86_64 system:
```patch
diff --git src/6model/reprs/MVMIter.c src/6model/reprs/MVMIter.c
index a94abf59a..9c7a7214a 100644
--- src/6model/reprs/MVMIter.c
+++ src/6model/reprs/MVMIter.c
@@ -55,9 +55,7 @@ static void shift(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *da
             body->array_state.index++;
             if (body->array_state.index >= body->array_state.limit)
                 MVM_exception_throw_adhoc(tc, "Iteration past end of iterator");
-            abody = (MVMArrayBody *)OBJECT_BODY(target);
-            MVMObject *found = abody->slots.o[abody->start + body->array_state.index];
-            value->o = found ? found : tc->instance->VMNull;
+            MVM_VMArray_at_pos_o(tc, STABLE(target), target, OBJECT_BODY(target), body->array_state.index, value);
             return;
         case MVM_ITER_MODE_ARRAY_INT:
             body->array_state.index++;
```
and it increases the time back up to what it is on main (i.e., ~1.57 main, ~1.31 branch, ~1.59 branch+patch). I tested with:
`MVM_SPESH_BLOCKING=1 time -p nqp -e 'my $l := nqp::list(); my int $p := 0; while $p++ < 10_000 { nqp::push($l, $p) }; my int $j := 10000; while $j-- { my $iter := nqp::iterator($l); nqp::while($iter, my $a := nqp::shift($iter)) }'`

PR #1820, opened 2024-07-08
Update OS versions in Azure CI config

MasterDuke17:
Use newest non-preview versions as default and name the older available versions.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -40,21 +40,26 @@ stages:
            MOAR_OPTIONS: '--relocatable'
 
          Mac_MVM:
-           IMAGE_NAME: 'macOS-12'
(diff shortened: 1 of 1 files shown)


patrickbkr (approved):
Looking good. Can we have the same change in NQP and rakudo as well to keep them as much in sync as possible?

MasterDuke17:
> Looking good. Can we have the same change in NQP and rakudo as well to keep them as much in sync as possible?

Any idea what's going on with the fails in the new windows jobs?

patrickbkr:
A quick googling tentatively points to a batch error. I don't see anything obviously wrong with the failing command though.

MasterDuke17:
Yeah, I can't find anything definitive. For now I'm just going to back out the windows changes.

Issue #1817, opened 2024-06-29
JIT issue with new version of .flat


lizmat:
Steps to reproduce (on Intel boxes):

1. checkout rakudo 2024.06
2. revert 739559906a6409f1e88301fdf92c0e227f8d0752
3. build rakudo  (it *may* already crash when installing modules)
4. run `raku t/spec/S17-promise/start.t`

For me, it reliably fails with:
```
...
ok 59 - No data races on closure interpolation in strings
An operation first awaited:
  in block <unit> at t/spec/S17-promise/start.t line 261

Died with the exception:
    Can only specify Awaitable objects to await (got a List)
      in block <unit> at t/spec/S17-promise/start.t line 261
```

5. run `MVM_JIT_DISABLE=1 raku t/spec/S17-promise/start.t`

Tests complete ok.

lizmat:
I've been able to golf this to:
```raku
my %h = :1a,:2b,:3c,:4d;
while (1,2,3,4,"a","b","c","d") eq (%h{*}:kv).sort {
    say ++$
}
```
This generally stops after 17 iterations for me.  Runs forever with `MVM_SPESH_DISABLE=1`.

Mind you, it is **STRANGE** that on Apple Silicon (M1) this runs without any problems, while I *do* think that spesh is enabled on non-Intel hardware, is it not?

Or is `MVM_SPESH_DISABLE=1` also disabling JIT?  I forgot  :-(

MasterDuke17:
Yeah, I think disabling spesh also disables the jit.Sent from my iPhoneOn Jun 29, 2024, at 2:11 PM, Elizabeth Mattijsen ***@***.***> wrote:﻿
I've been able to golf this to:
my %h = :1a,:2b,:3c,:4d;
while (1,2,3,4,"a","b","c","d") eq (%h{*}:kv).sort {
    say ++$
}
This generally stops after 17 iterations for me.  Runs forever with MVM_SPESH_DISABLE=1.
Mind you, it is STRANGE that on Apple Silicon (M1) this runs without any problems, while I do think that spesh is enabled on non-Intel hardware, is it not?
Or is MVM_SPESH_DISABLE=1 also disabling JIT?  I forgot  :-(

lizmat:
Looking at the code I wrote and trying a few changes, I wonder whether it could be the use of `nqp::decont` in a numerical, rather than a boolean way.

patrickbkr:
Honest question:
[...]

PR #1822, opened 2024-07-15
Disable the expression jit by default

MasterDuke17:
Per https://github.com/MoarVM/MoarVM/issues/1817#issuecomment-2217676548


niner (approved):

patrickbkr (approved):
I've pinged lizmat for feedback from her IRC log server tests. Once that feedback is in - Let's do this!

PR #1823, opened 2024-07-24
Silence a bunch of valgrind errors

MasterDuke17:
Running t/spec/S12-construction/destruction.t under valgrind results in a lot of `Conditional jump or move depends on uninitialised value(s)` and this silences a bunch of them.

It still reports a couple in `consume_reprop (graph.c:665)` and `add_resume_init_temp_to_fake (program.c:2671)`, but I haven't been able to figure out why or how to silence them.

### DIFF
--- a/src/jit/graph.c
+++ b/src/jit/graph.c
@@ -664,7 +664,7 @@ static void jg_sc_wb(MVMThreadContext *tc, MVMJitGraph *jg, MVMSpeshOperand chec
 static MVMint32 consume_reprop(MVMThreadContext *tc, MVMJitGraph *jg,
                                MVMSpeshIterator *iter, MVMSpeshIns *ins) {
     MVMint16 op = ins->info->opcode;
-    MVMSpeshOperand type_operand;
+    MVMSpeshOperand type_operand = {0};
     MVMSpeshFacts *type_facts = 0;
     MVMint32 alternative = 0;
 
--- a/src/spesh/inline.c
+++ b/src/spesh/inline.c
@@ -596,7 +596,7 @@ static MVMSpeshBB * merge_graph(MVMThreadContext *tc, MVMSpeshGraph *inliner,
             MVMSpeshDeoptUseEntry *due = inliner->facts[i][j].usage.deopt_users;
             while (due) {
                 if ((MVMuint32)due->deopt_idx == impl_deopt_idx) {
-                    MVMSpeshOperand o;
+                    MVMSpeshOperand o = {0};
                     o.reg.orig = i;
                     o.reg.i = j;
                     MVM_VECTOR_PUSH(regs_for_deopt, o);
@@ -789,7 +789,7 @@ static MVMSpeshBB * merge_graph(MVMThreadContext *tc, MVMSpeshGraph *inliner,
     orig_deopt_pea_mat_infos = MVM_VECTOR_ELEMS(inliner->deopt_pea.materialize_info);
     for (i = 0; i < MVM_VECTOR_ELEMS(inlinee->deopt_pea.materialize_info); i++) {
         MVMSpeshPEAMaterializeInfo mi_orig = inlinee->deopt_pea.materialize_info[i];
-        MVMSpeshPEAMaterializeInfo mi_new;
+        MVMSpeshPEAMaterializeInfo mi_new = {0};
         mi_new.stable_sslot = mi_orig.stable_sslot + inliner->num_spesh_slots;
         mi_new.num_attr_regs = mi_orig.num_attr_regs;
         if (mi_new.num_attr_regs) {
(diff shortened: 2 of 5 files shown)


lizmat:
OOC, would this have any runtime effects?

niner:
How deeply did you investigate every case (I haven't)? I'm a bit worried because not in every case is 0 a clear signal for missing value like it is for pointers. In some cases the valgrind warning may point at an actual problem.

src/jit/graph.c
```diff
@@ -664,7 +664,7 @@ static void jg_sc_wb(MVMThreadContext *tc, MVMJitGraph *jg, MVMSpeshOperand chec
 static MVMint32 consume_reprop(MVMThreadContext *tc, MVMJitGraph *jg,
                                MVMSpeshIterator *iter, MVMSpeshIns *ins) {
     MVMint16 op = ins->info->opcode;
-    MVMSpeshOperand type_operand;
+    MVMSpeshOperand type_operand = {0};
```
dontlaugh, line 667: This one is unclear to me

src/spesh/inline.c
```diff
@@ -596,7 +596,7 @@ static MVMSpeshBB * merge_graph(MVMThreadContext *tc, MVMSpeshGraph *inliner,
             MVMSpeshDeoptUseEntry *due = inliner->facts[i][j].usage.deopt_users;
             while (due) {
                 if ((MVMuint32)due->deopt_idx == impl_deopt_idx) {
-                    MVMSpeshOperand o;
+                    MVMSpeshOperand o = {0};
                     o.reg.orig = i;
                     o.reg.i = j;
```
dontlaugh, line 601: Seems ok; both fields of the `reg` variant are set.

The `MVMSpeshOperand` type could use some doc strings, fwiw.

src/spesh/inline.c
```diff
@@ -789,7 +789,7 @@ static MVMSpeshBB * merge_graph(MVMThreadContext *tc, MVMSpeshGraph *inliner,
     orig_deopt_pea_mat_infos = MVM_VECTOR_ELEMS(inliner->deopt_pea.materialize_info);
     for (i = 0; i < MVM_VECTOR_ELEMS(inlinee->deopt_pea.materialize_info); i++) {
         MVMSpeshPEAMaterializeInfo mi_orig = inlinee->deopt_pea.materialize_info[i];
-        MVMSpeshPEAMaterializeInfo mi_new;
+        MVMSpeshPEAMaterializeInfo mi_new = {0};
```
dontlaugh, line 792: All the fields of this struct are set :+1:

src/spesh/inline.c
```diff
@@ -804,7 +804,7 @@ static MVMSpeshBB * merge_graph(MVMThreadContext *tc, MVMSpeshGraph *inliner,
     }
     for (i = 0; i < MVM_VECTOR_ELEMS(inlinee->deopt_pea.deopt_point); i++) {
         MVMSpeshPEADeoptPoint dp_orig = inlinee->deopt_pea.deopt_point[i];
-        MVMSpeshPEADeoptPoint dp_new;
+        MVMSpeshPEADeoptPoint dp_new = {0};
```
dontlaugh, line 807: All the fields are set; seems okay :+1:

src/spesh/pea.c
```diff
@@ -141,7 +141,7 @@ static MVMuint16 get_deopt_materialization_info(MVMThreadContext *tc, MVMSpeshGr
         return alloc->deopt_materialization_idx;
     }
     else {
-        MVMSpeshPEAMaterializeInfo mi;
+        MVMSpeshPEAMaterializeInfo mi = {0};
```
dontlaugh, line 144: All the fields are set, but I'd move this declaration down to line 160, close to it's initialization.

We can leave a comment in its place to document the `else` block here if that helps.

src/spesh/pea.c
```diff
@@ -365,7 +365,7 @@ static MVMSpeshFacts * get_shadow_facts_c(MVMThreadContext *tc, GraphState *gs,
 static MVMSpeshFacts * create_shadow_facts_h(MVMThreadContext *tc, GraphState *gs, MVMuint16 idx) {
     MVMSpeshFacts *facts = get_shadow_facts_h(tc, gs, idx);
     if (!facts) {
-        ShadowFact sf;
+        ShadowFact sf = {0};
```
dontlaugh, line 368: Careful here. `ShadowFact` embeds a large struct `MVMSpeshFacts` and I don't know how that one works :shrug:

dontlaugh (commented):
I am a fan of trying to silence warnings, and also of explicit initialization, so I took some time to look at each of these changes and give my take.

dontlaugh:
> OOC, would this have any runtime effects?

Not a compiler expert, but it seems unlikely unless the struct is large, with many fields. And even in _that_ case, it would only matter for hot loops. `MVMDispProgramOp` is a kind of tagged-union type thing, so it's got a lot of fields but doesn't take a lot of memory on the stack.

> I'm a bit worried because not in every case is 0 a clear signal for missing value like it is for pointers. In some cases the valgrind warning may point at an actual problem.

This is a valid concern for anything in a struct that isn't a pointer. Valgrind is being helpful here because relying on uninitialized _anything_ needs scrutiny.

PR #1825, opened 2024-07-31
Iterate graphemes into in_situ_32 string if able

MasterDuke17:
Instead of checking after the iteration and converting it into an in_situ_32 string (alloc + write + copy + free) if able, turn it into an in_situ_32 beforehand (write).

Inspired by https://github.com/MoarVM/MoarVM/pull/1802#discussion_r1644267273.

A micro-benchmark of `MVM_SPESH_BLOCKING=1 nqp-m -e 'my str $s := "…b…d…f…h…j…l…n…p…r…t…v…x…z"; my int $i := 0; my str $a; my $n := nqp::time; while $i++ < 1_000_000 { my int $j := 0; while $j++ < 13 { $a := nqp::substr($s, $j * 2, 2); }; }; say(nqp::div_n(nqp::time - $n, 1000000000e0)); say($a)'` decreases in runtime from ~0.49s to ~0.36s, in instructions from ~6.34b to ~5.32b, and in allocations as reported by heaptrack from ~12m to ~236k.

### DIFF
--- a/src/strings/ops.c
+++ b/src/strings/ops.c
@@ -328,9 +328,15 @@ static void iterate_gi_into_string(MVMThreadContext *tc, MVMGraphemeIter *gi, MV
     }
     else {
         MVMStringIndex result_pos = 0;
-        result->body.storage_type            = MVM_STRING_GRAPHEME_32;
-        result32 = result->body.storage.blob_32 =
-            result_graphs ? MVM_malloc(result_graphs * sizeof(MVMGrapheme32)) : NULL;
+        if (result_graphs <= 2) {
+            result->body.storage_type            = MVM_STRING_IN_SITU_32;
+            result32 = result->body.storage.in_situ_32;
+        }
+        else {
+            result->body.storage_type            = MVM_STRING_GRAPHEME_32;
+            result32 = result->body.storage.blob_32 =
+                MVM_malloc(result_graphs * sizeof(MVMGrapheme32));
+        }
         while (1) {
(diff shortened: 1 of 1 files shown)


src/strings/ops.c
```diff
@@ -328,9 +328,15 @@ static void iterate_gi_into_string(MVMThreadContext *tc, MVMGraphemeIter *gi, MV
...
-        result32 = result->body.storage.blob_32 =
-            result_graphs ? MVM_malloc(result_graphs * sizeof(MVMGrapheme32)) : NULL;
+        if (result_graphs <= 2) {
+            result->body.storage_type            = MVM_STRING_IN_SITU_32;
+            result32 = result->body.storage.in_situ_32;
+        }
+        else {
+            result->body.storage_type            = MVM_STRING_GRAPHEME_32;
+            result32 = result->body.storage.blob_32 =
+                result_graphs ? MVM_malloc(result_graphs * sizeof(MVMGrapheme32)) : NULL;
```
timo, line 338: i think the ternary is not necessary since result_graphs is already checked to be > 2 in this branch, and [further up](https://github.com/MoarVM/MoarVM/pull/1825/files#diff-bc76be5614df2473912916db269fc246c79efbc83a0a065ee8b785aa3c68d040R258) we already bail out early if result_graphs is zero

timo (changes requested):
looks good to have

src/strings/ops.c
MasterDuke17: Yeah, good catch, I'll make the change.

PR #1826, opened 2024-08-02
When guarding conc or type and type is known, guard just conc/type

timo:
The difference in moar bytecode is very small, but in JITted code it's a bit bigger since the non-"just" variant also jits a spesh slot access first.


MasterDuke17:
I don't know spesh well enough to review, but CI is clean, any reason not to merge?

PR #1827, opened 2024-08-02
return pre-made empty string constant from latin1 and ascii

timo:


MasterDuke17:
Why not also for utf8?

PR #1832, opened 2024-08-27
Add MVM_string_decodestream_get_up_to_chars

timo:
Allows us to ask for an amount of characters, and get something out of the decodestream even when there's not as much to get as we ask for.

IO::Handle's readchars method by default tries to
get $*DEFAULT-READ-ELEMS from the decoder and won't return anything until that amount of chars has become available from repeated reads.

This change brings MVM_decoder_take_chars behaviour in line with the comment in front of it ("or all if there is not enough") and fixes both the readchars and Supply methods on $*IN with an encoding blocking until a lot of input has been read.

IO::Handle.lines.Supply already did what you wanted, and if you unset the encoding on $*IN first, .Supply.lines also works. Now both work.

resolves https://github.com/rakudo/rakudo/issues/5189

### DIFF
--- a/src/6model/reprs/Decoder.c
+++ b/src/6model/reprs/Decoder.c
@@ -252,7 +252,7 @@ MVMString * MVM_decoder_take_chars(MVMThreadContext *tc, MVMDecoder *decoder, MV
     MVMString *result = NULL;
     enter_single_user(tc, decoder);
     MVMROOT(tc, decoder, {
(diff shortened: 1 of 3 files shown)


timo:
i would like your input on this @jnthn since you made most of the streaming encoder stuff. t/nqp/116-streaming-decoder.t explicitly requires nqp::decodertakechars to return null unless the full requested amount can be returned at once. Decoder.consume-exactly-chars is implemented using this op, so the name kind of implies that we do want it to only return the exact amount (except when EOF).

PR #1839, opened 2024-09-16
Add example scripts using the Zulip API

dontlaugh:
The Raku community has a Zulip instance at raku.zulipchat.com with a 50GB file storage drive.

Individual files can be up to 25MB.

We could use Zulip as an automation target for sharing test results, deploys, releases, etc.

### DIFF
--- a/tools/zulip-file-upload.sh
+++ b/tools/zulip-file-upload.sh
@@ -1,0 +1,43 @@
+#!/bin/bash
+#
+# Docs https://zulip.com/api/upload-file
+#
+# Must set env vars:
+#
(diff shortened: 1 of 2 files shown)


dontlaugh:
Here's the zulip channel I posted to https://raku.zulipchat.com/#narrow/stream/393411-dev/topic/API.20Test

dontlaugh:
@timo I'm pasting these examples because I thought a zulip channel might be a useful place to post test results;

File uploads to zulip are kind of slow, and we only have 50GB. But it's there if it's needed.

Otherwise, posting a message to a pre-existing channel+topic is quite fast.

PR #1843, opened 2024-10-04
Give containers explicit uint store funcs

timo:
give the code_pair and value_descriptor an explicit function that takes an unsigned integer argument.

This silences an error with newer gcc versions that was first brought to our attention by the debian builders


MasterDuke17:
Huh, what's with all the segvs in the memory leak test?

lizmat:
So, ready to merge??

PR #1847, opened 2024-10-20
Add telemetry log output to more pieces of moar: queues, semaphores, async processes, compunit loading, plus syscall interface for user code

timo:
* Make sure that MVM_TELEMETRY_LOG is respected in nqp and rakudo runner binaries
* output some information when queues and semaphores do something
* output some information when an async process is spawned
* output some information when a compunit is loaded from a file, or from a bucket of bytes
* allow putting output in the telemetry log using `nqp::syscall` from nqp/raku internals

### DIFF
--- a/src/6model/reprs/ConcBlockingQueue.c
+++ b/src/6model/reprs/ConcBlockingQueue.c
@@ -126,7 +126,7 @@ static void at_pos(MVMThreadContext *tc, MVMSTable *st, MVMObject *root, void *d
         peeked = body->head->next;
         value->o = peeked ? peeked->value : tc->instance->VMNull;
         uv_mutex_unlock(&body->head_lock);
-        MVM_telemetry_interval_stop(tc, interval_id, "ConcBlockingQueue.at_pos");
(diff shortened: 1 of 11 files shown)


timo:
remaining CI failures seem to be also on the main branch.

timo:
```
env MVM_TELEMETRY_LOG=/tmp/custom_telemetry.txt rakudo -e 'use nqp; my int $end = nqp::syscall("telemetry-interval-start", 99); nqp::syscall("telemetry-interval-annotate", my int $ = $end, 1024, "hey there how are you?"); sleep 0.4; nqp::syscall("telemetry-interval-stop", my int64 $ = $end, 0x42);'
```
results in output like this:

```
51022020000       507151338 (-   "custom #99" (89)
        400                 ???  "hey there how are you?" (89)
51022020000      2391129183  -)  "custom #66" (89)
          0      2394325982  -)  "moarvm teardown" (1)
```

telemeh is built to heavily prefer constant strings from inside the C source code, because then no string copying or malloc/free is necessary. That's the reason why this user space API just allows numbers from 0 through 4095 for interval start and end, and using a "dynamic" string (even if it's a constant in the source code) is only possible with the interval-annotate syscall

PR #1848, opened 2024-10-20
Install gdb extension

timo:
gdb looks next to a .so file for a matching blabla-gdb.py (or others) for auto-loadable gdb extensions. We provide one with a small amount of features right now, plus more in a future version.


patrickbkr:
Can this be merged? Looki like a no-brainer.

PR #1851, opened 2024-10-24
Spesh inliner: max inline depth for "try_get_graph_from_unspecialized"

### LINKED ISSUE #1845: Spesh issue when `.gist`ing a `Seq`
This clearly is a MoarVM issue, not a Rakudo issue: https://github.com/rakudo/rakudo/issues/4921

timo:
fixes #1845

Quotes from the code:

```
    /* When creating graphs for inline candidates, track how deep we've
     * recursed, so we don't infinitely recurse.
     *
     * This limit should only cause try_get_graph_from_unspecialized to not
     * recurse deeply, since after an inline graph is created and merged into
     * the inlinee, we run optimize over it again and re-encounter ops that
     * may lead to inline attemps, like runbytecode.
```

```
 * When trying to make a spesh graph to inline from an unspecialized static
 * frame, we run MVM_spesh_optimize, which can lead to another attempt to
 * create a graph the same way. Don't do that!
```

### DIFF
--- a/src/spesh/inline.c
+++ b/src/spesh/inline.c
@@ -1,5 +1,12 @@
 #include "moar.h"
 
+/*
+ * When trying to make a spesh graph to inline from an unspecialized static
+ * frame, we run MVM_spesh_optimize, which can lead to another attempt to
+ * create a graph the same way. Don't do that!
+ */
+#define SPESH_UNSPECIALIZED_INLINE_RECURSION_DEPTH_LIMIT 3
+
 /* Ensures that a given compilation unit has access to the specified extop. */
 static void demand_extop(MVMThreadContext *tc, MVMCompUnit *target_cu,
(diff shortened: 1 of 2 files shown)


src/spesh/inline.c
```diff
@@ -1,5 +1,12 @@
 #include "moar.h"
 
+/*
+ * When trying to make a spesh graph to inline from an unspecialized static
+ * frame, we run MVM_spesh_optimize, which can lead to another attempt to
+ * create a graph the same way. Don't do that!
+ */
+#define SPESH_UNSPECIALIZED_INLINE_RECURSION_DEPTH_LIMIT 3
```
niner, line 8: Feels like there are still a lot of sensible numbers between 3 and infinity. In other words, this limit seems quite low. How did you come up with it?

PR #1859, opened 2024-11-13
Fix longjmp on Windows

patrickbkr:
See https://blog.lazym.io/2020/09/21/Unicorn-Devblog-setjmp-longjmp-on-Windows/ for a good explanation on what we're doing here.

### DIFF
--- a/src/platform/setjmp.h
+++ b/src/platform/setjmp.h
@@ -1,5 +1,5 @@
 /* MinGW x64 BUGFIX: Pass a NULL argument instead of the frame pointer to the
- * setjmp implementation to make the JIT not choke on exceptions on MinGW. 
+ * setjmp implementation to make the JIT not choke on exceptions on MinGW.
  * cygx++ for fix */
 #ifdef __MINGW32__
 #  ifndef USE_NO_MINGW_SETJMP_TWO_ARGS
@@ -23,5 +23,7 @@
 #    elif defined _M_ARM64
 #      define MVM_setjmp(BUF) do { setjmp((BUF)); tc->interp_jump[0] = 0; } while (0)
 #    endif
+#  else
+#    define MVM_setjmp setjmp
 #  endif
 #endif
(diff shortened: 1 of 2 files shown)


raiph:
Iirc the JIT has been disabled by default for main head with no plan to re-enable it by default. Is my memory about right and so this fix is "just" for those who decide to nevertheless manually re-enable it or am I misremembering or misunderstanding something?

patrickbkr:
Moar has long had two JITs. The LegoJIT and the ExpressionJIT. We disabled - and plan to remove - the ExprJIT. The LegoJIT stays.

src/platform/setjmp.h
```diff
@@ -1,6 +1,28 @@
-/* MinGW x64 BUGFIX: Pass a NULL argument instead of the frame pointer to the
- * setjmp implementation to make the JIT not choke on exceptions on MinGW.
- * cygx++ for fix */
+/* In later Windows versions the longjmp implementation tries to perform stack
+ * unwinding (for the purposes of cleanup). This is problematic for MoarVM
+ * because our JIT produces frames that this stack unwinder doesn't understand.
+ * Luckily the longjmp function can be instructed to not do the stack unwinding
+ * by setting a respective field in the jmpbuf structure. The setjmp function
+ * does so when it's *second* argument is NULL. (The frame pointer) Microsoft
```
MasterDuke17, line 6: `(The frame pointer)` seems kind of out of place?

src/platform/setjmp.h
```diff
@@ -1,6 +1,28 @@
...
+ * the function is marked as purely intrinsic and the compiler won't let you
+ * take the address of such a function. So we instead modify the resulting
+ * jmpbuf and set that flag ourself. Microsofts `setjmp.h` header file has
+ * proven helpful in disecting the otherwise opaque jmp_buf struct.
+ * All of the above (both, the two arg setjmp we use on MinGW and modifying
+ * the jmpbuf we do on MSVC) is not part of any public Windows API and could in
+ * principle break any time. Given that
+ * - Microsoft has a long history of being very careful to keep back-compat and
+ * - these hacks are in use by MinGW and by extension some high profile
+ *   projects (e.g. QEmu) and
```
MasterDuke17, line 20: I believe 'QEMU' is correct (i.e., all caps).

PR #1865, opened 2024-11-19
Debugserver loaded file notifications

timo:
> ### Loaded Files Request (50)
> 
> In order to reliably set breakpoints, the filename supplied to the breakpoint
> command needs to match what is in the annotation exactly.
> 
> This command allows a debug client to ask what file names have been
> seen so far.
> 
> With the "start_watching" key set to True, notifications when a new
> filename is seen will be sent with File Loaded Notification type and the
> id of this request.
> 
> With "suspend", the thread that first encounters the new filename will
> suspend itself.
> 
> With "stacktrace", the notifications will also immediately send a stacktrace
> along for the thread that encounters the new file.
> 
>     {
>         "type": 50,
>         "id": $id,
>         "start_watching": True,
>         "suspend": False,
>         "stacktrace": False,
>     }
> 
> ### File Loaded Notification (51)
> 
> Response to a Loaded Files Request, as well as notification when new files
> show up later on.
> 
> Initial response:
> 
>     {
>         "type": 51,
>         "id": $id,
>         "filenames": [
>             { "path": "src/vm/moar/ModuleLoader.nqp" },
>             { "path": "gen/moar/CORE.c.setting" },
>             { "path": "NQP::src/how/Archetypes.nqp" },
>             { "path": "SETTING::src/core.c/List.rakumod" },
>         ]
>     }
> 
> Notification:
> 
>     {
>         "type": 51,
>         "id": $id,
>         "thread": 1,
>         "filenames": [
>             { "path": "src/Perl6/Metamodel/PrivateMethodContainer.nqp" },
>         ],
>         "frames": [
>             ...
>         ]
>     }

### DIFF
--- a/docs/debug-server-protocol.md
+++ b/docs/debug-server-protocol.md
@@ -1083,6 +1083,14 @@ along for the thread that encounters the new file.
 Response to a Loaded Files Request, as well as notification when new files
 show up later on.
 
+Filename entries that were created not from a corresponding annotation being
+encountered but from requesting a breakpoint to be installed will have the
+"pending" key in addition to the "path" key.
+
+Creating a file by requesting a breakpoint does not cause a notification to
+be sent out, but the same file later being encountered will cause such a
+notification.
+
 Initial response:
 
     {
@@ -1093,6 +1101,7 @@ Initial response:
             { "path": "gen/moar/CORE.c.setting" },
             { "path": "NQP::src/how/Archetypes.nqp" },
             { "path": "SETTING::src/core.c/List.rakumod" },
+            { "path": "lib/ACME/Foobar.rakumod", "pending": True },
         ]
     }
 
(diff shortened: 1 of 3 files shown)


lizmat:
```
    "filenames": [
        { "path": "src/vm/moar/ModuleLoader.nqp" },
        { "path": "gen/moar/CORE.c.setting" },
        { "path": "NQP::src/how/Archetypes.nqp" },
        { "path": "SETTING::src/core.c/List.rakumod" },
    ]
```
Do you expect anything else but "path" as a key here?

If so, what could they be?

If not, why could "filenames" not be an array of strings?

timo:
I'm thinking later on, if it seems useful, each entry could include things like what bytecode files we've seen this filename in so far, maybe some information about frames from inside the file, if that doesn't belong in a different kind of request/response, and it'd probably be helpful to also have a list of line numbers that we've got some code for already, plus a way to tell if a line is present in multiple frames or just one (so far).

those are just the first things that come to my mind, but of course i've thought about this a little bit before

PR #1866, opened 2024-11-22
Counteract drift in timers with repeat

timo:
With regular repeating timers in libuv, we accumulate a small but significant amount of extra time with
every tick.

We only get to supply millisecond precision values for the timeout and repeat values, so we can
only correct the ticks by one millisecond or more, and we only notice when the difference is above that same amount as well.

Here's a test program to illustrate:

```
sub quick-stats(@vals) {
  my $num = +@vals;
  my $mean = ( @vals R/ @vals.sum ).fmt("%+2.8f");
  my @quartiles = |@vals.pairs.sort(*.value)[* div 4, * div 2, * * 3 div 4];
  my $min = @vals.min(:p).head; #.fmt("%4dth, %+2.8f");
  my $max = @vals.max(:p).head; #.fmt("%4dth, %+2.8f");

  say   "  $num values total";
  say   "    ranging from $min.value().fmt("%+2.8f") at idx $min.key().fmt("%4d")";
  say   "              to $max.value().fmt("%+2.8f") at idx $max.key().fmt("%4d")";
  say   "  with a mean of $mean.fmt("%+2.8f")";
  say   "    Quartiles:";
  for <25th 50th 75th> Z @quartiles {
    my $xxth = .[0];
    my $v = .[1];
    say "      $xxth      $v.value(  ).fmt("%+2.8f") at idx $v.key(  ).fmt("%4d")";
  }
}

for [0.1, 0.05] -> $interval {
  my $runtime = 3;
  say "===";
  say "Running test for $runtime<>s; intervals of $interval seconds";
  my @measurements;
  react {
    whenever Supply.interval($interval) {
      @measurements.push(now);
    }
    whenever Promise.in($runtime) { done }
  }
  my @diffs = ((@measurements X- @measurements[0]) Z- ((0..*) X* $interval)).skip(1);
  say "diff from expected absolute time";
  quick-stats(@diffs);
  my @intervals = @measurements.rotor(2=>-1).map({ .[1] - .[0] });
  say "time between consecutive ticks";
  quick-stats(@intervals);
  say "";
}
```

Output without this branch:

```
===
Running test for 30s; intervals of 0.1 seconds
diff from expected absolute time
  299 values total
    ranging from -0.00800413 at idx    0
              to +0.04615036 at idx  298
  with a mean of +0.01852002
    Quartiles:
      25th      +0.00438133 at idx   74
      50th      +0.01832324 at idx  149
      75th      +0.03275547 at idx  224
time between consecutive ticks
  299 values total
    ranging from +0.09199587 at idx    0
              to +0.10127693 at idx  150
  with a mean of +0.10015435
    Quartiles:
      25th      +0.10010626 at idx  129
      50th      +0.10013070 at idx  120
      75th      +0.10015709 at idx  199

===
Running test for 30s; intervals of 0.05 seconds
diff from expected absolute time
  598 values total
    ranging from -0.00010852 at idx    0
              to +0.08771799 at idx  597
  with a mean of +0.04322226
    Quartiles:
      25th      +0.01845824 at idx  149
      50th      +0.03989818 at idx  299
      75th      +0.06995165 at idx  449
time between consecutive ticks
  598 values total
    ranging from +0.04859222 at idx  381
              to +0.05269259 at idx  308
  with a mean of +0.05014669
    Quartiles:
      25th      +0.05004481 at idx   26
      50th      +0.05007415 at idx  549
      75th      +0.05014534 at idx  434
```

Output with this branch:

```
===
Running test for 30s; intervals of 0.1 seconds
diff from expected absolute time
  300 values total
    ranging from -0.00908117 at idx   88
              to -0.00700894 at idx   37
  with a mean of -0.00781980
    Quartiles:
      25th      -0.00808050 at idx  285
      50th      -0.00782713 at idx  170
      75th      -0.00755478 at idx  150
time between consecutive ticks
  300 values total
    ranging from +0.09270485 at idx    0
              to +0.10125536 at idx  105
  with a mean of +0.09997500
    Quartiles:
      25th      +0.10007666 at idx   15
      50th      +0.10012395 at idx   76
      75th      +0.10014862 at idx   83

===
Running test for 30s; intervals of 0.05 seconds
diff from expected absolute time
  600 values total
    ranging from -0.00123535 at idx   98
              to +0.00047485 at idx  367
  with a mean of -0.00033417
    Quartiles:
      25th      -0.00058419 at idx  250
      50th      -0.00034722 at idx   19
      75th      -0.00007597 at idx  373
time between consecutive ticks
  600 values total
    ranging from +0.04889808 at idx   48
              to +0.05118882 at idx   59
  with a mean of +0.04999986
    Quartiles:
      25th      +0.05003572 at idx  116
      50th      +0.05007477 at idx   82
      75th      +0.05011046 at idx  461
```

Observations:

* Without this branch:
  * we have fewer values than we would expect given the total test time
  * the "diff from expected absolute time" values
    * have their quartile values located evenly apart
    * the lowest value is at the first, and the highest value at the last index
    * → This is a line that slowly goes up.
    * reaches a value of up to 0.09 seconds after 30 seconds run time
      * up from almost 0 at the start, a range of 0.09s
  * the "time between consecutive ticks" values
    * have a minimum below the desired interval, but at least 75% of all values are above it
    * have the quartile values placed at no obviously significant positions
    * 75% of the values span a range of 0.00006s, all of them span a range of 1/10th the interval
    * the mean sits around the 75th percentile (sometimes above, sometimes below)
    * the mean is 0.00015 above (later than) our desired interval
* With this branch:
  * we have as many hits as the run time divided by the interval time
  * the "diff from expected absolute time" values
    * span a range of about 0.002 for 0.1s intervals and 0.001 for 0.05s intervals
    * have their quartiles at seemingly random positions
    * → some kind of jittery line that's mostly flat?
  * the "time between consecutive ticks" values
    * also have a minimum below the desired range and at least 75% of values are above it
    * uh-oh, the mean is below the 25th percentile, which i think means a few of the values are quite far below the mean?
    * span a range of 0.0023 for 0.05s interval and 0.0085 for 0.1s interval
      * 75% of the values span a range of 0.00007s, almost no difference from without the branch
    * the mean is 0.000025 below (earlier than) our desired interval (one zero more than without the branch!)

As you can see, I've abused this comment section for a bit of practice with stochastics :)

Anyway, I think it's better to be on average 1/10th as far off but being early rather than late for the tick, and as a result of that having 0.0005s accumulated after 30s instead of 0.09s, which is after all two zeroes more!

### DIFF
--- a/src/io/timers.c
+++ b/src/io/timers.c
@@ -7,6 +7,10 @@ typedef struct {
     uv_timer_t *handle;
     MVMThreadContext *tc;
     int work_idx;
+
+    /* In order to correct for clock drift in repeating timers, we adjust the
+     * repeat interval based on the last time our callback ran. */
+    MVMuint64 now_at_prev_callback;
 } TimerInfo;
 
 /* Frees the timer's handle memory. */
@@ -28,6 +32,29 @@ static void timer_cb(uv_timer_t *handle) {
         uv_close((uv_handle_t *)ti->handle, free_timer);
         MVM_io_eventloop_remove_active_work(tc, &(ti->work_idx));
     }
+    else if (ti->repeat) {
+        /* To counteract drift, we adjust the time based on the current time. */
+        MVMuint64 prev_now = ti->now_at_prev_callback;
+        MVMuint64 actually_passed = uv_now(handle->loop) - prev_now;
+        MVMint64 diff = actually_passed - ti->repeat;
+        if (diff * 2 > ti->repeat) {
+            /* We must have missed our tick for some reason.
+             * We will just pretend the timing was correct and not
+             * change anything. */
+            diff = 0;
+        }
+        else if (diff * -2 < -(MVMint64)ti->repeat) {
+            /* From the code in libuv's timer.c I think the timer callback
+             * can not be invoked if uv_now is not >= the due time, and the
+             * "clamped timeout" that is used is based on timeout + loop->time.
+             * In any case, even though this should not be possible, it's
+             * probably safe to just do nothing, if it becomes
+             * possible later down the line. */
+             diff = 0;
+        }
+        ti->now_at_prev_callback = uv_now(handle->loop) - diff;
+        uv_timer_start(handle, timer_cb, ti->repeat - diff, 0);
+    }
 }
 
 /* Sets the timer up on the event loop. */
@@ -38,7 +65,9 @@ static void setup(MVMThreadContext *tc, uv_loop_t *loop, MVMObject *async_task, 
     ti->work_idx     = MVM_io_eventloop_add_active_work(tc, async_task);
     ti->tc           = tc;
     ti->handle->data = ti;
-    uv_timer_start(ti->handle, timer_cb, ti->timeout, ti->repeat);
+    /* Instead of libuv's own repeating timer functionality, we take care of
+     * restarting the timer on every tick, so that we can correct for drift */
+    uv_timer_start(ti->handle, timer_cb, ti->timeout, 0);
 }
 
 /* Stops the timer. */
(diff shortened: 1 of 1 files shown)


timo:
main_vs_improved_plotted_intervals_005

I'm not sure i can explain why both main and my improved branch look so spiky with 0.05s interval

lizmat:
OOC, I have two questions:

- why aren't we using libuv's "tick" functionality?
- if that has an issue, shouldn't that be fixed in libuv?

lizmat (approved):

PR #1868, opened 2024-11-27
Windows long path support revival

patrickbkr:


patrickbkr:
Ping @ugexe, @MasterDuke17. This is https://github.com/MoarVM/MoarVM/pull/1780 rebased to latest main and the `rmdir` workarounds removed. I've tried some long path shenanigans on my Windows box. They work with this (and don't without).
The CI passed. I'm inclined to get this merged.

Last year you observed some flappiness around [`delete-by-compiler`](https://github.com/MoarVM/MoarVM/pull/1780#issuecomment-1834684344) / `rmdir`. I can't repro those locally. I'm pretty convinced that the `delete-by-compiler` issue is technically unrelated to the long paths change. It probably only changed the timing a little and thus surfaced the issue. If it shows up again, we can follow up.

I believe the best fix for that issue is adding retries in `delete-by-compiler` itself, not in Rakudo `rmdir`, nor MoarVM `rmdir`. `rmdir` itself is fine / innocent. Only when doing quick consecutive related deletes there is an issue. The [retry logic](https://nodejs.org/api/fs.html#fs_fs_rmdirsync_path_options) Node introduced only triggers when doing a recursive delete, they actually [retry the entire recursive delete](https://github.com/nodejs/node/blob/78855fdefee77ea13df9fa1887dc650b849a4f6d/lib/internal/fs/rimraf.js#L244), not only the individual failed `rmdir` request. They don't do any retries for individual deletes. Those would be on the user code. The issue at hand is user code (`delete-by-compiler`) doing a recursive delete "manually". Thus it's that functions responsibility to deal with interactions of those deletes.

ugexe:
I think my only concern would be in reference to various instances of this removed code:
```
    /* Must using UTF8ToUnicode for supporting CJK Windows file name. */
    wchar_t *wpathname = UTF8ToUnicode(pathname);
```
whereas I use
```
char *original_path = MVM_string_utf8_c8_encode_C_string(tc, path);
```
More specifically I did not explicitly handle wide characters in paths.

I do not know if these changes would regress on the supporting of CJK Windows file names.

patrickbkr:
I'm pretty sure this is fine. What I know:

- wchar_t is two bytes. Effectively UTF16.
- All the Windows API is dualistic. Functions ending in "A" deal with `char`s, ending in "W" deal with `wchar_t`. Functions without a trailing letter use one or the other depending on a compile time setting. That same setting switched out the `char` type as well. This mechanism was used in the early Unicode days to write programs that could either work with codepages or be switched over to the shiny new Unicode world.
- In the old times the "A" functions were exclusively dealing with single byte codepages.
- In the newer times Windows can be instructed to repurpose the "A" functions to do UTF8. I seem to recall it's a behavior individual executables can opt into and you additionally have to manually enable this system wide. This is experimental and not universally working. In a decade or so this will probably the way to go on Windows.
- So currently the only robust way to do Unicode is `wchar_t` and the "W" functions. Albeit this does not support the full Unicode range as it's using the UTF16 encoding.
- The libuv API is utf8 all the way. (Under the hood it's converting to `wchar_t` again on Windows.)
- `UTF8ToUnicode` simply converts UTF8 `char*` to `wchar_t*` which you need to do when you want to work with the "W" functions.

So as long as we use the libuv functions exclusively, everything is fine.

ugexe (approved):
I would generally be hesitant to approve my own changes, but I'll do so anyway to signal that I think we should be ok with this being merged

PR #1870, opened 2024-12-01
Fix a compile warning on cl / Windows

patrickbkr:
The previous check was wrong, adding "-fwrapv" whenever you did not specify "--os" on the command line. That option is unknown to `cl`. It seems `cl` doesn't recognize "-d2UndefIntOverflow-" (nor some variants I have tried) either. So the best we can do is just not adding any option of the like.

I do wonder if this has negative stability consequences.

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -466,7 +466,7 @@ $config{lddebugflags} = sprintf $config{lddebugflags}, defined_or $args{debug}, 
 # generate CFLAGS
 my @cflags;
 push @cflags, '-std=c99' if $defaults{os} eq 'mingw32';
-push @cflags, $args{os} eq 'MSWin32' ? '-d2UndefIntOverflow-' : '-fwrapv';
+push @cflags, '-fwrapv'             if $config{cc} ne 'cl';
 push @cflags, $config{ccmiscflags};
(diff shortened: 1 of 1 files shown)


patrickbkr:
@MasterDuke17 How bad is it that MSVC can't do the int overflow thing?

MasterDuke17:
> @MasterDuke17 How bad is it that MSVC can't do the int overflow thing?

Well, we've survived without it explicitly doing it so far. If the option/flag just doesn't exist then I guess we just leave it off and hope for the best. But does `/d2UndefIntOverflow` or `/d2UndefIntOverflow-`/ work?

MasterDuke17 (approved):

patrickbkr:
I've tried:

- `/d2UndefIntOverflow`
- `-d2UndefIntOverflow-`
- `/d2UndefIntOverflow/`
- `/-d2UndefIntOverflow-/`
- `/d2UndefIntOverflow-/`
- `/-d2UndefIntOverflow`

Maybe some more.

The error usually is:

```
fatal error C1007: unrecognized flag '-UndefIntOverflow-' in 'p2'
```

This indicates it did parse the flag somehow. So I suspect the format is right, but the flag is just not recognized.

PR #1879, opened 2024-12-23
Use 64 bit ints in decoder

ugexe:
Currently when slurping a 2gb file MoarVM will panic trying to allocate an impossible amount of memory. This is due to various length type values being 32 bit and overflowing with large files. This commit changes such values to instead be 64 bit.

Resolves https://github.com/rakudo/rakudo/issues/5727

### DIFF
--- a/src/strings/decode_stream.c
+++ b/src/strings/decode_stream.c
@@ -27,7 +27,7 @@ MVMDecodeStream * MVM_string_decodestream_create(MVMThreadContext *tc, MVMint32 
 }
 
 /* Adds another byte buffer into the decoding stream. */
-void MVM_string_decodestream_add_bytes(MVMThreadContext *tc, MVMDecodeStream *ds, MVMuint8 *bytes, MVMint32 length) {
+void MVM_string_decodestream_add_bytes(MVMThreadContext *tc, MVMDecodeStream *ds, MVMuint8 *bytes, MVMint64 length) {
     if (length > 0) {
         MVMDecodeStreamBytes *new_bytes = MVM_calloc(1, sizeof(MVMDecodeStreamBytes));
         new_bytes->bytes  = bytes;
@@ -45,7 +45,7 @@ void MVM_string_decodestream_add_bytes(MVMThreadContext *tc, MVMDecodeStream *ds
 }
 
 /* Adds another char result buffer into the decoding stream. */
-void MVM_string_decodestream_add_chars(MVMThreadContext *tc, MVMDecodeStream *ds, MVMGrapheme32 *chars, MVMint32 length) {
+void MVM_string_decodestream_add_chars(MVMThreadContext *tc, MVMDecodeStream *ds, MVMGrapheme32 *chars, MVMint64 length) {
     MVMDecodeStreamChars *new_chars;
     if (ds->chars_reuse) {
         new_chars = ds->chars_reuse;
@@ -74,7 +74,7 @@ static void free_chars(MVMThreadContext *tc, MVMDecodeStream *ds, MVMDecodeStrea
 }
 
 /* Throws away byte buffers no longer needed. */
-void MVM_string_decodestream_discard_to(MVMThreadContext *tc, MVMDecodeStream *ds, const MVMDecodeStreamBytes *bytes, MVMint32 pos) {
+void MVM_string_decodestream_discard_to(MVMThreadContext *tc, MVMDecodeStream *ds, const MVMDecodeStreamBytes *bytes, MVMint64 pos) {
     while (ds->bytes_head != bytes) {
         MVMDecodeStreamBytes *discard = ds->bytes_head;
         ds->abs_byte_pos += discard->length - ds->bytes_head_pos;
@@ -306,7 +306,7 @@ static MVMString * take_chars(MVMThreadContext *tc, MVMDecodeStream *ds, MVMint3
     return result;
 }
 MVMString * MVM_string_decodestream_get_chars(MVMThreadContext *tc, MVMDecodeStream *ds,
-                                              MVMint32 chars, MVMint64 eof) {
+                                              MVMint64 chars, MVMint64 eof) {
     MVMuint32 missing;
 
     /* If we request nothing, give empty string. */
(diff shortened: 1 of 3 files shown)


ugexe:
I'm not sure if all of these changes are sufficient or correct. Technically the minimum I needed to resolve the issue mentioned in the OP is:
```
diff --git a/src/strings/decode_stream.c b/src/strings/decode_stream.c
index 5e5f6cc87..adc5e5bc7 100644
--- a/src/strings/decode_stream.c
+++ b/src/strings/decode_stream.c
@@ -497,7 +497,7 @@ static MVMString * get_all_in_buffer(MVMThreadContext *tc, MVMDecodeStream *ds)
     /* Otherwise, need to assemble all the things. */
     else {
         /* Calculate length. */
-        MVMint32 length = 0, pos = 0;
+        MVMint64 length = 0, pos = 0;
         MVMDecodeStreamChars *cur_chars = ds->chars_head;
         while (cur_chars) {
             if (cur_chars == ds->chars_head)
@@ -516,7 +516,7 @@ static MVMString * get_all_in_buffer(MVMThreadContext *tc, MVMDecodeStream *ds)
         while (cur_chars) {
             MVMDecodeStreamChars *next_chars = cur_chars->next;
             if (cur_chars == ds->chars_head) {
-                MVMint32 to_copy = ds->chars_head->length - ds->chars_head_pos;
+                MVMint64 to_copy = ds->chars_head->length - ds->chars_head_pos;
                 memcpy(result->body.storage.blob_32 + pos, cur_chars->chars + ds->chars_head_pos,
                     to_copy * sizeof(MVMGrapheme32));
                 pos += to_copy;
```
but I noticed we using 32 bits in other areas that seemed potentially problematic and changed those as well.

lizmat:
Wouldn't that break 32bit builds?  Or don't we support those anymore anyway?

ugexe:
> Wouldn't that break 32bit builds? Or don't we support those anymore anyway?

Not sure. There are already other instances of `MVMint64`, for example https://github.com/MoarVM/MoarVM/blob/f4680765ff33c5815a8a9ae183ed08cce6cc6943/src/strings/decode_stream.h#L105

niner:
> Wouldn't that break 32bit builds?

No, even 32 bit platforms have 64 bit integer types.

src/strings/decode_stream.c
```diff
@@ -585,7 +585,7 @@ MVMint64 MVM_string_decodestream_bytes_available(MVMThreadContext *tc, const MVM
 /* Copies up to the requested number of bytes into the supplied buffer, and
  * returns the number of bytes we actually copied. Takes from from the start
  * of the stream. */
-MVMint64 MVM_string_decodestream_bytes_to_buf(MVMThreadContext *tc, MVMDecodeStream *ds, MVMuint8 **buf, MVMint32 bytes) {
+MVMint64 MVM_string_decodestream_bytes_to_buf(MVMThreadContext *tc, MVMDecodeStream *ds, MVMuint8 **buf, MVMint64 bytes) {
     MVMint32 taken = 0;
```
timo, line 589: I think `taken`, as well as `required` and `available` also need to be 64bit

src/strings/decode_stream.c
```diff
@@ -555,7 +555,7 @@ MVMString * MVM_string_decodestream_get_available(MVMThreadContext *tc, MVMDecod
 }
 
 /* Checks if we have the number of bytes requested. */
-MVMint64 MVM_string_decodestream_have_bytes(MVMThreadContext *tc, const MVMDecodeStream *ds, MVMint32 bytes) {
+MVMint64 MVM_string_decodestream_have_bytes(MVMThreadContext *tc, const MVMDecodeStream *ds, MVMint64 bytes) {
     MVMDecodeStreamBytes *cur_bytes = ds->bytes_head;
     MVMint32 found = 0;
```
timo, line 560: needs to be 64 I think

timo: Also it looks like I can't add comments to lines that aren't close to a line you've changed, but in MVM_string_decodestream_bytes_available we also need `available` as 64bit, that's line 575

timo (changes requested):
we are using the value in `result_size_guess` in many decoder implementations by putting them into 32 bit integers (unsigned integers in at least some places). I think it would be safe to keep result_size_guess as 32bit and make sure we don't put more than fits into a 32bit signed integer into it, then we don't have to change all the codec impls to match and never overflow.

We have some places where we are using 32bit integers to compare against absolute positions, for example

https://github.com/MoarVM/MoarVM/pull/1879/files#diff-3826a2d65017341f7f3b68b7ff86475e2fcf2ddc7032b5ec2a175034a39e9f54R179

https://github.com/MoarVM/MoarVM/pull/1879/files#diff-3826a2d65017341f7f3b68b7ff86475e2fcf2ddc7032b5ec2a175034a39e9f54R589-R595


I think functions like take_chars also need to have their insides and sometimes return values / arguments upgraded?


I think the sep_loc is a position inside the DecoderBytes or something, so also needs to be 64bit, at least in MVM_string_decodestream_get_until_sep_eof it looks to be the case, and if that's right, it'd also need changes in find_separator and so on.

OTOH, i think this should stay 32bit to match num_seps in MVMDecodeStreamSeparators (if someone uses 2**32 different sequences to separate their input on, maybe they are using the wrong hammer for that particular thumb):

https://github.com/MoarVM/MoarVM/pull/1879/files#diff-9588ebd4fec1731ba05dd32874bebf7b36115e4ca5ea6d02f2025b766cf87b42R97

I'm not sure if there's a flag for GCC or an __attribute__ or whatever that we could use to make sure the compiler spots cases where we mix 32bit and 64bit values related to the structs in question here, maybe without spewing boatloads of warnings that we don't care about at the moment.

timo:
I just wrote this on IRC, I figured it should also be in this ticket:

> we might actually be able to get away with doing a different change, however
> we have these MVMDecodeStreamBytes and MVMDecodeStreamChars structs that have 32bit size fields. Maybe when we encounter something that's too big, we simply split it into as many parts as needed
> then not so much code has to change, I think
> however, there's some places that still use 32bit ints to represent "position in the total data", or at least from a first glance it looks like they are. maybe those are actually all refering to inside a single DecodeStream*s and it's fine to keep them at 32?
> maybe the "minimum necessary change" in your first comment is closer to "the right solution" after all
> so bytes_head_pos and chars_head_pos as well as the length members of the MVMDecodeStream*s could probably stay 32bit without too much trouble

PR #1884, opened 2025-01-06
Micro-optimize P6opaque's get_attribute()

MasterDuke17:
Instead of checking `attr_st` inside each case of the `switch(kind)`, have an if/else for attr_st with a switch in each branch.

On this laptop, it seemed to reduce the instructions and time for compiling CORE.d and CORE.e (I didn't try CORE.c since that would have taken way too long under valgrind). I don't have access to my x86 desktop right now, and since this optimization seems like it will very likely be compiler-dependent, maybe @timo you could test it?

### DIFF
--- a/src/6model/reprs/P6opaque.c
+++ b/src/6model/reprs/P6opaque.c
@@ -266,10 +266,50 @@ static void get_attribute(MVMThreadContext *tc, MVMSTable *st, MVMObject *root,
         try_get_slot(tc, repr_data, class_handle, name);
     if (slot >= 0) {
         MVMSTable *attr_st = repr_data->flattened_stables[slot];
-        switch (kind) {
-        case MVM_reg_obj:
-        {
-            if (!attr_st) {
+        if (attr_st) {
+            switch (kind) {
+            case MVM_reg_obj: {
+                MVMROOT2(tc, root, attr_st) {
+                    /* Need to produce a boxed version of this attribute. */
+                    MVMObject *cloned = attr_st->REPR->allocate(tc, attr_st);
+
+                    /* Ordering here matters too. see comments above */
+                    result_reg->o = cloned;
+                    attr_st->REPR->copy_to(tc, attr_st,
+                        (char *)MVM_p6opaque_real_data(tc, OBJECT_BODY(root)) + repr_data->attribute_offsets[slot],
+                        cloned, OBJECT_BODY(cloned));
+                }
+                break;
+            }
+            case MVM_reg_int64: {
+                result_reg->i64 = attr_st->REPR->box_funcs.get_int(tc, attr_st, root,
+                    (char *)data + repr_data->attribute_offsets[slot]);
+                break;
+            }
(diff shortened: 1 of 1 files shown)


MasterDuke17:
Huh, no CI?

lizmat:
"Workflow runs completed with no jobs"  ??   @patrickbkr @timo ??

timo:
sorry my fault for the commit i accidentally pushed to the main branch yesterday

MasterDuke17:
Also, here's a little bit of logging of what happens during a CORE.c compile (8 is obj, 4 is int64, and 7 is str).
```
16452991 non attr_st, kind = 8
12195638 yes attr_st, kind = 4
4250708 yes attr_st, kind = 7
 494754 yes attr_st, kind = 8
   9046 yes attr_st, kind = 20
    109 yes attr_st, kind = 6
```
I believe a lot of the calls to `get_attribute()` are coming from e.g., `getattr_i` in interp.c. I wonder if it would make sense to create `get_attribute_[oisun]()` functions? `attr_funcs` might have to be made a bit bigger though, I don't know all the repercussions of that.

MasterDuke17:
I know the jit can devirtualize these in some instances, but since I don't have the jit on this hardware, I'm now wondering if spesh can be made to do anything.

timo:
spesh can turn getattr_* into different kinds of sp_get* instructions, which don't require jit compilation to give a nice speedup.

we could look at spesh logs to find places where the getattr_* remains in the "After" result to see if we spot anything that sticks out that we could optimize better?

there's also the case of getattr with a string for the attribute name, in which case spesh can't do anything if the attribute name isn't known at spesh time, though in theory we could have strings that have been seen logged and code-gen a string comparison and "specialized" branch with an optimized getattr variant. Would need to do some measurements to see what the opportunities are

MasterDuke17:
> we could look at spesh logs to find places where the getattr_* remains in the "After" result to see if we spot anything that sticks out that we could optimize better?

I have a 1.6gb spesh log of compiling CORE.c. There are ~2.3k getattr_* remaining in the 'After:' of >300 functions. Here's a random example from 'MATCH':
```
  BB 8 (0x340e0216ec8):
    line: 1399 (pc 984)
    Instructions:
      getattr_s        r6(13),  r25(4), r10(31), lits($!name), liti16(-1)
      isnull_s        r24(10),  r6(13)
      if_i            r24(10),  BB(12)
    Successors: 12, 9
    Predecessors: 7
    Dominance children: 9, 12
```

MasterDuke17:
The count of the types of getattr:
```
   1242 getattr_i
   1544 getattr_o
     48 getattr_s
      4 getattr_u
```

lizmat:
Is this PR good to go as is?

MasterDuke17:
> Is this PR good to go as is?

I think so. I believe it'll also generate slightly faster code on x86, but we could always merge now and I'll confirm that when I have access to my desktop again in a couple days.

PR #1889, opened 2025-01-16
Add an Ubuntu ARM64 runner to CI

MasterDuke17:


lizmat:
cancelled is not failed, right?

MasterDuke17:
Hm, https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/ seems to suggest it should be available?

timo:
that's only for github actions, not for azure pipelines :(

MasterDuke17:
> that's only for github actions, not for azure pipelines :(

I've never really understood the difference....

PR #1890, opened 2025-01-16
Bump mimalloc to v2.1.9

MasterDuke17:


MasterDuke17:
Looks like the Windows fail was fixed recently (https://github.com/microsoft/mimalloc/issues/984), so closing this and will try 2.1.10 whenever it's released.

PR #1894, opened 2025-01-20
Configure dyncall for arm64 on Windows

ugexe:


ugexe:
Note that the `/target-arm64` flag is not available with the version of dyncall we currently vendor. It was introduced in https://github.com/ugexe/dyncall/commit/c06b131325595add43568774bf0dddb04c9257e8#diff-b5e11e4ac62ecced8feac9095d90ae2a14627f91e8b5db85ac44547804f286e1R58

ugexe:
It is also possible that this detection is not needed and that the updated dyncall configure would Do The Right Thing using its own detection. When/if dyncall gets updated I'll test if this PR is actually needed.

PR #1896, opened 2025-01-20
Use rapidhash instead of siphash for string hashes

MasterDuke17:
It's currently the fastest recommended hash function according to SMHasher.

Valgrind reports that the instructions for `raku -e 'my %h; for ^100_000 -> int $i { %h{$i}++ };'` drop from ~1.50b to ~1.48b.


coke:
Let's wait on this until after the release this weekend.

MasterDuke17:
> Let's wait on this until after the release this weekend, and I can do a blin run as soon as it's in.

That’s what I assumed, there’s no urgency for this.

PR #1901, opened 2025-01-26
Update many Perl 6 -> Rakudo

coke:
Inspired by the "Rakudo Perl 6" on the moarvm.org site.


coke:
... and add a small note to the release guide, sorry

PR #1903, opened 2025-01-26
Make accesses to spesh arg guard atomic

timo:
On ARM and other platforms with less strict memory ordering guarantees than x86, the step that spesh does to switch the pointer around to the newly created spesh arg guard can be seen by another thread before the contents of the arg guard are seen, causing the thread to jump into uninitialised "code". Until recently, this just caused freezing or random behaviour. Then we got a panic or oops in moar when the code "jumped" out of bounds, which is both better and worse.

This is a real fix for the issue.

Need to check if the performance impact on x86_64 is significant. In that case, for that arch, we could conditionally use regular access instead of atomic ops. Alternatively the switch-over of the spesh arg guard pointer could be delayed, but then I don't know how to guarantee the new spesh arg guard contents are really visible at that time.

### DIFF
--- a/src/spesh/arg_guard.c
+++ b/src/spesh/arg_guard.c
@@ -655,6 +655,6 @@ void MVM_spesh_arg_guard_discard(MVMThreadContext *tc, MVMStaticFrame *sf) {
     MVMStaticFrameSpesh *spesh = sf->body.spesh;
     if (spesh && spesh->body.spesh_arg_guard) {
         MVM_spesh_arg_guard_destroy(tc, (MVMSpeshArgGuard *)MVM_load(&spesh->body.spesh_arg_guard), 1);
-        MVM_store(spesh->body.spesh_arg_guard, NULL);
+        MVM_store(&spesh->body.spesh_arg_guard, NULL);
     }
 }
--- a/src/spesh/arg_guard.c
+++ b/src/spesh/arg_guard.c
@@ -276,7 +276,7 @@ static MVMuint32 add_nodes_for_typed_argument(MVMThreadContext *tc,
             if (additional_update_node != -1)
                 tree->nodes[additional_update_node].no = certain_fallback;
         }
-        
+
         /* Clean up. */
         for (i = 0; i < MVM_VECTOR_ELEMS(by_type); i++)
             MVM_VECTOR_DESTROY(by_type[i].cand_idxs);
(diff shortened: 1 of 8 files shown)


src/spesh/arg_guard.c
```diff
@@ -383,12 +383,18 @@ void MVM_spesh_arg_guard_regenerate(MVMThreadContext *tc, MVMSpeshArgGuard **gua
 
     /* Install the produced argument guard. */
     if (*guard_ptr) {
-        MVMSpeshArgGuard *prev = *guard_ptr;
-        *guard_ptr = tree;
+        /* Ensure that the tree is already fully out of our cache.
+         * This is not necessary on x86_64, which has
+         * Total Store Ordering, but architectures like ARM are weaker
+         * and other cores can observe older contents of the  */
```
MasterDuke17, line 389: Missing the end of the sentence.

src/spesh/arg_guard.c
```diff
@@ -647,7 +653,9 @@ void MVM_spesh_arg_guard_destroy(MVMThreadContext *tc, MVMSpeshArgGuard *ag, MVM
 void MVM_spesh_arg_guard_discard(MVMThreadContext *tc, MVMStaticFrame *sf) {
     MVMStaticFrameSpesh *spesh = sf->body.spesh;
     if (spesh && spesh->body.spesh_arg_guard) {
-        MVM_spesh_arg_guard_destroy(tc, spesh->body.spesh_arg_guard, 1);
-        spesh->body.spesh_arg_guard = NULL;
+        /* No need to atomically exchange with NULL here, since actual freeing
+         * of data is postponed to the next safe point. */
+        MVM_spesh_arg_guard_destroy(tc, (MVMSpeshArgGuard *)MVM_load(&spesh->body.spesh_arg_guard), 1);
+        MVM_store(spesh->body.spesh_arg_guard, NULL);
```
MasterDuke17, line 659: If there's "No need to atomically exchange with NULL here ...", then why the `MVM_store(spesh->body.spesh_arg_guard, NULL);` instead of just `spesh->body.spesh_arg_guard = NULL;`?

src/spesh/arg_guard.c
```diff
@@ -383,12 +383,18 @@ void MVM_spesh_arg_guard_regenerate(MVMThreadContext *tc, MVMSpeshArgGuard **gua
...
+         * Total Store Ordering, but architectures like ARM are weaker
+         * and other cores can observe older contents of the  */
+        MVM_barrier();
+        MVMSpeshArgGuard *prev = (MVMSpeshArgGuard *)MVM_load(guard_ptr);
+        MVM_store(guard_ptr, tree);
         MVM_spesh_arg_guard_destroy(tc, prev, 1);
     }
     else {
-        *guard_ptr = tree;
+        MVM_barrier();
```
MasterDuke17, line 396: Why is this barrier needed? I can see in the other branch when there's both a load and a store, but there's just a store here and guard_ptr and tree aren't accessed after.

src/spesh/arg_guard.c
timo: removed the comment, it wasn't really helpful

src/spesh/arg_guard.c
MasterDuke17: From IRC "timo: the stores that we are worried about are all the stores to the tree->nodes[i] further above".

PR #1906, opened 2025-02-01
Reduce allocations

MasterDuke17:
This reduces the "temporary" allocations during a build of Rakudo's CORE.c by ~900k.


MasterDuke17:
I'm only not making this a draft because they don't run CI. This should definitely get a look for safety before it's merged.

lizmat:
Looks like CI is pretty red :-(

PR #1907, opened 2025-02-02
Maybe optimize MVM_string_latin1_decode()

MasterDuke17:
Do some checking up front to remove branches from loop bodies. Write to in-situ strings directly when able, instead of always allocating and then copy+free when done decoding.

This appears to reduce the total number of allocations done by `MVM_string_latin1_decode()` from ~533k to ~39k during a build of Rakudo's CORE.c (with MVM_SPESH_BLOCKING=1).

The loop to check whether it has to be 32-bit doesn't seem to get vectorized on this aarch64 laptop (if I added `MVM_LOOP_VECTORIZE` then clang complained `warning: loop not vectorized: the optimizer was unable to perform the requested transformation; the transformation might be disabled or specified as part of an unsupported transformation ordering [-Wpass-failed=transform-warning]`). Maybe it would be on x86-64?

### DIFF
--- a/src/strings/latin1.c
+++ b/src/strings/latin1.c
@@ -1,5 +1,16 @@
 #include "moar.h"
 
+#define DECODE_BODY                                                          \
+    for (i = 0; i < bytes; i++) {                                            \
+        if (latin1[i] == '\r' && i + 1 < bytes && latin1[i + 1] == '\n') {   \
+            storage[result_graphs++] = MVM_nfg_crlf_grapheme(tc);            \
+            i++;                                                             \
+        }                                                                    \
+        else {                                                               \
+            storage[result_graphs++] = latin1[i];                            \
+        }                                                                    \
+    }                                                                        \
+
 /* Decodes the specified number of bytes of latin1 into an NFG string,
  * creating a result of the specified type. The type must have the MVMString
  * REPR. */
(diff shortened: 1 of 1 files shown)


timo:
I think if we check before the decode loop that the very last byte is not `'\r'`, then we can remove the `&& bytes > i + 1 &&` piece, maybe that makes the vectorizer happier?

timo:
i'm thinking we would then have to either go to bytes - 1 (or more likely, bytes-- in front and bytes++ after) if we detect that, or maybe just always do the loop until bytes - 1 and to the very last byte separately (unless we did skip it because of crlf)?

MasterDuke17:
I was thinking of this loop (https://github.com/MoarVM/MoarVM/pull/1907/files#diff-0ad55191161689df28c2d7ea91da91c021715682293e6306e85c5f04b6c36551R28-R33), but maybe the break is throwing it off. Depending on how often and where 32-bit chars are, maybe it's faster to process the whole buffer if it can do so with vectorized code.

MasterDuke17:
> I was thinking of this loop (https://github.com/MoarVM/MoarVM/pull/1907/files#diff-0ad55191161689df28c2d7ea91da91c021715682293e6306e85c5f04b6c36551R28-R33), but maybe the break is throwing it off. Depending on how often and where 32-bit chars are, maybe it's faster to process the whole buffer if it can do so with vectorized code.

Yeah, it vectorizes with both clang and gcc if I change the body to just `writing_32bit |= (latin1[i] > 127);` (at least on my x86-64 desktop, I can't read the aarch64 assembly well enough to tell). Not exactly sure how to benchmark it though.

MasterDuke17:
Commit [743ba42](https://github.com/MoarVM/MoarVM/pull/1907/commits/743ba42384893811f871c3117609708909a02cb1) was faster on both my laptop and my desktop when looping through and decoding ~286k strings (taken from `.words` on the CORE.c setting).

MasterDuke17:
Any thoughts about merging this?

PR #1908, opened 2025-02-02
Convert old perl6 refs to Raku

coke:


MasterDuke17 (approved):

MasterDuke17:
Does this replace #1901?

PR #1912, opened 2025-02-12
Simplify decoding error reporting function

MasterDuke17:
We're already in an error/slow path, let's just simplify the code a little.


src/strings/utf8.c
```diff
@@ -169,32 +169,24 @@ static MVMint32 utf8_encode(MVMuint8 *bp, MVMCodepoint cp) {
 static void encoding_error(MVMThreadContext *tc, const MVMuint8 *bytes, int error_pos, int line, int col) {
                     /* Different error messages for the first few bytes, so that
                      * we print as much context as possible */
+    char location[35] = "";
+    if (line >= 0)
+        sprintf(location, " at line %u col %u", line, col);
```
niner, line 174: This can be a buffer overflow if sizeof(int) > 4 which it is allowed to be.

PR #1913, opened 2025-02-13
Re-instate lookup of synthetic codepoint info for find_cclass

timo:
It got left out accidentally. Since GRAPHEME_8 and IN_SITU_8 can store negative numbers, and especially newlines are a case of this because of the crlf synth, we need to go through the synthetic info lookup.

Fixes:

```
21:17 <timo> m: .raku.say for "Hello there\r\nHow are you".encode("latin-1").decode("latin-1").lines
21:17 <camelia> rakudo-moar 62c83711d: OUTPUT: «"Hello there\r\nHow are you"␤»
```

This was found in HTTP::Tiny's test suite.

We should maybe do something automated that double-checks everything we do with strings in the test suite with the different storage kinds?


MasterDuke17:
Nice catch, thanks.

PR #1918, opened 2025-03-01
MVM_platform_sleep takes no tc argument

timo:
This was causing the amount of seconds to land in the xmm1 register on windows, where the function wasn't looking, and therefore any loop that was attempting to run "slowly" by calling sleep to accidentally sleep by whatever happened to still be in that register, which turned out to be 0.0, which is roughly equivalent to a thread_yield operation.

On linux, the first float argument seen in a signature would always go into xmm0, which is why the jitted result was still behaving correctly.

Thanks to @zhuomingliang++ for tracking this down in https://github.com/rakudo/rakudo/issues/5789


zhuomingliang:
Good fix, I didn't notice the sleep should take no `tc` argument in the jit.

PR #1921, opened 2025-03-07
Don't turn slurpies into huge amounts of code

timo:
For example, a file with a huge hash literal created with the `infix:<,>` operator may end up with a callsite with thousands of entries, and peak memory usage from register facts and other stuff can balloon quite massively.

https://github.com/lizmat/Text-Emoji/issues/2 encountered a peak memory usage from MVM_spesh_args of over 3 gigabytes from "just" 5687 arguments to `infix:<,>`. Refusing to turn the slurpy into individual instructions gets the total process memory peak to go from 3,332,068k down to 196,852k which is a lot more reasonable.

### DIFF
--- a/src/spesh/args.c
+++ b/src/spesh/args.c
@@ -261,12 +261,12 @@ static void slurp_positional_arg(MVMThreadContext *tc, MVMSpeshGraph *g, MVMSpes
     /* Look up arg flags and name, and compute index. */
     MVMCallsiteFlags flags = g->cs->arg_flags[pos_idx];
 
-    /* Allocate temporary registers for the key and value. */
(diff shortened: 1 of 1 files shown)


MasterDuke17 (approved):

niner (approved):

lizmat:
FWIW, I cannot reproduce this gigantic memory usage on my M1 dev machine.  How did you get that number?

lizmat:
Turns out that on MacOS `snapper` was reporting memory sizes that were 1000x to small.  Fixed with https://github.com/rakudo/rakudo/commit/d02733a40b

PR #1923, opened 2025-03-11
Add 32bit/64bit native int write support

lizmat:
Following up on https://github.com/rakudo/rakudo/issues/5776


lizmat:
Also see discussion at https://irclogs.raku.org/moarvm/2025-03-11.html#14:15

lizmat:
Are there still objections to merging this?

PR #1924, opened 2025-03-14
In regular vm exit, stop and join eventloop

MasterDuke17:
This prevents stuff from happening on the event loop thread while mimalloc is destroying its structures in its atexit handler, which can otherwise cause aborts and other undesired behaviour.

### DIFF
--- a/src/moar.c
+++ b/src/moar.c
@@ -608,6 +608,11 @@ void MVM_vm_exit(MVMInstance *instance) {
     MVM_thread_join_foreground(instance->main_thread);
     MVM_io_flush_standard_handles(instance->main_thread);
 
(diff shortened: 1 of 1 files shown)


MasterDuke17:
I have run a *bunch* of spectests in a loop and this improves stability.

MasterDuke17:
> I have run a _bunch_ of spectests in a loop and this improves stability.

Ha! I say this and it gets a fail in t/02-rakudo/15-gh_1202.t (though this wasn't one where I saw an improvement).

niner:
15-gh_1202.t has been troublesome for years though. If I remember correctly it even fails most of the time on s390.

PR #1931, opened 2025-04-10
Revert fallback resolver

zhuomingliang:
This is replaced by https://github.com/Rakudo/Rakudo/commit/639f20a2218c418a94e1def03b922cb4a0690795

Just create this pull request first, could be merged once it's confirmed.


niner:
The :deoptallpoint flags are still an improvement and independent of the fallback resolver.

zhuomingliang:
> The :deoptallpoint flags are still an improvement and independent of the fallback resolver.

Thanks for review, I have force-pushed it.

niner (approved):

PR #1934, opened 2025-04-22
Have to gc-mark unwind data's payload attribute

timo:
Otherwise the pointer can become stale, as grondilu++ found out the hard way.

Couldn't come up with a short reproducing example. The important bit seems to be having a frame with a LEAVE block, and a payload has to already have been stored in the tc before the special return is installed (by the presence of a LEAVE block if I'm not mistaken)

In any case, I believe this commit is correct and necessary.


lizmat:
That looks pretty red in CI to me  :-(

timo:
```
2025.03-18-g2bd1e322f, which is too old. Wanted at least 2025.04
 at /home/vsts/work/1/nqp/3rdparty/nqp-configure/lib/NQP/Config.pm line 192.
 ```
 
 grmbl.

MasterDuke17 (approved):

PR #1936, opened 2025-04-28
Draft: Add a C library for JSON parsing tasks

ab5tract:
This is probably controversial and also very much a WIP and open to discussion.

Early performance checks have yielded timings of about 1/5 the nqp library, and we have done very little in terms of optimization.

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -452,6 +452,10 @@ else {
     $config{mimalloc_object} = "";
 }
 
+# TODO error handling or something
+setup_native_library('yyjson'); # if $config{pkgconfig_works};
+unshift @{$config{usrlibs}}, 'yyjson';
+
 # mangle library names
 $config{ldlibs} = join ' ',
     $config{lincludes},
--- a/src/disp/syscall.c
+++ b/src/disp/syscall.c
@@ -1685,16 +1685,23 @@ static MVMObject *yy_to_mvm(MVMThreadContext *tc, yyjson_val *val) {
             }
             return result;
         }
-        case YYJSON_TYPE_NUM:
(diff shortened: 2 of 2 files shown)


ab5tract:
> OOC, does the JVM come with similar capabilities?

IIUC, it should be straightforward to achieve similar gains in JVM by using Jackson or another JSON parsing library.

lizmat:
FWIW, with JSON becoming more and more important (e.g. in SBOM land, but also more generally with "lazy" hacks to work around JSON parsing during the module loading process) I feel it would be very important to have *fast* way of parsing JSON into Raku data structures.

So what would be needed to get this PR afloat again?

lizmat:
> There also seemed to be some question as to whether putting this into core was a good idea, but my memory is a bit fuzzy on that.

Well, there is.  But recently I learned that there is some pretty hacky stuff in the Compunit loading that appears to be just there to handle the slowness of JSON parsing (with JSON::Fast).  And the SBOM work I'm doing at the moment, also needs fast JSON parsing.

And zef would benefit from faster JSON parsing.

Mind you, if you find the time to work on this, make sure there's an "immutable" option built in, that would create an nqp::hash *without* any containers.  That in itself was good for a 5% speed increase, and 9% less memory use (for a 6MB JSON file).  Note that in that respect, the REA index is currently at ~15MB and the zef index is at ~7MB.

niner:
> But recently I learned that there is some pretty hacky stuff in the Compunit loading

What stuff is that?

lizmat:
> What stuff is that?

https://github.com/rakudo/rakudo/blob/main/src/core.c/CompUnit/Repository/Installation.rakumod#L42-L93

PR #1937, opened 2025-04-28
Draft: Alternative iterative version of the yyjson integration

ab5tract:
This currently fails at larger datasets.


ab5tract:
No convincing need to pursue this iteration based approach at this time.

PR #1943, opened 2025-06-08
Remove libatomic_ops

Leont:
This removes the usage of libatomic_ops, in favor of C11 atomics (Added in 7ff15b3e). The main expected impact of this is that it increases the minimum supported version of MSVC to 2022.

This simplifies a bunch of code, and allows me to make a bunch of improvements (clarification and performance) as a follow up.


Configure.pl
ugexe, line 1: There are still various references to `c11` in this file that can be removed such as the command line arguments `--[no-]c11-atomics`

PR #1942, opened 2025-05-24
Update Ubuntu images for CI

MasterDuke17:

PR #1945, opened 2025-06-08
Upgrade to Ubuntu 24.04

Leont:
This upgrades the CI runners to Ubuntu 24.04, and the backwards compatibility runner to 22.04


Leont:
I should have double-checked before writing this, this is exactly the same change as #1942.

PR #1955, opened 2025-09-25
Zero top of register when getting uint32 class member

MasterDuke17:
I incorrectly addressed nine++'s comment (https://github.com/MoarVM/MoarVM/pull/1746#discussion_r1134483289) by accidentally just copying what was in the `else` without adjusting it down to 32 bit operations.

Fixes https://github.com/rakudo/rakudo/issues/5954.


lizmat:
As I'm no longer on Intel hardware, I cannot test this, but it looks good to me.

PR #1967, opened 2025-10-25
Add libffi as a 3rdparty submodule

MasterDuke17:
Also remove dyncall, so the only options now are building libffi or using the system libffi.

This work on my ARM Linux laptop, but I've not yet tried to follow the directions here (https://github.com/libffi/libffi#installing-libffi) for Windows.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -29,7 +29,7 @@
 	url = https://github.com/Nicoshev/rapidhash
 [submodule "3rdparty/libffi"]
 	path = 3rdparty/libffi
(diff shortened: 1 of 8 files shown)


MasterDuke17:
> Can we add a layer of indirection here? Instead of adding a submodule that points at upstream, can we create a libffi repo for a fork and use that?
> 
> If we need to make patches to libffi, if is more easily managed in a MoarVM/libffi repo (imo).

I don't really have any objection, but if we can get upstream to make the changes that seems even easier.

PR #1982, opened 2025-12-24
Output JITDUMP format with MVM_JIT_PERF_DUMP

timo:
It's a bit of a hassle, but together with `perf inject` you can now get samples along with the correct assembly code shown in `perf report`.

Highly recommend building with `-fno-omit-frame-pointer` in the CFLAGS as well as LDFLAGS of moarvm and rakudo.

```
rm perf.data* perf.jit.data* jit-*.dump jitted-*.so
env MVM_JIT_PERF_DUMP=1 perf record -g -k 1 rakudo example_code.raku
perf inject --jit --input perf.data --output perf.jit.data
mv perf.jit.data perf.data
perf report
```

With this feature, `perf report` will allow you to open the "annotate" view for jitted code frames:

image
image

The jitdump format also allows adding line numbers to memory locations inside of the jitted code. I will have to experiment with that a little bit.

https://raw.githubusercontent.com/torvalds/linux/refs/heads/master/tools/perf/Documentation/jitdump-specification.txt

This article helped me get started:

https://theunixzoo.co.uk/blog/2025-09-14-linux-perf-jit.html

### DIFF
--- a/src/core/instance.h
+++ b/src/core/instance.h
@@ -353,6 +353,8 @@ struct MVMInstance {
     MVMint32 jit_expr_last_bb;
     /* File for JIT perf map logging */
     FILE *jit_perf_map;
+    /* File for JIT perf jitdump output to be used with perf-inject */
+    FILE *jit_perf_jitdump;
 
     /* Directory name for JIT bytecode dumps */
     char *jit_bytecode_dir;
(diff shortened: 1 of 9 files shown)


timo:
https://share.firefox.dev/4vU8PP2 using https://github.com/mstange/samply `samply record` with `MVM_JIT_PERF_DUMP` set to a path to an existing directory gives you jitted frames sampled with names and assembly with no extra work, which is quite nice. you can share your recording using share.firefox.dev for others to look at, too.

PR #1986, opened 2026-01-10
Switch from Ryu to Żmij for our double-to-ascii implementation

MasterDuke17:
It's faster, and the output is formatted more like what we require, so it needs less modifying afterwards.

`MVM_SPESH_BLOCKING=1 time taskset -c 4,5 nqp -e 'my int $s := 0; my num $n := nqp::time; my int $i := 0; my str $ns; while $i++ < 10_000_000 { $ns := $n; $s := $s + nqp::chars($ns) }; say($s); say($n)'` drops from ~0.50s to ~0.40s.


MasterDuke17:
> However, I also realized it needs `-std=gnu11` to build, but we only do `-std=gnu99`.

I have a PR with zmij to fix the current build problem. I recently realized that we use `c17` for Windows (https://github.com/MoarVM/MoarVM/blob/main/build/setup.pm#L422), so I'm not terribly concerned about bumping non-Windows to `c11`.

MasterDuke17:
It's now using the v1.1 release of Żmij. Will merge this after the release unless there are any objections.

PR #1996, opened 2026-03-28
Support for fuzzing moar itself and nqp/raku code with AFL

timo:
There are two more-or-less orthogonal sets of features in this PR:

* Building specific binaries for small parts of MoarVM that we can fuzz with feedback from the C code itself.
* Building a nqp-m or rakudo-m (later) that can run like regular but also take inputs from a fuzzer harness (AFL in our case) and provide coverage feedback to the harness.

There is also the future possibility to send combined coverage feedback from C instrumentation (which `afl-cc` spits out) and the coverage feedback from running instrumented moar bytecode (controllable through an env var at the moment).

The most annoying part of this work is to get the build system to cooperate:

* Any C files that coverage should be provided for need to be compiled with LLVM's SanCov (there is also support for GCC, but it's kind-of deprecated?) and `afl-cc` sets this up for us.
* In order to maximize effectiveness of a fuzzing campaign, at least two variants of all `.c` files need to be compiled:
  * Once with just the regular instrumentation
  * Once with `cmplog` instrumentation to facilitate the `input2state` and `redqueen` techniques (`-c` flag in `afl-fuzz`)
  * Optionally: At least one further binary with ASan and maybe other sanitizers to be used with `afl-fuzz`'s `-w` flag
* For fuzzing nqp/raku code, it's probably a good idea to leave out almost all the C-level coverage feedback?
  * Right now I'm just compiling all the C files with `afl-cc` just so I get `__AFL_COMPILER` set and the `__AFL_INIT`, `__AFL_FUZZ_INIT`, and `__AFL_LOOP` macros, and use an `AFL_LLVM_ALLOWLIST` that matches nothing. This is pretty bad and should go away soon.
  * There is probably some files that would be interesting to include coverage for, to be determined later.
  * Here we don't need to compile different binaries to get `cmplog`, it can be switched on and off with an env var.
  * It would be cool to automatically build a variant with `GC_DEBUG=1` or even `GC_DEBUG=2`

Right now the Makefile is shoddily hacked together to create multiple variants of all .c files if another rule requests them, and then there's make variables that change all entries to have the right suffix with the `addsuffix` GNU makefile function ...

This is obviously not fantastic.

There is one specific C-level fuzzing target that can exercise the streaming decoder. You can ask the Makefile to build the binaries for this target with `make fuzzing` and its sibling targets `fuzzing-extras`, `fuzzing-complete`.

For the data that `MVM_BB_COVERAGE_DUMP_FIRST_EDGE_HIT`, `MVM_BB_COVERAGE_DUMP_BB_IDS`, and 
`MVM_BB_COVERAGE_DUMP_BB_LINENOS` spit out, there's `tools/parse_edge_coverage.raku` which collects hit edges per frame including frames going between BBs of different frames.

image

<details><summary>Text from the screenshot</summary>
<pre>

$NQPLIB/NQPHLL.moarvm[32]      "termish"

      1        <- $NQPLIB/NQPCORE.setting.moarvm[166] BB 89
      1        <- $NQPLIB/NQPHLL.moarvm[45]    "nulltermish" BB 9
      1        <- $NQPLIB/NQPHLL.moarvm[92]    "EXPR" BB 11
      1        -> $NQPLIB/NQPCORE.setting.moarvm[160] BB 1
      1        -> $NQPLIB/QRegex.moarvm[82]    "!cursor_start" BB 1
      2        <- $NQPLIB/QRegex.moarvm[82]    "!cursor_start" BB 35
      2 ->   3
      3 ->   4
      4 ->  12
     12 ->  13
     13 ->  15
     15 ->  16
     16        -> $NQPLIB/NQPCORE.setting.moarvm[160] BB 1
     16        -> $NQPLIB/NQPHLL.moarvm[41]    "prefixish" BB 1
     17        <- $NQPLIB/NQPHLL.moarvm[41]    "prefixish" BB 25
     17        <- $NQPLIB/QRegex.moarvm[91]    "!cursor_fail" BB 6
     17 ->  21
     17 ->  86
     21        -> $NQPLIB/NQPCORE.setting.moarvm[160] BB 1
     21        -> $NQPLIB/QRegex.moarvm[86]    "!cursor_capture" BB 1
     22        <- $NQPLIB/QRegex.moarvm[86]    "!cursor_capture" BB 12
     22 ->  23
     23 ->  24
     24 ->  25
     24 ->  26
     26 ->  28
     29 ->  30
     30 ->  31
     30 ->  32

</pre>
</details> 

To-Do and Wishlist:

* Reduce amount of `bb_entered` instructions generated in instrumentation
* Make instrumented versions of functions dumpable with `--dump` or similar (also interesting for tests!)
  * Something like this is pretty much required if a developer wants to use edge coverage to find out what BB indices correspond to in the actual code
* Allowlist / Denylist for frames to be instrumented
  * Maybe a few presets like for NQPClassHow's cache, NFA optimizer, and similar things
* Automatically turn coverage reporting off and on when entering/leaving dispatcher code
* Implementation of the new ops in the JIT
* IJON features for nqp/raku code

Research topics:

* How much is edges between different frames worth, either for feedback to the fuzzer, or for users?
  * I think they are pretty cool to figure out how execution flows.
  * If we only handle edges inside of frames, we can enumerate all of them ahead of time and have guaranteed conflict-free IDs for edges (this is also what you get with afl's PCGUARD and LTO instrumentation modes)
* Expand from edge coverage to path coverage, or at least n-gram edge coverage?
  * Shouldn't be too hard until you want guaranteed conflict-free coverage bits for paths ...
* Can we give bits to "exception with text X was thrown inside BB Y"? Is that helpful?
* For fuzzing recursive parsers, can we get "partial" context-aware coverage?
  * Tokens like `ws` appear basically everywhere, so treating the insides of `ws` as distinct for every spot it's included in might be good to help the fuzzer figure out putting whitespace after otherwise correctly matched tokens (easier to find `class` + whitespace instead of `classsssssss`).

Plenty of ideas for future work:

* AFL can be told with allowlist / denylist what functions should be instrumented.
  * Stuff we don't care as much about in terms of code path coverage can be turned off. For example, some parts of the garbage collector, or the insides of our hashes, might be good candidates for exclusion
* The "IJON" APIs allow for giving AFL feedback on interesting states a program can be in that aren't visible from just execution path coverage, and also to direct AFL to try to maximize or minimize supplied values
* More parts of moarvm's insides should get fuzzing targets made for them to exercise all their quirks.
* It would probably be good to include a corpus for existing fuzz targets to get started with looking for new stuff quickly after changes in the future.


coke:
build failures are pointing to a missing cmp library.

src/instrument/bb_edge_coverage.c
```diff
@@ -0,0 +1,1156 @@
...
+
+            /* Now go through instructions to see if any are annotated with a
+            * precise filename/lineno as well. */
+            while (ins) {
+                MVMSpeshAnn *ann = ins->annotations;
+
+                while (ann) {
+                    if (ann->type == MVM_SPESH_ANN_LINENO
+                            && ann->data.lineno.filename_string_index != last_filename
+                            && ann->data.lineno.line_number != last_line_number) {
```
gdonald, line 312: Should this second `&&` be `||`?

PR #1997, opened 2026-04-13
A bunch of debugserver bugfixes & any-break support

patrickbkr:


src/debug/debugserver.c
```diff
@@ -2378,9 +2402,50 @@ static MVMint32 request_object_attributes(MVMThreadContext *dtc, cmp_ctx_t *ctx,
...
+                                            /* For 64-bit 2's complement numbers the positive max is 2**63-1, which is 63 bits,
+                                             * but the negative max is -(2**63), which is 64 bits. */
+                                            if ((MP_NEG == i->sign && bits > 64) || (MP_NEG != i->sign && bits > 63)) {
+                                                cmp_write_conststr(ctx, "bigint");
+                                                cmp_write_conststr(ctx, "value");
+                                                mp_err err;
+                                                int len;
+                                                char *buf;
+                                                if ((err = mp_radix_size(i, 10, &len)) != MP_OKAY) {
+                                                    MVM_exception_throw_adhoc(dtc, "Error calculating the size of a big integer: %s", mp_error_to_string(err));
```
timo: I wrote this on IRC but to make sure it doesn't get lost here it is on the PR again:

I don't think we can throw exceptions inside the debugserver, so these exception throws should instead become a logged error (when debugspam is on) and putting some kind of signal in the msgpack data that tells the client that the value couldn't be sent for whatever reason.

We also can't reliably "un-send" parts of a message, so in order for the client to be able to keep the connection we need to send the remainder of the expected message, too.

I think ideally, the remaining data we're going through, attributes in this case, should be sent as if nothing had gone wrong, so that one bugged attribute in an object doesn't prevent reading the rest of the attributes

timo (changes requested):
This already looks very good and I'm looking forward to landing it in main. I'm requesting one change related to exceptions on the debugserver thread.

src/debug/debugserver.c
patrickbkr: I've changed it to just push the error inline to the client.

timo (approved):
looks good now.

i'd be okay with merging it into main like this, though maybe we can talk about signaling the errors in bigint sending a different way than putting a string in the "value" slot :)

patrickbkr:
> ... maybe we can talk about signaling the errors in bigint sending a different way than putting a string in the "value" slot :)

Since we have a month's time to fix up the error handling, I've merged it now.

PR #1998, opened 2026-04-17
Clear gen2 default alloc flag when reporting erroneous bytecode

timo:
If an exception is thrown by any of the error cases when unpacking bytecode, we jump out of the function to run a handler which will not be prepared to deal with the gen2 flag being set (for example by having sp_fastcreate in it)

Found with afl++

I was fuzzing nqp's stage parse, AFL decided to fuzz its way into compiling and running at parse time via the `constant` keyword, managed to write code where a named argument came before a positional argument in a callsite that was serialized into the bytecode file, which then threw an exception in bytecode unpacking, triggering this issue.


MasterDuke17 (approved):

lizmat:
Cool!

PR #2006, opened 2026-05-20
Correctly populate dest array in nqp::strtocodes

MasterDuke17:
The array populated by nqp::strtocodes (i.e.,
MVM_unicode_string_to_codepoints), wasn't getting its `ssize` set. Some uses of it would be ok, but attempting to resize it could go bady.

Fixes `use nqp; nqp::strtocodes("3.9999999999", 1, my int32 @a); dd @a; nqp::splice(@a, nqp::list, 1, 1); dd @a`. It used to segfault, now it prints:
```
int32  = array[int32].new(51, 46, 57, 57, 57, 57, 57, 57, 57, 57, 57, 57)
int32  = array[int32].new(51, 57, 57, 57, 57, 57, 57, 57, 57, 57, 57)
```

### DIFF
--- a/src/strings/normalize.c
+++ b/src/strings/normalize.c
@@ -200,6 +200,7 @@ void MVM_unicode_string_to_codepoints(MVMThreadContext *tc, MVMString *s, MVMNor
     ((MVMArray *)out)->body.slots.u32 = (MVMuint32 *)result;
     ((MVMArray *)out)->body.start     = 0;
     ((MVMArray *)out)->body.elems     = result_pos;
+    ((MVMArray *)out)->body.ssize     = result_alloc;
 }
 
 /* Initialize the MVMNormalizer pointed to to perform the specified kind of


src/strings/normalize.c
```diff
@@ -200,6 +200,7 @@ void MVM_unicode_string_to_codepoints(MVMThreadContext *tc, MVMString *s, MVMNor
     ((MVMArray *)out)->body.slots.u32 = (MVMuint32 *)result;
     ((MVMArray *)out)->body.start     = 0;
     ((MVMArray *)out)->body.elems     = result_pos;
+    ((MVMArray *)out)->body.ssize     = result_pos;
```
niner: I think it should be result_alloc rather than result_pos as maybe_grow_result will increase the size in 32 element steps. Thus we may end up with an array that has some room to spare.

niner (changes requested):

niner (approved):

coke:
We are very close to the release this Saturday; is this critical to get in 2026.05? If so, please merge and bump nqp/rakudo ASAP.

MasterDuke17:
> is this critical to get in 2026.05?

I doubt `nqp::strtocodes()` is used in a lot of critical sections of Rakudo (but that's just an uninformed opinion), but then again the fix is so trivial I don't think getting it in is high risk. I'll merge this in a minute, but wait for a second opinion before bumping NQP+Rakudo.

coke:
... You've merged it. we HAVE to bump nqp+rakudo or we need to revert it now.

PR #2009, opened 2026-05-27
Fix smallint branch of MVM_bigint_mod for values that wrap

### LINKED ISSUE #6210: Modulo with some large positive integers give incorrect negative values
# The problem

If you modulo a large number by a larger number, at a certain point it will start giving negative values:

```
say 700_000_000 % 1_450_380_277
```
will result in `-694206742`

# Expected behaviour

It should just give the first number `700_000_000`.

# Version

```
Welcome to Rakudo v2026.05.
Implementing the Raku Programming Language v6.d.
Built on MoarVM version 2026.05.
```
using [this Dockerfile](https://github.com/code-golf/code-golf/blob/master/langs/raku/Dockerfile)

# Notes

There was a recent change at https://github.com/rakudo/rakudo/pull/6177 which also references the https://github.com/MoarVM/MoarVM/pull/1989 which could also be the issue. The MoarVM one only seems to be a change for smallints though...

I love all the chat about missing spectests there because I wouldn't have thought to have one for modulo by a larger number either.

MasterDuke17:
by promoting all intermediate calculations to 64-bit.

Fixes https://github.com/rakudo/rakudo/issues/6210.

### DIFF
--- a/docs/ChangeLog
+++ b/docs/ChangeLog
@@ -1,3 +1,6 @@
+New in 2026.06
++ Fix smallint branch of MVM_bigint_mod for values that wrap
+
 New in 2026.05
 + Output JITDUMP format with MVM_JIT_PERF_DUMP
 + debugserver improvements
--- a/src/math/bigintops.c
+++ b/src/math/bigintops.c
@@ -558,8 +558,10 @@ MVMObject * MVM_bigint_mod(MVMThreadContext *tc, MVMObject *result_type, MVMObje
             store_bigint_result(bc, ic);
             adjust_nursery(tc, bc);
         } else {
-            /* % in C technically gives the remainder, but we want the modulus. */
-            store_int64_result(tc, bc, ((ba->u.smallint.value % bb->u.smallint.value) + bb->u.smallint.value) % bb->u.smallint.value);
+            /* % in C technically gives the remainder, but we want the modulus. Force everything to 64-bit so intermediate calculations don't overflow/wrap. */
+            MVMint64 a = ba->u.smallint.value;
+            MVMint64 b = bb->u.smallint.value;
+            store_int64_result(tc, bc, ((a % b) + b) % b);
         }
     }
 


coke:
Please also update the Changelog, thanks.

lizmat:
Seems to cause breakage in t/04-nativecall/02-simple-args.t  ?

niner (approved):

dzaima:
The first `%` still needs should have some handling of the undefined behavior case (simple option if desiring to not change much be to widen its input too).

`say (-2147483647-1) % -1` crashes in a downloaded `Test.T.LinMVMreloc` build artifact at least.

MasterDuke17:
> Seems to cause breakage in t/04-nativecall/02-simple-args.t ?

For you locally? I don't see that locally or in CI. If you're talking about the CI clang failures those are longstanding that need https://github.com/MoarVM/MoarVM/pull/1967 to fix.

MasterDuke17:
> The first `%` still needs should have some handling of the undefined behavior case (simple option if desiring to not change much be to widen its input too).
> 
> `say (-2147483647-1) % -1` crashes in a downloaded `Test.T.LinMVMreloc` build artifact at least.

Interesting. I tried `say -2147483648 % -1` (copied from https://github.com/rakudo/rakudo/issues/6210#issuecomment-4548672406) and that worked fine (prints `0`) on my x86-64 linux destop and my apple silicon linux laptop, even under UBSAN. **However**, `say (-2147483647-1) % -1` does give `runtime error: division of -2147483648 by -1 cannot be represented in type 'int'` with UBSAN on the desktop and laptop when building with GCC, with clang I get a floating point exception, but UBSAN is quiet about it.

dzaima:
> with clang I get a floating point exception, but UBSAN is quiet about it.

Seems to be a bug of clang's `-fwrapv` incorrectly disabling the overflow check of division/modulo overflow. Reported that: https://github.com/llvm/llvm-project/issues/200152

PR #2010, opened 2026-05-28
fix gettid on older glibc's

patrickbkr:
Glibc < 2.30 does not define `gettid()`. The man page states:

    Glibc does not provide a wrapper for this system call; call it using
    syscall(2).

So we just define a respective `gettid()`.

### DIFF
--- a/src/jit/compile.c
+++ b/src/jit/compile.c
@@ -11,6 +11,8 @@ void MVM_jit_compile_expr_tree(MVMThreadContext *tc, MVMJitCompiler *compiler, M
 #if linux
 #include <unistd.h>
 #include <time.h>
+#include <sys/syscall.h>
+#define gettid() syscall(SYS_gettid)
 
 struct jitdump_jit_load_record {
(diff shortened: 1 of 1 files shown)


coke:
Can we get a Changelog update as part of the PR as well? Thanks.

ShimmerFairy:
It would be better, I think, if the macro was only defined when it was truly needed, and using the system's `gettid()` instead where available. Since we can seemingly judge its availability based on the glibc version, something like this should do fine:

```c
#if (__GLIBC__ < 2) || (__GLIBC__ == 2 && __GLIBC_MINOR__ < 30)
#include <sys/syscall.h>
#define gettid() syscall(SYS_gettid)
#endif
```

At least on my system, this prevents the macro from being defined. Do note that if `_GNU_SOURCE` isn't guaranteed to be on at this point, the test should also check for that, since it's required for `gettid()` to be available.

Edit: just found that `__GLIBC_PREREQ(2, 30)` would get the job done in a lot less typing, and be less prone to typos.

PR #2018, opened 2026-06-14
Add code coverage instrumentation filtering

gdonald:
When using `MVM_COVERAGE_LOG` to measure code coverage of a single module or app, MoarVM currently instruments _everything_ that loads, including the Raku standard library and NQP internals.

Add the new `MVM_COVERAGE_FILES` environment variable that filters code coverage instrumentation to source files whose path matches one of a set of comma-delimited substrings. This allows skipping the non-matching files at instrumentation time so they never reach the coverage log and incur no instrumentation overhead.

When `MVM_COVERAGE_FILES` is unset or empty, _everything_ gets instrumented, so this is opt-in and backward compatible with existing behavior.

Here is a demo script for a simple method:

```raku
#!/usr/bin/env raku

# Set MVM_COVERAGE_FILES to a substring and only files whose path contains it
# get coverage instrumentation.

my $workdir = $*TMPDIR.add('test-moar-coverage');
$workdir.mkdir;

my $target = $workdir.add('coverage_demo_target.raku');
$target.spurt: q:to/CODE/;
    sub greet($name) {
        say "hello, $name";
    }
    greet('coverage');
    CODE

# Distinct source files appearing in the coverage log
sub coverage-files(%extra --> List) {
    my $log = $workdir.add('cov.log');
    my $proc = run $*EXECUTABLE, $target.absolute, :out, :err,
                   env => %( %*ENV, MVM_COVERAGE_LOG => $log.absolute, |%extra );
    
    $proc.out.slurp(:close);
    $proc.err.slurp(:close);

    # Each log line is: HIT  <file>  <line>
    my @files = $log.lines.map({ m/^ 'HIT' \s+ (.+) \s+ \d+ $/ ?? ~$0 !! Empty }).unique.sort;
    $log.unlink;
    @files.List;
}

my @without = coverage-files(%());
my @with    = coverage-files(%( MVM_COVERAGE_FILES => 'coverage_demo_target' ));

say "without filter: { @without.elems }";
say "with filter: { @with.elems }";

$target.unlink;
$workdir.rmdir;

```

```shell
> ./moar_coverage.raku
without filter: 366
with filter: 1
```

Tested against latest source builds:

```shell
> raku --version
Welcome to Rakudo™ v2026.05-117-g9b162bcf7.
Implementing the Raku® Programming Language v6.d.
Built on MoarVM version 2026.05.1-12-g2c16915d5.
```

### DIFF
--- a/src/moar.c
+++ b/src/moar.c
@@ -469,25 +469,21 @@ MVMInstance * MVM_vm_create_instance(void) {
             char *coverage_files = getenv("MVM_COVERAGE_FILES");
 
             if (coverage_files && coverage_files[0]) {
-                /* Copy, so we can chop into substrings in place. */
-                char *buf = MVM_malloc(strlen(coverage_files) + 1);
                 char **filters;
                 char *p;
-                char *tok = buf;
+                char *tok = coverage_files;
                 MVMuint32 count = 1;
                 MVMuint32 i = 0;
                 int last;
 
-                strcpy(buf, coverage_files);
-
                 /* Count comma-delimited substrings. */
-                for (p = buf; *p; p++)
+                for (p = coverage_files; *p; p++)
                     if (*p == ',') count++;
 
                 filters = MVM_malloc(count * sizeof(char *));
 
                 /* Replace commas with \0 and store pointers to non-empty substrings. */
-                for (p = buf; ; p++) {
+                for (p = coverage_files; ; p++) {
                     if (*p == ',' || *p == '\0') {
                         last = (*p == '\0');
                         *p = '\0';
@@ -501,8 +497,7 @@ MVMInstance * MVM_vm_create_instance(void) {
                     }
                 }
 
-                /* Assign filters.
-                 * buf and filters intentionally kept for the lifetime of the VM. */
+                /* Assign filters. */
                 instance->coverage_file_filters = filters;
                 instance->coverage_file_filter_count = i;
             }
@@ -813,6 +808,9 @@ void MVM_vm_destroy_instance(MVMInstance *instance) {
     if (instance->jit_breakpoints) {
         MVM_VECTOR_DESTROY(instance->jit_breakpoints);
     }
+
+    /* Clean up coverage file filters */
+    MVM_free(instance->coverage_file_filters);
 
     /* Clean up cross-thread-write-logging mutex */
     uv_mutex_destroy(&instance->mutex_cross_thread_write_logging);
--- a/src/moar.c
+++ b/src/moar.c
@@ -463,6 +463,50 @@ MVMInstance * MVM_vm_create_instance(void) {
         else {
             instance->coverage_control = 0;
         }
+
+        /* Parse coverage file filters. */
+        {
+            char *coverage_files = getenv("MVM_COVERAGE_FILES");
+
+            if (coverage_files && coverage_files[0]) {
+                /* Copy, so we can chop into substrings in place. */
+                char *buf = MVM_malloc(strlen(coverage_files) + 1);
+                char **filters;
+                char *p;
+                char *tok = buf;
+                MVMuint32 count = 1;
+                MVMuint32 i = 0;
+                int last;
+
+                strcpy(buf, coverage_files);
+
+                /* Count comma-delimited substrings. */
+                for (p = buf; *p; p++)
+                    if (*p == ',') count++;
+
+                filters = MVM_malloc(count * sizeof(char *));
+
+                /* Replace commas with \0 and store pointers to non-empty substrings. */
+                for (p = buf; ; p++) {
+                    if (*p == ',' || *p == '\0') {
+                        last = (*p == '\0');
+                        *p = '\0';
+
+                        if (tok[0])
+                            filters[i++] = tok;
+
+                        if (last) break;
+
+                        tok = p + 1;
+                    }
+                }
+
+                /* Assign filters.
+                 * buf and filters intentionally kept for the lifetime of the VM. */
+                instance->coverage_file_filters = filters;
+                instance->coverage_file_filter_count = i;
+            }
+        }
     }
     else {
         instance->coverage_logging = 0;
--- a/docs/moar.pod
+++ b/docs/moar.pod
@@ -47,6 +47,31 @@ to an object it didn't allocate and doesn't have a lock for.
 Same as MVM_CROSS_THREAD_WRITE_LOG, except objects that are locked are included
 as well.
 
+=item MVM_COVERAGE_LOG
+
+If set, MoarVM instruments bytecode to log a C<HIT E<lt>fileE<gt> E<lt>lineE<gt>>
+entry to this file each time a new source line is entered. If the value is an
+empty string the log is written to stderr. The file is opened in append mode.
+If the path contains C<%d> it is replaced with the process ID.
+
+By default each (file, line) pair is logged once per static frame ("de-duped").
+See C<MVM_COVERAGE_CONTROL> to disable.
+
+=item MVM_COVERAGE_CONTROL
+
+Controls de-duping behaviour of C<MVM_COVERAGE_LOG>. If set to C<1>, the
+non-de-duping mode is enabled at runtime when C<nqp::coveragecontrol(1)> is
+called. If set to C<2>, non-de-duping mode is enabled from the start of the
+program. Has no effect unless C<MVM_COVERAGE_LOG> is also set.
+
+=item MVM_COVERAGE_FILES
+
+A comma-separated list of substrings. When set, only source files whose path
+contains at least one of the substrings are instrumented for coverage. Other files
+(such as the Raku standard library and NQP internals) are left uninstrumented and
+incur no coverage overhead. Unset or empty means instrument everything
+(the default). Has no effect unless C<MVM_COVERAGE_LOG> is also set.
+
 =back
 
 =head1 REPORTING BUGS
--- a/src/core/instance.h
+++ b/src/core/instance.h
@@ -552,6 +552,10 @@ struct MVMInstance {
     FILE *coverage_log_fh;
     MVMuint32  coverage_control;
 
+    /* Filters for coverage instrumentation. */
+    char     **coverage_file_filters;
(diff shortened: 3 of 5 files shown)


lizmat:
Nice, this should also help with [Test::Coverage](https://raku.land/zef:lizmat/Test::Coverage)

src/moar.c
```diff
@@ -463,6 +463,50 @@ MVMInstance * MVM_vm_create_instance(void) {
...
+                            filters[i++] = tok;
+
+                        if (last) break;
+
+                        tok = p + 1;
+                    }
+                }
+
+                /* Assign filters.
+                 * buf and filters intentionally kept for the lifetime of the VM. */
```
MasterDuke17: I don't like this. They should be freed by `--full-cleanup`.

src/moar.c
```diff
@@ -463,6 +463,50 @@ MVMInstance * MVM_vm_create_instance(void) {
         else {
             instance->coverage_control = 0;
         }
+
+        /* Parse coverage file filters. */
+        {
+            char *coverage_files = getenv("MVM_COVERAGE_FILES");
+
+            if (coverage_files && coverage_files[0]) {
+                /* Copy, so we can chop into substrings in place. */
```
MasterDuke17: I know it's not great form to modify through the pointer returned by `getenv()`, but wouldn't it be safe in this case? Then all the copying wouldn't be needed.

MasterDuke17:
Overall question, would this mostly be doable via `nqp::coveragecontrol()` without adding anything new?

timo:
In my fuzzing branch I've implemented a filter for skipping entire CUs by filename or individual functions by the name of the file they are annotated to come from. I wrote it already a little while ago, but it was hidden deep within my git stashes.

I think the code in that branch that introduces coverage based on edges between BBs instead of lines should be picked apart a little bit more so it can be re-used for line coverage reports as well, then both could directly share the filter implementation.

The filter I'm referring to is only for skipping, though. What you have here is an allowlist rather than a denylist like I have. Ideally, I think we should have both (though I'm not sure if moar should error out or ignore one of the two when both are specified).

I would like to hear what you think of how exactly I have implemented it in the other branch.

Also, I think it's high time for the coverage log output format to be switched over to something more efficient and something that doesn't break if someone does something silly like putting a newline or two spaces into a filename. In that branch over there I'm outputting messagepack, since that's already used for the debug server protocol. There's also a raku script in the tools/ folder that reads the output of the bb edge coverage and generates a coverage report HTML which can be used as an example for how to read a format like that.

Here's the link to the commit where I finally committed the skip list environment variable support, but the branch is not yet in a state that's ready to merge to main.

Here:

https://github.com/MoarVM/MoarVM/commit/9360b0ae9f845121d522e7d715127dda6509b9c9#diff-264c06a4d51272d318979e63865aad42fcb8835f55a13ef46098d19f68c74670R124-R141

and here:

https://github.com/MoarVM/MoarVM/commit/9360b0ae9f845121d522e7d715127dda6509b9c9#diff-264c06a4d51272d318979e63865aad42fcb8835f55a13ef46098d19f68c74670R333-R362

timo:
@MasterDuke17 
> Overall question, would this mostly be doable via `nqp::coveragecontrol()` without adding anything new?

nqp::coveragecontrol is currently an nqp op that just takes a single integer, so that's not good enough to do something complex like a filter list. If we switch it over to a syscall, we have a lot of freedom in terms of what the arguments can look like.

On the other hand, I think putting something like the allow/denylist for instrumentation into user code would run it too late for most purposes. In other words, before you can set up a list of files you want to have included in the coverage report, you'd already have a few thousand lines from startup and probably also compilation?

Alternatively, coverage reporting can be turned off until an explicit coveragecontrol-equivalent call is done to install filters. Then you'd have to set an env var to opt in to "no coverage tracking or logging at all until it's manually configured" behaviour, pretty much like MVM_COVERAGE_CONTROL currently works. Maybe that's what you meant all along?

I think for now I'm in favor of env vars for filters, though.

src/moar.c
gdonald: I added `MVM_free(instance->coverage_file_filters);` in `MVM_vm_destroy_instance`.

src/moar.c
gdonald: I didn't think it would be great form either, but yes it would be safe in this case. I have dropped the copying code.

gdonald:
> Overall question, would this mostly be doable via `nqp::coveragecontrol()` without adding anything new?

`nqp::coveragecontrol()` is an on/off switch and duplicate enabling/prevention mechanism.  It will recognize these possible values via `MVM_COVERAGE_CONTROL`:

0 = log each line once
1 = log nothing
2 = log every single hit

None of these options helps to filter out coverage reporting for Raku standard library and NQP internals.

gdonald:
> Nice, this should also help with [Test::Coverage](https://raku.land/zef:lizmat/Test::Coverage)

For [BDD::Behave's](https://behave.dev/) own test suite:

    behave --coverage

currently takes 43 minutes.  While

    MVM_COVERAGE_FILES='BDD-Behave/lib' behave --coverage

take less than 4 minutes.  So I can't wait for this change to land :)

MasterDuke17:
> Alternatively, coverage reporting can be turned off until an explicit coveragecontrol-equivalent call is done to install filters. Then you'd have to set an env var to opt in to "no coverage tracking or logging at all until it's manually configured" behaviour, pretty much like MVM_COVERAGE_CONTROL currently works. Maybe that's what you meant all along?


> None of these options helps to filter out coverage reporting for Raku standard library and NQP internals.

Yeah, I was sort of thinking if you set `MVM_COVERAGE_CONTROL` and then turned coverage on/off in the user code with `nqp::coveragecontrol()` you'd sort of be filtering, but I guess you're really just removing "startup" code (i.e., anything before the user code start running), not all subsequent Rakudo/NQP code.

MasterDuke17 (approved):

gdonald:
> The filter I'm referring to is only for skipping, though. What you have here
> is an allowlist rather than a denylist like I have. Ideally, I think we
> should have both (though I'm not sure if moar should error out or ignore one
> of the two when both are specified).

For file scope, having both makes sense to me. I'd model it so allowlist picks up what's in scope, then the denylist carves out exceptions. On a tie, don't error out, deny wins.

For more granular exclusions I prefer wrapping the scope I want to ignore directly in the implementation code with `:nocov:` or similar.

> I would like to hear what you think of how exactly I have implemented it in
> the other branch.

It looks like `MVM_COVERAGE_SKIP_FILE_FILTER` and `MVM_COVERAGE_SKIP_CU_FILTER` are parsed again and again per instrumented frame. How about one-time parsing, and raise their stature a bit?

```
struct {
    char **patterns;
    MVMuint32 count;
} cu_skip, file_skip;
```

> Also, I think it's high time for the coverage log output format to be
> switched over to something more efficient and something that doesn't break if
> someone does something silly like putting a newline or two spaces into a
> filename. In that branch over there I'm outputting messagepack, since that's
> already used for the debug server protocol.

Using the same format again sounds like a good idea.

I'd also want to fix the process (or the human) that is being silly with newlines and double space character filenames :P

> There's also a raku script in the
> tools/ folder that reads the output of the bb edge coverage and generates a
> coverage report HTML which can be used as an example for how to read a format
> like that.

I looked at `tools/parse_edge_coverage.raku` a bit.

How's the interval helping?  Isn't that going to involve some kind of non-local timer thread?

`my $inf-bits = supply { whenever Supply.interval(0.01) { emit $inf.read(10240); done if $inf.eof } };`

Why not just?

`my $inf-bits = supply { emit $inf.read(10240) until $inf.eof; done }`

Otherwise, looks pretty normal.

timo:
> How's the interval helping? Isn't that going to involve some kind of non-local timer thread?
> 
> `my $inf-bits = supply { whenever Supply.interval(0.01) { emit $inf.read(10240); done if $inf.eof } };`
> 
> Why not just?
> 
> `my $inf-bits = supply { emit $inf.read(10240) until $inf.eof; done }`

you're right. by just putting it as `supply { emit ...; done }` there's no parallelism between reading the file and processing the data. in other words, the next statement of the outside code happens after the `done` happens inside the supply (not on the exact line where `$inf-bits` is declared, since the body of a `supply` runs whenever something subscribes to it). (Though I haven't tested your exact suggestion yet, maybe it just happens to work as you hope?)

The `interval` is probably not the most elegant way to do it. I want to try something else, but not exactly sure what else.

timo:
> It looks like `MVM_COVERAGE_SKIP_FILE_FILTER` and `MVM_COVERAGE_SKIP_CU_FILTER` are parsed again and again per instrumented frame. How about one-time parsing, and raise their stature a bit?
> 
> ```
> struct {
>     char **patterns;
>     MVMuint32 count;
> } cu_skip, file_skip;
> ```

It might make more sense to create `VMString` instances and put them into an `MVMArray`, then we can use the substring searching ops. That will prevent trouble with multi-byte characters and multi-codepoint graphemes, that ... nobody is ever going to experience :) :) :)

gdonald:
Just checking in on this PR. Please let me know if any further code changes are required to get this approved and merged. Thanks.

MasterDuke17:
> Just checking in on this PR. Please let me know if any further code changes are required to get this approved and merged. Thanks.

I'd say go ahead and merge it *after* the release planned for this weekend.

coke:
Was there any followup on @timo's comment on Jun 22?

gdonald:
> > ```
> > struct {
> >     char **patterns;
> >     MVMuint32 count;
> > } cu_skip, file_skip;
> > ```
> 
> It might make more sense to create `VMString` instances and put them into an `MVMArray`, then we can use the substring searching ops. That will prevent trouble with multi-byte characters and multi-codepoint graphemes, that ... nobody is ever going to experience :) :) :)

That makes sense to me.  I will use a `MVMArray` of `VMString` instances when I work on this in the future.  Thank you for the guidance.

gdonald:
> Was there any followup on @timo's comment on Jun 22?

I followed up. Sorry that you had to prod me :)

gdonald:
I'm checking in on this PR again. Please let me know if there's anything more required on my end to get this approved and merged. Thanks.

PR #2021, opened 2026-06-30
utf8: Do not lock up when very long string has invalid utf8 in it.

timo:
Marking the thread blocked when doing the work made it possible for other threads to do GC while it's working, but if the thread tries to allocate an exception object to throw while it's marked blocked, it will enter a gc run that no other thread ever start.


lizmat (approved):

coke:
Please also add a Changelog entry.

MasterDuke17 (approved):

PR #2025, opened 2026-07-22
general GDB plugin improvements + prototype of `moar rrdb` command

timo:
* Robustness improvements for string pretty printer
* `moar bt`: Robust command to get a backtrace that also works in postmortem mode
* `moar tc`: Robust command to get the current thread's `MVMThreadContext *`
* `moar frame`, `moar frame {up,out}` Commands to look at individual frames and their arguments, and move to a given frame's caller or outer. [Example output of `moar frame` and `moar bt`](https://gist.github.com/timo/20b8e7f029b21581997f151be3bdce7f).

When running an `rr replay`, there's some extra functionality available:

* `moar rrdb` runs the recording and stores information about interesting / important events in an sqlite3 database
  * also adding `trackgc` records before and after addresses of objects moved during GC runs
  * also adding `stack` will store a stack trace at every rr event (roughly every time a syscall or signal happens, or the scheduler switched threads)
  * also adding `deopt` records deopt events
  * Not much in the plugin actually uses the recorded data by itself, but querying the database by hand with the `sqlite3` CLI is possible.
* `moar timeline` shows the current location in the replay in the context of GC runs and a sampling of calls to and returns from routines, with the tick numbers you can use to jump to the exact moment. [Here's some example output with the amount of events to display cranked up](https://gist.githubusercontent.com/timo/423eac9a06f26120ee8331e720ee16f6/raw/c0a368b5a7e9cb0ec12388df606529d2d776add1/gistfile1.txt).

Lots of "next steps" that can make especially stuff when running in rr a lot better are on my mind, but for now I really want to get all the improvements of the script in general into a release.

### DIFF
--- a/docs/ChangeLog
+++ b/docs/ChangeLog
@@ -1,6 +1,7 @@
 New in 2026.07
 
 + Bump mimalloc to v2.4.1
++ Many improvements and additions to the gdb python plugin
 
 New in 2026.06.1
 
--- a/src/6model/reprs/P6opaque.c
+++ b/src/6model/reprs/P6opaque.c
@@ -2427,7 +2427,7 @@ static void dump_p6opaque(MVMThreadContext *tc, MVMObject *obj, int nested) {
                                 else
                                     fprintf(stderr, "<already seen>");
                             }
-                            if (value != NULL && REPR(value)->ID == MVM_REPR_ID_MVMCode) {
+                            else if (value != NULL && REPR(value)->ID == MVM_REPR_ID_MVMCode) {
                                 MVMCode *code = (MVMCode*)value;
                                 MVMStaticFrame *sf = code->body.sf;
                                 fprintf(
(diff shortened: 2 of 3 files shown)


timo:
Since the gdb plugin is fully optional, and problems with it are unlikely to actually cause issues, and the only other change in this PR is in a debug helper function you would have to call by hand, I have decided that merging this without a full review just before the release is not an issue.

coke:
> Since the gdb plugin is fully optional, and problems with it are unlikely to actually cause issues, and the only other change in this PR is in a debug helper function you would have to call by hand, I have decided that merging this without a full review just before the release is not an issue.

For PR Historians: The release manager should still be consulted when we are this close to the release date.

coke:
~~Note: 3 of the checks failed (and we just went through and updated things so all are expected to pass) - might just need to be rerun.~~

PR #2031, opened 2026-08-05
Increase GC nursery size by 4x

MasterDuke17:
The value hasn't been updated since 2014. This provides a small but noticeable time decrease when compiling Rakudo's CORE.c.setting, with only a small increase in maxresident (as reported by /usr/bin/time).


timo:
what kinds of hardware did you measure this with?

MasterDuke17:
My Zen 2 desktop (32gb ram) and my M2 MBA (24gb ram).Sent from my iPhoneOn Aug 6, 2026, at 8:10 AM, timo ***@***.***> wrote:﻿timo left a comment (MoarVM/MoarVM#2031)
what kinds of hardware did you measure this with?

Issue #2037, opened 2026-09-01
Spesh issue - with guard causes spesh to run method belonging to wrong Role-composed class


m-doughty:
## Summary

In a role, guarding a private method call with `with` seems to get mis-specialized. 

After one composing class (Class A) is warmed, a different composing class (Class B) calling the same method runs Class A's private call, crashing with a wrong attribute error.

Class A & Class B are unrelated siblings, both use the role but neither inherits from the other. `MVM_SPESH_DISABLE=1` makes the same program run as I would expect it to, at any iteration count.

## Minimal Reproduction

Tested on MacOS Tahoe 26.3.1 (2025.05/2026.07) & Windows 11 Pro (2026.07). I did have a look and couldn't find anything in 2026.08's changelog that would address this issue but have not tested 2026.08.

`main.raku`:

```raku
use MinRole;

class A does MinRole {
    method go {
        self.execute(input => 's');
    }
}

class B does MinRole {
    method go {
        self.execute(input => 's');
    }
}

my $*SEEN = (Nil, Nil);

A.new.go for ^100_000;

say "B: ", (try {
    B.new.go; $*SEEN.raku;
}) // $!.message;
```

`lib/MinRole.rakumod`:

```raku
unit role MinRole;

has $!id = 'P';

method !apply($x --> Nil) {
    $*SEEN = ($!id, self.^name);
    Nil
}

method execute(:$input --> Nil) {
    self!apply($input) with $input;
    Nil
}
```

## Observations

1. This has to be split across the two files. A single file version does not reproduce the issue.
2. The block version appears to work for `with` i.e. `with $input { self!apply($input) }` -- this is my current workaround. It is only using postfix `with` guard that I could replicate the issue.
3. `MVM_JIT_DISABLE=1`, `MVM_SPESH_INLINE_DISABLE=1`, `MVM_SPESH_OSR_DISABLE=1`, and `MVM_SPESH_PEA_DISABLE=1` all still crash.
4. Alternating A/B from the start never fails, one class has to warm the frame alone. Then the precomp is poisoned for process lifetime. This is intermittently broken in real code, deterministically broken in the repro above (this is what made it very hard to reproduce).

## Output

MacOS:

```
% MVM_SPESH_DISABLE=1 raku -Ilib main.raku
B: $("P", "B")

% raku -Ilib main.raku
B: P6opaque: no such attribute '$!id' on type A in a B when trying to get a value
```

Windows:

```
[...]

PR #2038, opened 2026-09-04
#2037 - Fix unsound getlexperinvtype_o constant folding

m-doughty:
## Resolves

#2037 

## Problem

Spesh folded `getlexperinvtype_o` into an unguarded constant whenever the first argument was an object with a known type.

When a frame produced by a role method on a role-composing class, with a postfix topic-setting modifier, was passed to the spesh, the *topic class* (i.e. `Str`) was arg 0, not the invocant class (`self` in the outer scope).

Therefore, Spesh saw all of the warmup calls with the same `Str`-typed arg 0, and specialized by inlining a constant of the `::?CLASS` (the copy of the role that it was in, call it `A`). 

When `B` called an identical method (`!apply`), it matched the specialization for `A` because its argument tuple was also `(Str)`, so the `private-method` dispatch resolved `!apply` in A's private method table. A's apply then reads `$!id` on class A, the object is a B, so the attribute access throws a Raku exception.

With inheritance instead of siblings, it does not even throw, it just runs the wrong copy.

## Solution

Gave the opt a deopt point so that the type facts guard it like a `getlex`. Replaced the fold with resolving the name statically so that the read is indexed instead of by name.

## Confirmation

Running the code from the initial report:

```
% export PATH=~/code/raku-2037-install/bin:$PATH
% unset DYLD_LIBRARY_PATH

% raku --version
Welcome to Rakudo™ v2026.08-136-gfdc7d63571.
Implementing the Raku® Programming Language v6.d.
Built on MoarVM version 2026.08-7-g00c8b043b.

% raku -Ilib main.raku
B: $("P", "B")
```

## Benchmark

I was aware that this may affect performance beyond the affected code shape, so set up some benchmarks to confirm the impact of the fix.

### Results (on MacBook Pro M4 Max)

**default** (iters=2,000,000, churn=100,000)

| shape | before p50 | p90 | p99 | after p50 | p90 | p99 | Δ p50 | Δ p90 | Δ p99 |
|---|---|---|---|---|---|---|---|---|---|
| private call, method body | 72.3 ms | 73.5 | 76.7 | 75.1 ms | 76.7 | 78.8 | +3.8% | +4.3% | +2.6% |
| private call, postfix for | 104.8 ms | 107.1 | 108.0 | 70.0 ms | 72.0 | 72.7 | -33.2% | -32.8% | -32.7% |
| generic param type check | 146.5 ms | 153.9 | 156.6 | 154.5 ms | 157.2 | 162.7 | +5.5% | +2.1% | +3.9% |
| metamodel mixin churn | 73.9 ms | 77.2 | 79.5 | 69.9 ms | 71.7 | 72.7 | -5.4% | -7.2% | -8.6% |

**MVM_SPESH_BLOCKING=1** (iters=2,000,000, churn=100,000)

| shape | before p50 | p90 | p99 | after p50 | p90 | p99 | Δ p50 | Δ p90 | Δ p99 |
|---|---|---|---|---|---|---|---|---|---|
| private call, method body | 73.3 ms | 76.8 | 77.6 | 78.0 ms | 79.2 | 80.3 | +6.5% | +3.2% | +3.5% |
| private call, postfix for | 112.4 ms | 113.3 | 115.4 | 72.1 ms | 73.8 | 75.0 | -35.9% | -34.9% | -35.0% |
| generic param type check | 148.7 ms | 154.5 | 162.2 | 156.6 ms | 158.5 | 171.3 | +5.4% | +2.6% | +5.6% |
| metamodel mixin churn | 74.1 ms | 75.9 | 76.9 | 72.1 ms | 73.1 | 74.6 | -2.7% | -3.7% | -3.0% |

**MVM_SPESH_INLINE_DISABLE=1** (iters=2,000,000, churn=100,000)

| shape | before p50 | p90 | p99 | after p50 | p90 | p99 | Δ p50 | Δ p90 | Δ p99 |
|---|---|---|---|---|---|---|---|---|---|
| private call, method body | 116.0 ms | 120.0 | 121.4 | 116.4 ms | 122.1 | 124.8 | +0.3% | +1.8% | +2.8% |
| private call, postfix for | 215.9 ms | 217.7 | 223.7 | 173.7 ms | 176.3 | 181.9 | -19.5% | -19.0% | -18.7% |
| generic param type check | 192.6 ms | 194.1 | 197.4 | 201.2 ms | 203.4 | 206.5 | +4.5% | +4.8% | +4.6% |
| metamodel mixin churn | 85.6 ms | 87.1 | 88.7 | 85.7 ms | 87.6 | 89.7 | +0.1% | +0.6% | +1.1% |

**MVM_SPESH_DISABLE=1, control** (iters=2,000,000, churn=100,000)

| shape | before p50 | p90 | p99 | after p50 | p90 | p99 | Δ p50 | Δ p90 | Δ p99 |
|---|---|---|---|---|---|---|---|---|---|
| private call, method body | 412.8 ms | 418.0 | 426.4 | 412.3 ms | 415.6 | 417.7 | -0.1% | -0.6% | -2.0% |
| private call, postfix for | 517.9 ms | 520.8 | 523.8 | 513.8 ms | 519.5 | 528.3 | -0.8% | -0.3% | +0.9% |
| generic param type check | 628.3 ms | 632.3 | 636.8 | 628.7 ms | 632.6 | 636.9 | +0.1% | +0.0% | +0.0% |
| metamodel mixin churn | 141.0 ms | 143.2 | 145.6 | 142.7 ms | 144.2 | 146.6 | +1.2% | +0.7% | +0.7% |

Notes: 

- the cost on the generic check is one indexed read and one type guard per call.
- the for postfix gain was an unintended side-effect: previously the fold didn't get hit at all when the topic wasn't an object.
- we are only doing 60 runs of 2,000,000 iterations each so p99 is the worst timed run (technically p98.something).
- I have not included the JIT tables as I am benchmarking on a Mac, they came out within a noise interval of the default. If someone wants to run on x64, the code is below, may need some tweaks for non-Mac environment.

### Code

lib/BenchRole.rakumod:

```raku
unit module BenchRole;

role Priv is export {
    has int $!n = 0;
    method !bump() { $!n++ }
    method loop-in-method(int $k) { my int $i = 0; while $i < $k { self!bump; $i++ }; $!n }
    method loop-in-thunk(int $k)  { self!bump for ^$k; $!n }
}

role Gen[::T] is export {
    method check(T $x) { $x }
}
```

bench.raku:

```raku
use BenchRole;

class C does Priv {}
class G does Gen[Int] {}

sub MAIN(Int :$reps = 60, Int :$iters = 2_000_000, Int :$churn = 100_000, Int :$warmup = 3) {
    my @shapes =
        'private call, method body'  => { C.new.loop-in-method($iters) },
        'private call, postfix for'  => { C.new.loop-in-thunk($iters) },
        'generic param type check'   => { my $g = G.new; my int $i = 0; my int $n = $iters; while $i < $n { $g.check(42); $i++ } },
        'metamodel mixin churn'      => { my int $i = 0; my int $n = $churn; while $i < $n { my $o = Any but role { method x { 1 } }; $o.x; $i++ } };
    for @shapes -> (:key($name), :value(&code)) {
        code() for ^$warmup;
        for ^$reps { my $t = now; code(); say "$name\t{now - $t}" }
    }
}
```

runbench.py:

```python
import subprocess, os, sys, math, collections, time
B = os.path.dirname(os.path.abspath(__file__))
STACKS = [('before', '/Users/matt/code/raku-2037-baseline'),
          ('after',  '/Users/matt/code/raku-2037-install')]
CONFIGS = [  # (label, env, iters, churn)
    ('default',                    {},                              2_000_000, 100_000),
    ('MVM_SPESH_BLOCKING=1',       {'MVM_SPESH_BLOCKING': '1'},     2_000_000, 100_000),
    ('MVM_SPESH_INLINE_DISABLE=1', {'MVM_SPESH_INLINE_DISABLE': '1'}, 2_000_000, 100_000),
    ('MVM_JIT_EXPR_DISABLE=1',     {'MVM_JIT_EXPR_DISABLE': '1'},   2_000_000, 100_000),
    ('MVM_JIT_DISABLE=1',          {'MVM_JIT_DISABLE': '1'},          2_000_000,  100_000),
    ('MVM_SPESH_DISABLE=1 (control)', {'MVM_SPESH_DISABLE': '1'},     2_000_000,  100_000),
]
REPS = 60
def pct(xs, p):
    xs = sorted(xs); k = max(0, min(len(xs) - 1, math.ceil(p / 100 * len(xs)) - 1)); return xs[k]
base_env = dict(os.environ); base_env.pop('DYLD_LIBRARY_PATH', None); base_env.pop('DYLD_FALLBACK_LIBRARY_PATH', None)
raw = open(os.path.join(B, 'raw.tsv'), 'w')
out = []
for name, prefix in STACKS:
    v = subprocess.run([prefix + '/bin/raku', '--version'], capture_output=True, text=True, env=base_env).stdout.strip().split('\n')
    out.append(f"{name}: {v[0]} | {v[-1]}")
out.append(f"reps per shape: {REPS} timed after 3 warm-up runs; p99 at this sample size is the worst timed run")
results = {}
for label, env, iters, churn in CONFIGS:
    for name, prefix in STACKS:
        e = dict(base_env); e.update(env)
        t0 = time.time()
        p = subprocess.run([prefix + '/bin/raku', '-Ilib', 'bench2.raku', f'--reps={REPS}', f'--iters={iters}', f'--churn={churn}'],
                           cwd=B, capture_output=True, text=True, env=e)
        if p.returncode != 0:
            out.append(f"FAILED {label} {name}: {p.stderr[-300:]}"); continue
        d = collections.defaultdict(list)
        for line in p.stdout.splitlines():
            shape, secs = line.split('\t'); d[shape].append(float(secs)); raw.write(f"{label}\t{name}\t{shape}\t{secs}\n")
        results[(label, name)] = d
        sys.stderr.write(f"{label} {name} done in {time.time()-t0:.0f}s\n"); sys.stderr.flush()
    out.append(f"\n### {label}  (iters={iters:,}, churn={churn:,})\n")
    out.append("| shape | before p50 | p90 | p99 | after p50 | p90 | p99 | Δ p50 | Δ p90 | Δ p99 |")
    out.append("|---|---|---|---|---|---|---|---|---|---|")
    b = results.get((label, 'before'), {}); a = results.get((label, 'after'), {})
    for shape in b:
        if shape not in a: continue
        bp = [pct(b[shape], q) * 1000 for q in (50, 90, 99)]; ap = [pct(a[shape], q) * 1000 for q in (50, 90, 99)]
        dl = [(ap[i] - bp[i]) / bp[i] * 100 for i in range(3)]
        out.append(f"| {shape} | {bp[0]:.1f} ms | {bp[1]:.1f} | {bp[2]:.1f} | {ap[0]:.1f} ms | {ap[1]:.1f} | {ap[2]:.1f} | {dl[0]:+.1f}% | {dl[1]:+.1f}% | {dl[2]:+.1f}% |")
raw.close()
open(os.path.join(B, 'results.md'), 'w').write('\n'.join(out) + '\n')
print('\n'.join(out))
```

## Disclosure

LLMs were used in developing this fix:

* Claude Fable 5/5.1 - Investigation, confirmation & benchmarking
* DeepSeek V4 Flash (local) - Codebase navigation & comprehension assistance

The fix itself (all code in the commit) was written and validated by hand.
The benchmark code itself was written by Claude Fable 5.1.


lizmat:
Changed into a WIP after seeing discussions on https://irclogs.raku.org/moarvm/2026-09-05.html and https://irclogs.raku.org/moarvm/2026-09-06.html

timo:
I thought about this some more and I'm pretty sure the right fix is for rakudo to not emit a `getlexperinvtype` when the invocant type of a block has nothing to do with the value of the lexicals, such as when there's a `with $foo` or `given $foo` or similar block. rakudo has all the necessary knowledge at compile time to do this correctly, whereas moarvm has none. Pessimizing every `getlexperinvtype` just because rakudo in some cases emits it when the prerequisites for it aren't met is a bad idea in the first place.

Issue #2039, opened 2026-09-15
Breakpoint instrumentation rewrites bytecode but not annotations table, causing reported line numbers to drift


kliklamala:
### Summary

When debugging with `--debug-port`, reported line numbers in `Breakpoint Notification` (17), `Step Completed` (23), and `Thread Stack Trace Response` (14) drift forward away from the line actually executing. The drift accumulates progressively the longer a frame runs or the more statements precede the current position.

### Environment

- **MoarVM / Rakudo**: 2026.07 (and current `main`)
- **OS**: Linux x86_64
- **Debug protocol**: 1.5

### Root Cause

When starting with `--debug-port`, every static frame gets rewritten the first time it runs via `MVM_breakpoint_instrument` -> `line_numbers_instrument` -> `add_instrumentation` (`src/instrument/line_coverage.c`):

```c
static void add_instrumentation(MVMThreadContext *tc, MVMStaticFrame *sf, MVMuint8 want_coverage) {
    MVMSpeshGraph *sg = MVM_spesh_graph_create(tc, sf, 1, 0);
    instrument_graph_with_breakpoints(tc, sg);   // inserts an MVM_OP_breakpoint per line change
    sc = MVM_spesh_codegen(tc, sg);              // regenerates the bytecode array
    ...
    sf->body.bytecode      = sf->body.instrumentation->instrumented_bytecode;
    sf->body.bytecode_size = sf->body.instrumentation->instrumented_bytecode_size;
}
```

1. `sf->body.bytecode`, `.bytecode_size`, and `.handlers` are replaced by the instrumented versions.
2. However, `sf->body.annotations_data` and `.num_annotations` — the byte-offset-keyed table mapping bytecode offsets to `{filename, line_number}` — remain pointing to the original, uninstrumented table.
3. Each inserted `MVM_OP_breakpoint` is 10 bytes (2-byte opcode + two 4-byte operands). Every inserted breakpoint shifts all subsequent instructions 10 bytes further in the instrumented bytecode.
4. When `write_stacktrace_frames` (`src/debug/debugserver.c:1062-1065`) inspects `cur_op` (which points into `instrumented_bytecode`) and calculates:
   ```c
   MVMuint32 offset = cur_op - MVM_frame_effective_bytecode(cur_frame);
   MVMBytecodeAnnotation *annot = MVM_bytecode_resolve_annotation(tc, &cur_frame->static_info->body,
                                     offset > 0 ? offset - 1 : 0);
   ```
[...]

PR #2040, opened 2026-09-15
Update line number annotations when instrumenting bytecode

kliklamala:
### Problem

See issue #2039 

When debugging under `--debug-port` (or running with line coverage), bytecode is rewritten and instrumented via `MVM_breakpoint_instrument` -> `add_instrumentation` -> `MVM_spesh_codegen`.

Each inserted `MVM_OP_breakpoint` shifts bytecode offsets forward by 10 bytes. However, `sf->body.annotations_data` and `sf->body.num_annotations` were left pointing to the original, uninstrumented table. As execution advanced through a frame, the offset into instrumented bytecode drifted progressively further ahead of the stale annotation table.

As a result:
- Stacktraces (`Thread Stack Trace Response`), breakpoint notifications, and `Step Completed` reported drifted/wrong line numbers (freezing on later lines or jumping ahead).
- In long routines or after multiple step-overs, the reported location drifted significantly from the actual execution position.

### Solution

1. **Codegen generates updated annotations**: During `MVM_spesh_codegen`, `write_instructions` now processes `MVM_SPESH_ANN_LINENO` annotations already present in the spesh graph, recording their new byte offsets (`ws->bytecode_pos`) into a newly constructed `annotations_data` buffer.
2. **Breakpoint instruction starts the line**: In `instrument_graph_with_breakpoints`, the `MVM_SPESH_ANN_LINENO` annotation is attached directly to the inserted `breakpoint_ins`. Because the interpreter advances past the breakpoint instruction before calling the debug server, `offset > 0 ? offset - 1 : 0` in `write_stacktrace_frames` lands cleanly inside the breakpoint op and resolves immediately to the correct line.
3. **Frame update & memory management**: `line_numbers_instrument` updates `sf->body.annotations_data` and `sf->body.num_annotations` from the instrumentation struct, and the heap-allocated instrumented annotations table is freed in `gc_free`.

### Verification

Tested with a sequential stepping script across various statement counts ($N=0$, $N=200$, $N=1000$ preceding statements before breakpoint):

- **Before fix ($N=200$):**
  Reported line froze on line 210 for 7 consecutive steps while execution was at 203..209 (delta up to 7 lines).
- **After fix ($N=0, 200, 1000$):**
  Reported line matched the executing line with delta 0 across all steps and stack traces.


patrickbkr:
I'm curious, how did you find out about this issue? Have you been using a debugger? Which one?

kliklamala:
I'm playing around to develop a Raku-LSP for VSCode and that contains a debugger. While implementing "Step Over" I recognized the line drift.

patrickbkr:
Oh nice! I'd be interested to know how you approach this. I've been toying with the idea of creating an LSP for a long time and am still mulling on some of the challenges. Maybe we should have a chat some time. :-)

Do you try to resolve calls to the callee, allowing "go to definition"? If so, how?

kliklamala:
It's a little bit ot :smirk:, but currently my LSP maintains an index in the workspace folder.  Local symbols are resolved via the AST of the current file. The implementation is nearly complete, but still in alpha state.

src/6model/reprs/MVMStaticFrame.c
```diff
@@ -190,7 +190,11 @@ static void gc_free(MVMThreadContext *tc, MVMObject *obj) {
     MVM_free(body->local_types);
     MVM_free(body->lexical_types);
     MVM_free(body->lexical_names_list);
-    MVM_free(body->instrumentation);
+    if (body->instrumentation) {
+        if (body->instrumentation->instrumented_annotations_data)
```
patrickbkr: This `if` is redundant.

src/6model/reprs/MVMSpeshCandidate.c
```diff
@@ -296,6 +296,8 @@ void MVM_spesh_candidate_add(MVMThreadContext *tc, MVMSpeshPlanned *p) {
     candidate->body.local_types   = sg->local_types;
     candidate->body.lexical_types = sg->lexical_types;
 
+    if (sc->annotations_data)
```
patrickbkr: This `if` is redundant.

src/instrument/crossthreadwrite.c
```diff
@@ -119,6 +119,8 @@ static void add_instrumentation(MVMThreadContext *tc, MVMStaticFrame *sf) {
     ins->uninstrumented_bytecode_size = sf->body.bytecode_size;
     sf->body.instrumentation = ins;
     MVM_spesh_graph_destroy(tc, sg);
+    if (sc->annotations_data)
```
patrickbkr: This `if` is redundant.

src/profiler/instrument.c
```diff
@@ -332,6 +332,8 @@ static void add_instrumentation(MVMThreadContext *tc, MVMStaticFrame *sf) {
     ins->uninstrumented_bytecode_size = sf->body.bytecode_size;
     sf->body.instrumentation = ins;
     MVM_spesh_graph_destroy(tc, sg);
+    if (sc->annotations_data)
```
patrickbkr: This `if` is redundant.

src/spesh/codegen.c
```diff
@@ -35,8 +35,39 @@ typedef struct {
 
     /* Working deopt users state (so we can allocate it once and re-use it). */
     AllDeoptUsers all_deopt_users;
+
+    /* Bytecode annotations data and count. */
+    MVMuint8 *annotations_data;
```
patrickbkr: This is specifically for file/line info. Can we name it something more specific?
`location_data`, `src_loc_data`, `src_line_data`, or something like that?

patrickbkr: Is there a reason to have a plain byte array instead of a struct? Using a struct would get rid of the bytefiddling work with this datastructure.

src/spesh/codegen.c
```diff
@@ -35,8 +35,39 @@ typedef struct {
...
+static void add_annotation(SpeshWriterState *ws, MVMuint32 bytecode_pos, MVMuint32 filename_idx, MVMuint32 line_number) {
+    if (ws->num_annotations > 0) {
+        MVMuint8 *last_ann = ws->annotations_data + (ws->num_annotations - 1) * 12;
+        MVMuint32 last_offset = *(MVMuint32 *)last_ann;
+        MVMuint32 last_filename = *(MVMuint32 *)(last_ann + 4);
+        MVMuint32 last_line = *(MVMuint32 *)(last_ann + 8);
+        if (last_filename == filename_idx && last_line == line_number) {
+            return;
+        }
+        if (last_offset == bytecode_pos) {
```
patrickbkr: Maybe add a comment explicitly stating the implications of this. I.e. When multiple lines collapse, we'll reference the last line. I.e. the debugger will jump to the last line of the lines that were collapsed.

src/instrument/line_coverage.c
```diff
@@ -309,9 +316,13 @@ static void add_instrumentation(MVMThreadContext *tc, MVMStaticFrame *sf, MVMuin
     ins->instrumented_bytecode        = sc->bytecode;
     ins->instrumented_handlers        = sc->handlers;
     ins->instrumented_bytecode_size   = sc->bytecode_size;
+    ins->instrumented_annotations_data = sc->annotations_data;
+    ins->instrumented_num_annotations  = sc->num_annotations;
     ins->uninstrumented_bytecode      = sf->body.bytecode;
     ins->uninstrumented_handlers      = sf->body.handlers;
     ins->uninstrumented_bytecode_size = sf->body.bytecode_size;
+    ins->uninstrumented_annotations_data = sf->body.annotations_data;
+    ins->uninstrumented_num_annotations  = sf->body.num_annotations;
```
patrickbkr, line 318: I'm curious: Do you know why these are persisted? They don't seem to be read anywhere. (Same for `ins->uninstrumented_bytecode` and `ins->uninstrumented_handlers`.)

src/instrument/line_coverage.c
```diff
@@ -329,6 +340,10 @@ static void line_numbers_instrument(MVMThreadContext *tc, MVMStaticFrame *sf, MV
             MVM_free(sf->body.handlers);
         sf->body.handlers      = sf->body.instrumentation->instrumented_handlers;
         sf->body.bytecode_size = sf->body.instrumentation->instrumented_bytecode_size;
+        if (sf->body.instrumentation->instrumented_annotations_data) {
```
patrickbkr: It seems this `if` is unnecessary.

src/instrument/line_coverage.c
```diff
@@ -102,6 +102,13 @@ static void instrument_graph_with_breakpoints(MVMThreadContext *tc, MVMSpeshGrap
             breakpoint_ins->operands[0].lit_i32 = file_bp_idx;
             breakpoint_ins->operands[1].lit_i32 = line_number;
 
+            MVMSpeshAnn *lineno_ann = MVM_spesh_alloc(tc, g, sizeof(MVMSpeshAnn));
+            lineno_ann->next = breakpoint_ins->annotations;
```
patrickbkr: This fails to change the previous' `next` pointer to the inserted annotation. Doesn't it?

src/instrument/line_coverage.c
```diff
@@ -102,6 +102,13 @@ static void instrument_graph_with_breakpoints(MVMThreadContext *tc, MVMSpeshGrap
...
+            lineno_ann->type = MVM_SPESH_ANN_LINENO;
+            lineno_ann->data.lineno.filename_string_index = filename_string_index;
+            lineno_ann->data.lineno.line_number = line_number;
+            breakpoint_ins->annotations = lineno_ann;
+
```
patrickbkr: If I understand correctly, adding a line no annotation to the breakpoint instruction is an optimization and does not contribute to fixing the line number drift. If so, can you move this to a separate commit?

patrickbkr (changes requested):
All in all this change looks good. Thank you!

src/instrument/line_coverage.c
kliklamala: The comment at `MVMStaticFrame.h` (line 135) suggests, that this extra state data is carried, "so that the can later be removed".  I assume it was meant for reverting instrumentation?!

src/instrument/line_coverage.c
kliklamala: Yes, you're right. Better move this to a separate pr.

src/spesh/codegen.c
```diff
@@ -8,6 +8,16 @@ typedef struct {
...
+/* One source location entry: the bytecode offset it starts at, and the
+ * file/line it maps to. Layout must stay binary-compatible with the
+ * MVMuint8 *annotations_data format read generically by
+ * MVM_bytecode_resolve_annotation / MVM_bytecode_advance_annotation. */
+typedef struct {
+    MVMuint32 bytecode_offset;
+    MVMuint32 filename_string_heap_index;
+    MVMuint32 line_number;
+} SpeshAnnotationEntry;
+
```
patrickbkr: I revert my opinion. Having looked at `MVM_bytecode_resolve_annotation` I realize, that the term `annotation` for file/line info as well as the data layout in a plain byte array are already there. I guess we should try to stay consistent with the existing code. So back to `annotation_data` I'd say. But can you add a few comments making it clear what this actually contains? Especially the fact that the spesh graph has annotations that are really many-kinded (line/file info being one of many) makes this confusing.

What's your opinion with respect to byte-fiddling vs using a struct? Is this serialized somewhere? If so that might be the reason for not using a struct.

In any case, we should either do byte-fiddling like the rest of the file/line annotation code, or we change it to a struct everywhere.

patrickbkr:
FYI: Lizmat and I decided this should not go into the upcoming release, because there major RakuAST refactor is about to go live and we shouldn't pile more changes onto that. That was probably the reason why she marked the PR as draft. As in "don't merge this right now".

src/spesh/codegen.c
kliklamala: In general I would prefer structs. They make clear, what the intend is, but I follow your argument, to stay consistent with the handling at other locations.

Btw: this `annotation_data` seems not to be serialized anywhere, but the one on `MVMStaticFrameBody` deserializes  (bytecode.c:538).

